PR #787, opened 2026-04-12
Fix sub-component compilation and export handling

### LINKED ISSUE #786: Sub-components (non-exported function components) in same file emit raw jsxDEV calls
## Summary

When defining a helper function component in the same file as the main `'use client'` exported component, the compiler does not process JSX inside the helper. Instead, it emits raw `jsxDEV_<hash>(...)` calls, causing `ReferenceError` at runtime.

## Minimal reproduction

```tsx
'use client'

interface Item {
  id: string
  name: string
}

// Sub-component — JSX here is NOT compiled by BareFootJS
function ItemRow({ item, onDelete }: { item: Item; onDelete: (id: string) => void }) {
  return (
    <div className="row">
      <span>{item.name}</span>
      <button onClick={() => onDelete(item.id)}>Delete</button>
    </div>
  )
}

// Main exported component — JSX here IS compiled
export function ItemList(props: { items: Item[] }) {
  const [items, setItems] = createSignal<Item[]>(props.items)

  const handleDelete = (id: string) => {
    setItems(items().filter((i: Item) => i.id !== id))
  }

  return (
    <div>
      {items().map((item: Item) => (
        <ItemRow key={item.id} item={item} onDelete={handleDelete} />
      ))}
    </div>
  )
}
```

## Expected behavior

The compiler should process JSX inside `ItemRow` and generate proper template literals + `bindEvents` for the sub-component, just like it does for the exported component.

## Actual behavior

The compiled `.client.js` contains raw `jsxDEV_<hash>(...)` calls for the sub-component's JSX:

```javascript
function ItemRow({item, onDelete}) {
  return jsxDEV_7x81h0kn("div", {
    className: "row",
    children: [
      jsxDEV_7x81h0kn("span", { children: item.name }),
      jsxDEV_7x81h0kn("button", { onClick: () => onDelete(item.id), children: "Delete" })
    ]
  })
}
```

The `jsxDEV_7x81h0kn` function is never defined, causing:

```
Uncaught ReferenceError: jsxDEV_7x81h0kn is not defined
```

## Why this matters

Sub-components are a natural way to reduce duplication in React/JSX code. Without this, users must either:
- Inline all JSX (leading to large, duplicated code)
- Use class constant strings for styling (losing JSX composability)
- Create separate `'use client'` files for each component (excessive file fragmentation)

## Workaround

Inline all JSX directly in the exported component and use shared class-name constants for deduplication.

## Environment

- barefootjs: local (v0.0.1)
- Hono adapter

## Related

- #783 (conditional branch `.map()` also emits raw `jsxDEV`)

kfly8:
## Summary

Closes #786

- **Non-exported components no longer receive `export` keyword** in adapter output. Added `isExported` field to `IRMetadata`, populated during analysis from `ExportKeyword` modifier and `export { Name }` re-export patterns.
- **Raw JSX no longer leaks into client JS** from local functions with multiple JSX returns. Added `sanitizeJsxInBody()` to strip JSX from non-inlinable function bodies, and `BF045` compiler warning guiding users to extract top-level components or use single-return patterns.
- **Renamed `listExportedComponents()` to `listComponentFunctions()`** since it returns ALL PascalCase function declarations regardless of export status. Old name preserved as re-export alias for compatibility.

## Test plan

- [x] New test file `sub-component-compilation.test.ts` with 10 test cases covering:
  - Top-level non-exported function component (exact reproduction case)
  - Top-level non-exported arrow function component
  - Non-exported stateless component
  - Non-PascalCase helper with conditional JSX returns (BF045 warning)
  - `export { Name }` re-export pattern
- [x] Compiler unit tests (529 pass)
- [x] Adapter conformance tests (161 pass)
- [x] Client runtime tests (142 pass)
- [x] TypeScript type checks pass for both `packages/jsx` and `packages/hono`


kfly8:
## CI Status

`e2e-site-ui` fails on 22 chart Playground tests (area-chart, bar-chart, pie-chart, radar-chart, radial-chart). All other CI jobs pass (test, test-ui, update-expected-html, e2e-hono, e2e-echo).

**Root cause analysis:**
- Verified the compiled output is identical between `main` and this branch (only difference is client JS content hash)
- All 22 failing tests pass locally (macOS, Playwright)
- The failures are `toBeVisible()` on `[data-*-preview]` elements — the element exists but is "hidden" in CI's headless Chromium
- These are chart SVG rendering tests that may be sensitive to CI environment (headless Chromium GPU rendering, timing)
- No `__bfInlineScripts` errors, no runtime errors in WebServer logs

This appears to be a CI-environment flaky test issue, not caused by this PR's changes.

PR #926, opened 2026-04-20
fix(xyflow): import reactive APIs from @barefootjs/client/runtime

kfly8:
## Summary

- Normalize every `@barefootjs/client` import in `packages/xyflow/src` (plus the xyflow tests) to `@barefootjs/client/runtime`, matching what `flow.ts` already does.
- Keeps the whole library on a single reactive runtime copy, which is what the existing `setViewport triggers reactive updates` test expects.

## What was broken

`xyflow/src/flow.ts` imported `createEffect` from `@barefootjs/client/runtime`, but `xyflow/src/store.ts` (and most siblings) imported `createSignal` / `createEffect` from `@barefootjs/client`. When an app bundles xyflow, each entry resolves to its own pre-built dist (`client/dist/index.js` and `client/dist/runtime/index.js`), and each dist inlines its own copy of the reactive runtime — separate `Listener` / `Owner` closures and separate subscriber sets.

The practical effect downstream (seen in a Hono + Cloudflare Workers app that bundles DeskCanvas.tsx with `bun build`): `store.setViewport(...)` updates the signal in one runtime copy, while `createEffect(() => { const vp = store.viewport(); viewportEl.style.transform = ... })` in `flow.ts` is registered with the other copy's scheduler. The DOM effect never re-runs, so:

- Scroll-to-pan (`panOnScroll: true`) never updates `viewportEl.style.transform` even though `d3-zoom`'s internal `__zoom` moves.
- Cmd+scroll (`zoomActivationKeyCode: 'Meta'`) has the same symptom — zoom applies in d3 but the viewport DOM stays at `translate(0px, 0px) scale(1)`.

## What this PR does

- `packages/xyflow/src/**/*.ts`: switch `from '@barefootjs/client'` → `from '@barefootjs/client/runtime'` for the 11 modules that import reactive / lifecycle primitives (`createSignal`, `createEffect`, `createMemo`, `onCleanup`, `untrack`, `createContext`).
- `packages/xyflow/src/types.ts`: collapse `Signal` / `Memo` / `ComponentDef` type imports into a single `/runtime` import.
- `packages/xyflow/src/__tests__/*.ts`: mirror the same change so unit tests exercise the library through the same runtime copy.

`flow.ts`, `hooks.ts`, and the existing `/runtime` imports in `node-wrapper.ts` already pointed at `/runtime`, so this PR just removes the inconsistency.

## Test plan

- [x] `bun run --filter '@barefootjs/client' build && cd packages/xyflow && bun test` — 30 pass / 1 pre-existing Playwright-in-bun failure / 1 pre-existing error (unchanged from `main`).
- [x] Reproduced the viewport-not-updating bug with a downstream app (scroll-to-pan + Cmd+zoom), confirmed both behaviors recover after this change.

### DIFF
--- a/packages/xyflow/src/__tests__/compat.test.ts
+++ b/packages/xyflow/src/__tests__/compat.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from 'bun:test'
-import { createRoot } from '@barefootjs/client'
+import { createRoot } from '@barefootjs/client/runtime'
 import { createFlowStore } from '../store'
 
 /**
--- a/packages/xyflow/src/__tests__/edge-renderer.test.ts
+++ b/packages/xyflow/src/__tests__/edge-renderer.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from 'bun:test'
-import { createRoot } from '@barefootjs/client'
+import { createRoot } from '@barefootjs/client/runtime'
 import { createFlowStore } from '../store'
 
 describe('Edge processing in store', () => {
--- a/packages/xyflow/src/__tests__/store.test.ts
+++ b/packages/xyflow/src/__tests__/store.test.ts
@@ -1,5 +1,5 @@
 import { describe, test, expect } from 'bun:test'
-import { createEffect, createRoot } from '@barefootjs/client'
+import { createEffect, createRoot } from '@barefootjs/client/runtime'
 import { createFlowStore } from '../store'
 
 describe('createFlowStore', () => {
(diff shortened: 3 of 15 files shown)


kfly8:
Closing — this only fixes the symptom in xyflow's source. The root cause is that `@barefootjs/client` and `@barefootjs/client/runtime` ship as two independent dist bundles, each inlining its own copy of the reactive runtime. Downstream bundlers (e.g. `bun build DeskCanvas.tsx`) that resolve both entries end up with divergent `Listener`/`Owner` closures. The existing CLI build step (`packages/cli/src/lib/build.ts` L460+) only rewrites `/runtime` imports in compiler-emitted output — libraries like xyflow that are consumed as npm packages slip through. Will open an issue to discuss the right layer for the fix (likely xyflow's build normalizing imports post-bundle, or client's dist structure sharing the reactive module).

Issue #937, opened 2026-04-21
compiler: adopt Solid-style wrap-by-default fallback to eliminate silent-drop bugs


kfly8:
Follow-up to PR #934 / #935 / #936 (issues #933 / #932 / #931). All three shared one meta-pattern; this issue closes it at the architecture level.

## The bug class we're fixing

Every analyzer / emitter pass holds an implicit **allow-list of recognised shapes**. Anything outside the allow-list is silently dropped — producing a broken client JS, a missing SSR helper, or a subtly-wrong hydration that only surfaces at E2E time.

| Issue | Allow-list | What got silently dropped |
|-------|------------|--------------------------|
| #933 | `collectConstant` knew about module-scope declarations; `collectInitStatement` knew about writes. Link between them missing. | Module-scope `let` referenced by an init statement → client-JS emit lost the declaration → ESM strict-mode `ReferenceError` |
| #932 | `extractSingleJsxReturn` recognised one-exit JSX bodies only | A PascalCase helper with `switch` / if-else-chain multi-return → dropped from marked-template output → SSR `ReferenceError: NavIcon is not defined` |
| #931 | `isSignalDeclaration` matched literal `createSignal` callee only | `const [a, b] = myFactory(0)` where `myFactory` wraps `createSignal` → zero signal declarations emitted, zero warnings |

Same failure mode, three different entry points. Fixing each occurrence individually is whack-a-mole because the architectural gap (silent drop vs safe fallback) is not closed.

### Project priority (memorialised in CLAUDE memory)

1. **Faithful TSX reproduction.** If it compiles, it must behave as written. Silent drops are the worst possible outcome.
2. **Performance / small client JS.** A tuning knob, not a hard constraint.

### What SolidJS does (and why it avoids silent-drop bugs structurally)

Research on `babel-plugin-jsx-dom-expressions`:

- No custom IR. Babel AST in, Babel AST out — no information-loss boundary.
- Heuristic: **if an expression contains a function call or property access, wrap it in `createEffect` / runtime-tracked getter. If it doesn't, leave it static.** ([babel-plugin-jsx-dom-expressions README](https://github.com/ryansolid/babel-plugin-jsx-dom-expressions))
- False positive (unnecessary wrap) = 1 extra closure at runtime. Harmless.
- False negative = impossible — any unknown shape falls through to the \"wrap in effect\" default.
[...]

Issue #932, opened 2026-04-20
compiler: multi-return JSX helper functions dropped from marked-template output


kfly8:
## Context

Discovered while verifying PR #930. Sibling to #931.

A module-level helper function that returns JSX via a **multi-branch `switch` / `if-else` with multiple `return` statements** is silently dropped from the emitted marked-template file, producing `ReferenceError: <HelperName> is not defined` at SSR time.

### Repro

```tsx
// admin-shell.tsx (no "use client")
function NavIcon({ name }: { name: string }) {
  switch (name) {
    case 'home':  return <svg><path d="..." /></svg>
    case 'chart': return <svg><path d="..." /></svg>
    // …
    default: return null
  }
}

export function AdminShell() {
  return (
    <nav>
      {NAV_ITEMS.map(item => <a><NavIcon name={item.icon} /></a>)}
    </nav>
  )
}
```

Compiled `admin-shell.tsx` (marked template) imports only the direct external dependencies and the main component — **`NavIcon`'s function body is missing entirely**, yet the JSX inside `AdminShell`'s return still references `<NavIcon name={item.icon} />`. SSR fails with:

```
ReferenceError: NavIcon is not defined
  at /dist/components/gallery/admin/admin-shell.tsx:69:1198
```

The companion `admin-shell.client.js` _does_ contain `hydrate('NavIcon', {...})` with a correct template — so the compiler knows NavIcon exists for client purposes. SSR and client paths are asymmetric.

### Working baseline

`site/ui/components/empty-demo.tsx`'s `PackageIcon` is structurally identical **except** it has a single `return <svg>…</svg>`. It is preserved in the marked-template output. The crucial difference is single-return vs multi-return.

### Root cause (code anchor)

`packages/jsx/src/analyzer.ts:903-912` — the detection that makes a helper eligible for inline preservation:

```ts
let isJsxFunction = false
if (containsJsx && node.body) {
  const jsxReturn = extractSingleJsxReturn(node.body)
  if (jsxReturn) {
    isJsxFunction = true
    ctx.jsxFunctions.set(name, { jsxReturn, params: … })
  }
}
```

[...]

Issue #931, opened 2026-04-20
compiler: silent hydration break when wrapping createSignal in a helper


kfly8:
# Support user-authored reactive-factory helpers (same-file inline expansion)

_Draft — discovered while building the Admin Dashboard gallery (#929, PR #930).
To be filed as a standalone compiler issue once reviewed._

## Context

While wiring up `sessionStorage`-backed state in PR #930, I wrote the natural SolidJS / React-style helper:

```ts
// shared/gallery-admin-storage.ts
export function createPersistentSignal<T>(key: string, initial: T) {
  const [get, set] = createSignal<T>(readStored(key, initial))
  const setAndPersist = (v: T | ((prev: T) => T)) => {
    const next = typeof v === 'function' ? (v as (p: T) => T)(get()) : v
    set(next)
    writeStored(key, next)
    return next
  }
  return [get, setAndPersist] as const
}
```

Then used it in a component:

```tsx
"use client"
import { createPersistentSignal } from './shared/gallery-admin-storage'

export function AdminTimeRange() {
  const [timeRange, setTimeRange] = createPersistentSignal('timeRange', '30d')
  return (
    <button onClick={() => setTimeRange('7d')}>{timeRange()}</button>
  )
}
```

**Outcome:** the compiled output is silently broken. The emitted client JS contains `setTimeRange(...)` and `timeRange()` calls, but **no signal declaration** — `const [timeRange, setTimeRange] = createSignal(...)` is never emitted. The marked template (SSR) references `timeRange()` with no stub, so server-side rendering also throws `ReferenceError: timeRange is not defined`.

The compiler emits **zero errors and zero warnings.**

### Why this happens (code anchors)

The analyzer only recognizes `createSignal` / `createMemo` when the callee identifier is literally named `createSignal` / `createMemo`:

- [`packages/jsx/src/analyzer.ts:512-521` — `isSignalDeclaration`](../../packages/jsx/src/analyzer.ts)
- [`packages/jsx/src/analyzer.ts:575-584` — `isMemoDeclaration`](../../packages/jsx/src/analyzer.ts)

[...]

PR #939, opened 2026-04-21
fix(compiler): Solid-style wrap-by-default for text interpolation (#937)

kfly8:
## Summary

Pilot for #937. Widen the dynamic-text gate in `collect-elements.ts` and the slot allocator in `jsx-to-ir.ts` so any JSX text interpolation with a function call or signal-getter call gets wrapped in `createEffect`, even when the analyzer cannot statically prove the expression reactive. Pure literals and bare identifiers stay un-wrapped.

This closes the silent-drop failure class behind #931 / #932 at the collector layer:

- **False positive** (extra `createEffect` that subscribes to nothing): one closure at init time, no rerun — harmless.
- **False negative** (the old behaviour): SSR value frozen in the DOM, no client-side update — the bug class we're closing.

## Changes

- `packages/jsx/src/ir-to-client-js/collect-elements.ts` — widened the `expression` case: now also wraps when `node.hasFunctionCalls` or `node.callsReactiveGetters` is true, not only when `node.reactive` is proven.
- `packages/jsx/src/jsx-to-ir.ts` — widened `needsSlot` in `transformExpression()` with the same two flags so the expression reaches the gate with a `slotId` allocated. Without this, the collect gate widening would be a no-op for the exact silent-drop cases we want to catch.
- `packages/jsx/src/__tests__/runtime-fallback-wrap.test.ts` — 6 regression cases (known signal, unrecognised outer call, unrecognised call with no reactive source, static literal, bare identifier, method chain on props).

## Fixture diff

Baseline: `site/ui/dist/components/**/*.js` on `main`. After rebuild, **2 files of 250 differ** (≈0.8%, well under the 30% rollback threshold from the issue):

- `site/ui/dist/components/analytics-dashboard-demo-*.js`
- `site/ui/dist/components/gallery/admin/analytics-demo-*.js`

Both diffs reduce to one new harmless `createEffect` per file, for `{formatCurrency(REVENUE_TARGET)}` (a call on a module-level constant — previously a silent drop, now a no-op effect). Everything else is slot-id number shift.

## Test plan

- [x] `bun test packages/jsx` — 656 pre-existing tests pass; 6 new tests pass.
- [x] `bun test packages/{jsx,dom,cli,client,hono,go-template,adapter-tests}` — 1464 tests pass.
- [x] Clean `bun run build` succeeds on both `main` and this branch.
- [x] Site/ui Playwright E2E: 1385 pass / 2 fail — the two failures (`form-builder.spec.ts` required-toggle + bug-regression) reproduce on `main` with no changes applied, so they are pre-existing and unrelated.

## Follow-ups (out of scope for this PR)

As the issue lays out, the pattern replicates to: JSX attributes, conditionals, loops, and prop bindings on child components. Each is its own PR.

### DIFF
--- a/packages/jsx/src/__tests__/runtime-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/runtime-fallback-wrap.test.ts
@@ -102,10 +102,10 @@ describe('Solid-style wrap-by-default fallback (#937)', () => {
     `
 
     const clientJs = getClientJs(source, 'Greeting.tsx')
-    // No createEffect for the text — only the event handler and signal setup.
-    // We look specifically for the absence of a createEffect that writes the
-    // literal into a text node.
-    expect(clientJs).not.toMatch(/createEffect\([^)]*\)\s*=>\s*\{[^}]*['"]hello['"]/)
+    // A pure literal has no slotId and no dynamic binding — nothing to
+    // update at runtime. The only signal in this component is read-only in
+    // the event handler, so no createEffect should be emitted at all.
+    expect(clientJs).not.toContain('createEffect')
   })
 
   test('bare identifier (no calls) stays un-wrapped', () => {
@@ -124,6 +124,10 @@ describe('Solid-style wrap-by-default fallback (#937)', () => {
     `
 
     const clientJs = getClientJs(source, 'Label.tsx')
+    // Guard against the identifier being dropped entirely from the module —
+    // an over-eager regression could satisfy the negative assertion below by
+    // removing `label` altogether. We assert presence in the init scope too.
+    expect(clientJs).toContain('label')
     // No text-update createEffect should be emitted for a static bare ident.
     expect(clientJs).not.toMatch(/nodeValue\s*=\s*String\(label/)
   })
--- a/packages/jsx/src/__tests__/runtime-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/runtime-fallback-wrap.test.ts
@@ -1,0 +1,146 @@
+/**
+ * Regression tests for #937: Solid-style wrap-by-default fallback for JSX
+ * text interpolation.
+ *
+ * Before this pilot, `collect-elements.ts` gated `dynamicElements` on
+ * `node.reactive && node.slotId`. Any expression the analyzer couldn't
+ * statically prove reactive silently flowed through as a static text node —
+ * its SSR value was frozen in the DOM and never updated. That silent-drop
+ * path is the architectural gap behind #931 / #932 / and various
+ * factory-wrapping regressions.
+ *
+ * The fix widens the gate to also wrap expressions whose AST carries
+ * `hasFunctionCalls` or `callsReactiveGetters`. Over-wrapping a pure call
+ * that subscribes to nothing is harmless (one extra closure); under-wrapping
+ * a reactive read is the bug class we're closing. Pure literals and bare
+ * identifiers without calls stay un-wrapped so the optimisation the flag was
+ * originally computed for is preserved.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { compileJSXSync } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
+
+const adapter = new TestAdapter()
+
+function getClientJs(source: string, filename: string): string {
+  const result = compileJSXSync(source, filename, { adapter })
+  expect(result.errors.filter(e => e.severity === 'error')).toHaveLength(0)
+  const clientJs = result.files.find(f => f.type === 'clientJs')
+  expect(clientJs).toBeDefined()
+  return clientJs!.content
+}
+
+describe('Solid-style wrap-by-default fallback (#937)', () => {
+  test('known signal getter still wraps (regression guard)', () => {
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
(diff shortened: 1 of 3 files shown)


kfly8 (commented):
## Overview

Pilot implementation for issue #937. Minimum-scope change to close the silent-drop bug class for `{expr}` JSX text interpolation. Two source files plus a new test file, +181/-11 lines total.

- `jsx-to-ir.ts::transformExpression` — widens `needsSlot` from `reactive || isClientOnly || refsLoopParam` to also include `callsReactive || hasCalls`. Expressions containing function calls now get a `slotId`.
- `collect-elements.ts::collectElements` `case 'expression'` — widens the gate from `node.reactive && slotId` to `shouldWrap = reactive || callsReactiveGetters || hasFunctionCalls`.
- `runtime-fallback-wrap.test.ts` — 6 cases (known signal / unrecognised call / `Date.now()` / static literal / bare identifier / props method chain).

## Strengths

- **Scope matches the issue's pilot plan exactly.** One function per file, focused change — a good template for horizontal rollout.
- **Both files are necessary in combination.** Touching only `collect-elements.ts` would be a no-op when `slotId` is null, so the `jsx-to-ir.ts` slot-allocation widening is a prerequisite. Missing either side would make the pilot non-functional.
- **Fixture diff is 2 files out of 250 (~0.8%)** — well under the 30% rollback threshold set by the issue. Both changes are module-scope `formatCurrency` wrapping — exactly the known silent-drop shape we wanted to catch.
- **Code comments focus on the WHY.** False-positive vs false-negative trade-off, back-references to #931/#932, rationale for widening slot allocation — all on the non-obvious points. No redundant "what" comments.

## Concerns (should address before merge)

### 1. Orphan slot markers in the SSR template are unverified

The `needsSlot` widening now allocates a `slotId` to `conditional` / `logical` / `binary` expressions too. `collect-elements.ts`'s non-`expression` branches never push into `dynamicElements`, so those slots don't drive client-side effects — but if any adapter path emits `bf:sN` markers purely based on `slotId` presence, unused markers could leak into SSR HTML. The PR description says the diff is "just slot-id number shift," which suggests no behavioural change, but **adding a line to the PR description confirming the SSR HTML fixture was diffed** would be safer. If not yet checked, worth running a snapshot diff against `packages/adapter-tests/fixtures/`.

### 2. The "bare identifier stays un-wrapped" test has only a negative assertion

From `runtime-fallback-wrap.test.ts`:

```ts
expect(clientJs).not.toMatch(/nodeValue\s*=\s*String\(label/)
```

This passes if the whole `label` reference is accidentally dropped from the output — not just the effect. Add a positive assertion to guard against that:

```ts
expect(clientJs).toContain('label')
```

### 3. The "static string literal" test uses a dense regex

```
/createEffect\([^)]*\)\s*=>\s*\{[^}]*['"]hello['"]/
```

The `[^}]` matches newlines in JS regex so the pattern works, but it's hard to read and easy to accidentally loosen in future refactors. Simpler alternative: a higher-level check like `!clientJs.includes("'hello'") || !clientJs.includes('createEffect')`, or an occurrence count. Low priority but easy to tighten.

## Nice to have

- **Follow-ups are explicit** in the PR description (attributes / conditionals / loops / prop-bindings). Consider adding 1–2 lines noting which expression shapes still have the silent-drop path until their respective PRs land — helps triage future bug reports.
- **Pre-existing E2E failures are called out explicitly**, which aligns with the project's "never claim merge-ready before CI is green" discipline. Good.

- **Correctness**: closes the targeted silent-drop gate. Only residual risk is concern #1 (orphan slot markers), which appears benign but is unverified in the PR record.
- **Performance**: 2/250 files gain one init-time closure (~tens of bytes each). Negligible.
- **Security**: internal compiler change, no user input surface. N/A.

## Verdict

**Ready to merge** with concern #1 confirmed (a note in the PR body is enough). Concerns #2 and #3 are test-robustness nits, not blockers. The pilot cleanly implements the pattern defined in #937 and sets up the template for rolling out attributes → conditionals → loops as subsequent PRs.

kfly8:
Thanks for the thorough review. Addressed as follows (commit `7d0522ba` on the rebased branch):

### Concern 1 — SSR fixture verification
Ran `bun test packages/adapter-tests` (CSR conformance + SSR-hydration contract) → **177 pass / 0 fail**. Also ran `packages/hono` + `packages/go-template` adapter suites → **127 pass / 0 fail**. No orphan slot markers introduced: fixtures compare `expectedHtml` byte-for-byte against adapter output, so any surplus `bf=` / `<!--bf:...-->` marker would have failed the conformance tests.

### Concern 2 — bare-identifier positive assertion
Added `expect(clientJs).toContain('label')` alongside the existing `not.toMatch(/nodeValue\s*=\s*String\(label/)`. If a regression drops the identifier entirely, the positive check catches it.

### Concern 3 — static-literal assertion simplified
Replaced the fragile `/createEffect\([^)]*\)\s*=>\s*\{[^}]*['"]hello['"]/` with the simpler, higher-level `expect(clientJs).not.toContain('createEffect')` — the Greeting component has no signal reads, so any text-update effect emission would violate the optimisation this case is guarding.

### Nice-to-have
Follow-up shapes (attributes, conditionals, loops, prop-bindings) are listed at the bottom of the PR description; will carry that list into the follow-up issues after merge.

Three commits now, all reproducing green locally. Also added `Co-Authored-By: Claude` to every commit on this branch per the work-style preference.

Issue #943, opened 2026-04-21
compiler: extend wrap-by-default fallback to loops (#937 follow-up)


kfly8:
Follow-up to #937. Pilot landed in #939 for JSX text interpolation. This issue extends the wrap-by-default pattern to **loops** — `{arr.map(...)}` constructs.

This is the highest-risk of the four follow-ups because loops interact with filter / sort chains, nested loops, `isStaticArray` optimisations, and reconciliation. Do not start until #940 (attributes) and #942 (child-component prop bindings) have landed and stabilised — those exercise the same runtime helpers without the loop-specific complications.

## The silent-drop gate today

`packages/jsx/src/jsx-to-ir.ts:1649`
```ts
const isStaticArray = !isSignalOrMemoArray(array, ctx)
```

`isSignalOrMemoArray` (`jsx-to-ir.ts:2188`) matches only literal signal-getter or memo callees. When `isStaticArray === true`, the loop is rendered once at SSR and never reconciled on the client — no `reconcileList` / `reconcileElements` is wired up. For `{computeItems().map(...)}` where `computeItems` is an imported helper, the SSR result is frozen. That is the silent-drop path.

Downstream consumption:
- `collect-elements.ts:253` (top-level loops)
- `collect-elements.ts:596` (branch loops inside conditionals)
- `useElementReconciliation` decision around line 283 — driven by `!isStaticArray && (hasNestedComps || hasInnerLoops)`

## Proposed change

Widen `isSignalOrMemoArray` (or add a companion `arrayNeedsReconciliation`) to return `true` when the array expression contains a call expression — even if the callee is unrecognised. Use AST-level detection via `exprHasFunctionCalls`, which is already available at `jsx-to-ir.ts:99`.

Care points:

1. **Filter / sort chains.** `todos.filter(t => t.done).map(...)` already matches because `todos` is normally a signal/prop. With the wider gate, `getItems().filter(...).map(...)` (callee not recognised) also gets reconcileList — verify filter/sort metadata flows through correctly.
2. **`isStaticArray` is also read by adapters** (e.g., `packages/go-template/`) to decide SSR rendering strategy. Audit adapter code before widening, and run `packages/adapter-tests` to verify `expectedHtml` fixtures remain byte-equal.
3. **Nested loops.** `collectInnerLoops` at `collect-elements.ts:32` recomputes similar metadata. Ensure the wider gate propagates into nested-loop reconciliation.
[...]

Issue #942, opened 2026-04-21
compiler: extend wrap-by-default fallback to child-component prop bindings (#937 follow-up)


kfly8:
Follow-up to #937. Pilot landed in #939 for JSX text interpolation. This issue extends the wrap-by-default pattern to **prop bindings on child components** — `<Card title={expr} />`, `<Badge variant={expr} />`, etc.

## The silent-drop gate today

`packages/jsx/src/ir-to-client-js/collect-elements.ts:416`
```ts
const hasPropsRef = expandedValue.includes('props.')
const hasReactiveExpr = needsEffectWrapper(expandedValue, ctx)
if (hasPropsRef || hasReactiveExpr) {
  ctx.reactiveChildProps.push({ componentName: node.name, slotId: node.slotId, propName: prop.name, attrName, expression: expandedValue, ...pickAttrMeta(prop) })
}
```

Same allow-list pattern as the attribute gate: `needsEffectWrapper` (`reactivity.ts:32`) only recognises known signal getters, memos, and prop parameter names. For `<Card title={formatTitle(page)}>` where `formatTitle` is imported and `page` is a local const holding a signal-getter-wrapping object, the analyzer can't prove the expression reactive → no `reactiveChildProps` entry → the SSR-time title is frozen in the hydrated child.

The `hasPropsRef` branch catches bare `props.xxx` access but not call-based usages like `myFactory(props.id)` where the prop only appears inside a call.

## Proposed change — same shape as #939 pilot

Widen the gate at line 416:

```ts
const hasPropsRef = expandedValue.includes('props.')
const hasReactiveExpr = needsEffectWrapper(expandedValue, ctx)
const hasAnyCall = /\b\w+\s*\(/.test(expandedValue) // wrap-by-default fallback
if (hasPropsRef || hasReactiveExpr || hasAnyCall) { ... }
```

`slotId` on the component node is already allocated (the component always has one when rendered), so no `needsSlot` widening is required — unlike the text-interpolation pilot, this is a one-line change in one file.

Caveats:

- **Spread props** (`{...props}`): handled by `forwardProps()` and already tracks dynamism via the spreadSource path — unaffected by this gate.
- **Event handlers** (`onClick={...}`): the gate is guarded by `isEventHandler` checks upstream, so event props won't accidentally be re-wrapped.
[...]

Issue #940, opened 2026-04-21
compiler: extend wrap-by-default fallback to JSX attribute bindings (#937 follow-up)


kfly8:
Follow-up to #937. Pilot landed in #939 for JSX text interpolation. This issue carries the same pattern to **JSX attribute bindings** — `<div class={expr}>`, `<button disabled={expr}>`, etc.

## The silent-drop gate today

`packages/jsx/src/ir-to-client-js/collect-elements.ts:543`
```ts
if (needsEffectWrapper(expandedValueStr, ctx)) {
  ctx.reactiveAttrs.push({ slotId: element.slotId, attrName: attr.name, expression: expandedValueStr, ...pickAttrMeta(attr) })
}
```

`needsEffectWrapper` (`packages/jsx/src/ir-to-client-js/reactivity.ts:32`) is an allow-list: it returns `true` only if the expression matches a known signal getter, a known memo, or a known prop. Anything else — e.g., `format(count)` where `format` is imported, or `props.items.join(',')` if `items` fails to register as a prop — is silently dropped from `ctx.reactiveAttrs`. At SSR the expression is evaluated into the attribute value once; the client never re-evaluates it, so the attribute stays frozen.

Unlike text interpolation, the `slotId` for attribute-bearing elements is **already** allocated (it is driven by events / reactive children / refs at `jsx-to-ir.ts:353`), so this is a one-file change — only the collector gate needs widening. No `needsSlot` update required.

## Proposed change — one gate, ~6 lines

**File**: `packages/jsx/src/ir-to-client-js/collect-elements.ts` around line 543.

Replace the `needsEffectWrapper` check with a wrap-by-default fallback: if the expression contains a function call (`identifier(...)` or `x.y()`), push it to `reactiveAttrs` even when the analyzer can't statically prove it reactive. A cheap regex (`/\b\w+\s*\(/`) over the expanded expression is sufficient; no AST access here (Phase 2). Over-wrap is harmless (one closure at init time); under-wrap is the silent-drop bug we're closing.

Cross-reference: the same `needsEffectWrapper` gate also fires at line 416 for **child-component prop bindings** — that's tracked separately in #942.

## Test coverage — add `attribute-fallback-wrap.test.ts`

Mirror `runtime-fallback-wrap.test.ts` from #939. Minimum 6 cases:

1. Known signal in `class={`{classFor(count())}`}` still wraps (regression guard).
2. Unrecognised call in `class={format(label)}` now wraps (new behaviour).
[...]

PR #946, opened 2026-04-21
fix(compiler): wrap-by-default fallback for JSX conditional expressions (#941)

### LINKED ISSUE #941: compiler: extend wrap-by-default fallback to conditional expressions (#937 follow-up)
Follow-up to #937. Pilot landed in #939 for JSX text interpolation. This issue extends the wrap-by-default pattern to **conditional expressions used as JSX children** — `{cond ? a : b}`, `{cond && b}`, `{a ?? b}`, `{a || b}`.

## The silent-drop gate today

Three near-identical gates in `packages/jsx/src/jsx-to-ir.ts`:

- `transformConditional` at line 971: `const reactive = isReactiveExpression(condition, ctx, node.condition)`
- `transformLogicalAnd` at line 997: `const reactive = isReactiveExpression(condition, ctx, node.left)`
- `transformNullishCoalescing` at line 1051: `const reactive = isReactiveExpression(leftText, ctx, node.left)`

In all three, if `reactive` is false and the condition doesn't reference a loop param, `slotId` is null and the node never makes it into `ctx.conditionalElements` in `collect-elements.ts`. At SSR the branch is chosen once based on the condition's initial value; on the client the condition is never re-evaluated, so the rendered branch is frozen.

The failure mode is the same shape as the text-interpolation silent-drop: `{computeVisible(x) ? <A/> : <B/>}` where `computeVisible` is an imported helper silently renders only the SSR-time branch.

## Proposed change

Mirror the pilot. Compute `exprHasFunctionCalls(astNode)` and `exprCallsReactiveGetters(astNode, ctx)` (already available at `jsx-to-ir.ts:99` / `:73`) at each of the three gates, and:

1. Widen the `slotId` allocation so any condition containing a call gets a slotId.
2. Store the flags on the IR (IRConditional currently has no `hasFunctionCalls` field — add it, mirroring `IRExpression` at `types.ts:118-123`).
3. Widen the collector gate in `collect-elements.ts:232` (`case 'conditional'`) to push to `ctx.conditionalElements` when `reactive || hasFunctionCalls` is true.

Careful points:

- **Branch re-rendering.** When the condition flips on the client, the current implementation uses `insert()` with a `template` function that rebuilds the branch HTML. That path already works for `reactive === true`; verify the same path activates cleanly under the wider gate with fixtures from `site/ui/dist/components/**/*.js`.
- **Nested conditionals.** Nested branches are collected per-parent via `collectBranchConditionals` at `collect-elements.ts:716`. The gate there reads `n.reactive && n.slotId` — widen symmetrically.
- **Loop-param conditions.** Already handled by `loopParamReactive`. Preserve that path unchanged.

## Test coverage — add `conditional-fallback-wrap.test.ts`

Mirror `runtime-fallback-wrap.test.ts` from #939. Minimum 6 cases:

1. `{count() > 0 ? <A/> : <B/>}` where `count` is a known signal still wraps (regression guard).
2. `{isVisible(x) ? <A/> : <B/>}` where `isVisible` is an imported helper now wraps (new behaviour).
3. `{flags.enabled && <A/>}` where `flags` is an untracked const with calls on it wraps.
4. `{staticValue ? <A/> : <B/>}` where `staticValue` is a literal-derived const stays un-wrapped.
5. `{a ?? <Fallback/>}` where `a` is a bare identifier stays un-wrapped.
6. Nested — `{outer() ? (inner() ? <A/> : <B/>) : <C/>}` — both levels wrap.

## Fixture sweep

Same rollback threshold as #937 (≈30 %). Pilot was 0.8 %; conditionals are less common than text interpolation but each wrap touches a wider template region, so expect moderate fixture diffs — review per commit body before merge.

## Non-goals

- Redesigning the branch-template mechanism. This issue only widens the gate.
- Rewriting logical-OR / nullish-coalescing detection to be non-string-based. Stays as is.
- Deduplicating the three near-identical gates into one helper. A structural cleanup, not in scope here.

## References

- Sibling follow-ups: #940 (attributes), #942 (child-component prop bindings), #943 (loops)

kfly8:
## Summary

Follow-up to #937. Carries the Solid-style wrap-by-default fallback landed in #939 (text interpolation) and #940 (attributes) to **conditional expressions used as JSX children** — `{cond ? a : b}`, `{cond && b}`, `{a ?? b}`, `{a || b}`. Closes #941.

Before this change, three near-identical gates in `jsx-to-ir.ts` (`transformConditional`, `transformLogicalAnd`, `transformNullishCoalescing`) allocated `IRConditional.slotId` only when `isReactiveExpression(condition)` returned true (an allow-list of known signal getters, memos, and props). Any condition the analyzer couldn't statically prove reactive — e.g. `{isVisible(x) ? <A/> : <B/>}` where `isVisible` is an imported helper — got no slotId. At SSR the branch was chosen once; the client never re-evaluated the condition, so the rendered branch stayed frozen.

The fix mirrors the pilot: compute `exprHasFunctionCalls` / `exprCallsReactiveGetters` on the condition AST at each site, widen `needsSlot` with both flags, store them on the `IRConditional`, and widen the collector gate in `collect-elements.ts` (top-level + nested branch) to wrap on `reactive || callsReactiveGetters || hasFunctionCalls`.

## Changes

- `packages/jsx/src/types.ts` — add `callsReactiveGetters?` and `hasFunctionCalls?` to `IRConditional`, mirroring `IRExpression`.
- `packages/jsx/src/jsx-to-ir.ts`:
  - `transformConditional` (line 966) — widen `needsSlot` with the two AST-derived flags; store both on the returned node.
  - `transformLogicalAnd` (line 992) — same.
  - `transformNullishCoalescing` (line 1044) — hoist the flag computation so both the `IRConditional` slotId decision and the `whenTrue` `IRExpression` share the same values (previously only the whenTrue had them).
- `packages/jsx/src/ir-to-client-js/collect-elements.ts`:
  - `case 'conditional'` in `collectElements` — widen the gate.
  - `collectBranchConditionals` — widen symmetrically for nested branches.
- `packages/jsx/src/__tests__/conditional-fallback-wrap.test.ts` — 6 regression cases mirroring `runtime-fallback-wrap.test.ts` from #939:
  1. Known signal in ternary still wraps (regression guard).
  2. Unrecognised call in ternary now wraps (new behaviour).
  3. Logical-AND with method-call condition wraps.
  4. Static literal-derived condition stays un-wrapped.
  5. Bare identifier in nullish coalescing stays un-wrapped.
  6. Nested ternaries with unrecognised calls wrap at both levels.

## Fixture sweep

Baseline: `site/ui/dist/components/**/*.js` on `main` (includes #939 + #940). After rebuild, **6 of 181 distinct components differ (~3.3 %)** — well under the 30 % rollback threshold from #937. All diffs trace to a single source change: the Calendar component gaining a correct \`insert()\` for \`{numMonths() >= 2 && <SecondMonth />}\`, which was previously frozen to its SSR-time branch.

Affected bundles (all Calendar-bundle related):
- `calendar-demo`, `calendar-playground`, `calendar-usage-demo`
- `date-picker-demo`, `date-picker-playground` (wrap Calendar)
- `studio-canvas` (embeds Calendar)

## Scope (non-goals)

- Did **not** redesign the branch-template mechanism. Only the gate is widened.
- Did **not** widen `collectLoopChildConditionals` (loop-body conditionals) — tracked in #943.
- Did **not** deduplicate the three near-identical gates into a helper. Structural cleanup out of scope.
- Did **not** touch `IRIfStatement` (component-level \`if (cond) return <X/>\` chain) — a separate code path not relevant to JSX-inline silent-drops.

## Test plan

- [x] `bun test packages/jsx` — 674 tests pass (6 new, 668 pre-existing).
- [x] `bun test packages/{dom,cli,client,hono,go-template,adapter-tests}` — 802 tests pass.
- [x] Fixture sweep: 6 / 181 components gain a correct new `insert()`.

## References

- Sibling follow-ups: #940 (attributes, merged), #942 (child-component prop bindings), #943 (loops)

### DIFF
--- a/packages/jsx/src/__tests__/conditional-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/conditional-fallback-wrap.test.ts
@@ -1,0 +1,180 @@
+/**
+ * Regression tests for #941: Solid-style wrap-by-default fallback for JSX
+ * conditional expressions used as children — `{cond ? a : b}`,
+ * `{cond && b}`, `{a ?? b}`, `{a || b}`. Follow-up to #937 (architecture)
+ * and #939 (text interpolation pilot).
+ *
+ * Before this change, `jsx-to-ir.ts` gated `IRConditional.slotId` on
+ * `isReactiveExpression(...)` (an allow-list of known signal getters,
+ * memos, and props). Any condition the analyzer couldn't statically prove
+ * reactive — e.g. `{isVisible(x) ? <A/> : <B/>}` where `isVisible` is an
+ * imported helper — was allocated no slotId. At SSR the branch was chosen
+ * once based on the condition's initial value; the client never
+ * re-evaluated it, so the rendered branch stayed frozen.
+ *
+ * The fix computes `exprHasFunctionCalls` / `exprCallsReactiveGetters` on
+ * the condition AST at each of the three sites and includes those flags
+ * in both the `needsSlot` decision and the `IRConditional` node. The
+ * collector in `collect-elements.ts` then wraps on the widened gate.
+ * Over-wrap is harmless; under-wrap is the silent-drop bug we're closing.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { compileJSXSync } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
+
+const adapter = new TestAdapter()
+
+function getClientJs(source: string, filename: string): string {
+  const result = compileJSXSync(source, filename, { adapter })
+  expect(result.errors.filter(e => e.severity === 'error')).toHaveLength(0)
+  const clientJs = result.files.find(f => f.type === 'clientJs')
+  expect(clientJs).toBeDefined()
+  return clientJs!.content
+}
+
+describe('Solid-style wrap-by-default fallback for conditionals (#941)', () => {
+  test('known signal in ternary still wraps (regression guard)', () => {
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function Gate() {
+        const [count, setCount] = createSignal(0)
+        return (
+          <button onClick={() => setCount(c => c + 1)}>
+            {count() > 0 ? <span>pos</span> : <span>zero</span>}
+          </button>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Gate.tsx')
+    // A reactive ternary emits an insert(...) call for branch switching.
+    expect(clientJs).toContain('insert')
+    expect(clientJs).toContain('count()')
+  })
+
+  test('unrecognised call in ternary now wraps in insert (new behaviour)', () => {
+    // `isVisible` is an imported helper the analyzer can't prove reactive.
+    // Before the fix, no slotId was allocated, and the branch stayed frozen
+    // at its SSR value. With wrap-by-default the call shape alone allocates
+    // a slotId and the collector emits an insert() for branch switching.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { isVisible } from './visibility'
+
+      export function Gate() {
+        const [, setFoo] = createSignal(0)
+        const x = 1
+        return (
+          <button onClick={() => setFoo(1)}>
+            {isVisible(x) ? <span>yes</span> : <span>no</span>}
+          </button>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Gate.tsx')
+    expect(clientJs).toContain('insert')
+    expect(clientJs).toContain('isVisible(')
+  })
+
+  test('logical-AND with method-call condition wraps', () => {
+    // `flags.enabled()` is a method call — neither signal, memo, nor prop.
+    // Under the old gate this was dropped; under wrap-by-default the
+    // function-call shape alone triggers the fallback.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function Banner() {
+        const [, setFoo] = createSignal(0)
+        const flags = { enabled: () => true }
+        return (
+          <div onClick={() => setFoo(1)}>
+            {flags.enabled() && <span>on</span>}
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Banner.tsx')
+    expect(clientJs).toContain('insert')
+    expect(clientJs).toContain('flags.enabled()')
+  })
+
+  test('static literal-derived condition stays un-wrapped', () => {
+    // `staticValue` is a bare local const with no function calls; neither
+    // the old gate nor the new fallback fires. No insert() — the branch
+    // baked into SSR output stays as-is on the client.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function Banner() {
+        const [, setFoo] = createSignal(0)
+        const staticValue = true
+        return (
+          <div onClick={() => setFoo(1)}>
+            {staticValue ? <span>a</span> : <span>b</span>}
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Banner.tsx')
+    expect(clientJs).not.toContain('insert')
+  })
+
+  test('bare identifier in nullish coalescing stays un-wrapped', () => {
+    // `a` is a bare local const with no calls. `a ?? <Fallback/>` has
+    // nothing to re-evaluate on the client. No insert() emitted.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function Message() {
+        const [, setFoo] = createSignal(0)
+        const a = 'hello'
+        return (
+          <div onClick={() => setFoo(1)}>
+            {a ?? <span>fallback</span>}
+          </div>
+        )
+      }
+    `
+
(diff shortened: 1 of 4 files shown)


kfly8 (commented):
## Overview

Third application of the #937 wrap-by-default pattern, this time to `IRConditional` — covering `{cond ? a : b}`, `{cond && b}`, and `{a ?? b}` / `{a || b}`. +233/-12, 4 source files + 1 new test file, all CI green.

- `types.ts` — `IRConditional` gains `callsReactiveGetters?` / `hasFunctionCalls?`, mirroring `IRExpression`.
- `jsx-to-ir.ts::transformConditional` / `transformLogicalAnd` / `transformNullishCoalescing` — same three-line widening at each site (compute `callsReactive` + `hasCalls`, extend `needsSlot`, attach both flags to the returned node).
- `collect-elements.ts::collectElements` `case 'conditional'` — widens the top-level gate to `reactive || callsReactiveGetters || hasFunctionCalls`.
- `collect-elements.ts::collectBranchConditionals` — widens the nested-branch gate symmetrically (important — easy to miss).
- `conditional-fallback-wrap.test.ts` — 6 cases covering all three operators + nested + two negative cases.

## Strengths

- **Pattern template from #939 / #940 applied cleanly.** Same shape of change at four call sites; each site's comment back-references #937/#939 and explains the trade-off briefly.
- **`collectBranchConditionals` widening in sync with the top-level gate.** This is critical — if the nested gate had been left at `n.reactive && n.slotId`, a nested `{outer() ? (inner() ? …) : …}` would silently drop the inner on-client-update. The PR catches this explicitly. The code comment even cross-references the top-level gate for future auditors.
- **`transformNullishCoalescing` drive-by consistency fix.** The previous code computed `exprCallsReactiveGetters(node.left)` and `exprHasFunctionCalls(node.left)` for the `whenTrue` `IRExpression` but *not* for the `IRConditional` slot decision — a latent bug where `{helper() ?? <F/>}` would mark the expression reactive but allocate no slot. Hoisting the computation fixes both places at once. Good cleanup.
- **Fixture sweep is concrete.** 6 / 181 components differ (~3.3%) and every diff traces to a single real source change (`numMonths() >= 2 && <SecondMonth />` in Calendar). Under the 30% rollback threshold from #937, and the uplift is behavioural (frozen branch → correctly reactive), not noise.
- **Scope discipline.** Non-goals are enumerated (loop-body conditionals → #943; `IRIfStatement` component-level early returns → separate; DRY consolidation → future cleanup). Prevents scope creep.
- **E2E + unit tests all green on CI.**

## Concerns (non-blocking)

### 1. Test 6 uses `toBeGreaterThanOrEqual(2)` instead of `toBe(2)`

`conditional-fallback-wrap.test.ts` final test:
```ts
const insertCount = (clientJs.match(/\binsert\s*\(/g) ?? []).length
expect(insertCount).toBeGreaterThanOrEqual(2)
```
The intent is "both outer and inner conditionals emit `insert()`" — `toBe(2)` enforces this precisely and would also catch a future regression that consolidates nested inserts in a way we don't expect. `>= 2` lets silent bloat past. If there's a reason more than 2 inserts could legitimately appear for this source, a comment explaining why would help.

### 2. Test 4 negative assertion is broad

```ts
expect(clientJs).not.toContain('insert')
```
A future unrelated change that adds an `insert()` for a different concern in the same component would flip this red, even when the `staticValue` branch still correctly stays un-wrapped. A slot-scoped assertion (e.g. look for `insert(__scope, 's0', …)` specifically for the conditional's slotId, or count `insert(` for the conditional's slotId) would be tighter. Same pattern concern also applies to test 5.

### 3. Four-site DRY is now tempting

After #939 (text), #940 (attr), and this PR (3× conditional operators), the exact three-line sequence

```ts
const callsReactive = exprCallsReactiveGetters(expr, ctx)
const hasCalls = exprHasFunctionCalls(expr)
const needsSlot = reactive || loopParamReactive || callsReactive || hasCalls
```

appears **five** times in `jsx-to-ir.ts` alone, plus twice in `collect-elements.ts`. The PR explicitly defers this, which is fine for *this* PR, but when #942 (child-component prop bindings) or #943 (loops) adds a 6th / 7th site it will be cheaper to extract a helper at that point than to keep copying. Worth a note in the follow-up issues.

### 4. `loopParamReactive` is now partially redundant

`loopParamReactive = !reactive && referencesLoopParam(condition, ctx)` still runs at all three sites. In practice any expression that references a loop param AND is not already `reactive` almost certainly has either a property access or a call (`item.foo`, `item.id`, `process(item)`), which `hasCalls` / `callsReactive` would already catch for call shapes. Not a bug — just an observation that the condition is now partially subsumed. Cleanup territory, not this PR's problem.

- **Correctness**: Closes the conditional-shape silent-drop gap uncovered by the Calendar component. The `collectBranchConditionals` sync is the main correctness win beyond the #939 pattern.
- **Performance**: 6 additional `insert(...)` call sites emitted across 6 components, roughly one closure each at hydration time. Negligible.
- **Security**: Internal compiler change, no user input surface.

## Verdict

**Ready to merge.** Concerns 1 and 2 are test-tightness nits worth addressing opportunistically — neither blocks. The `transformNullishCoalescing` hoist is a nice bonus correctness fix on top of the #941 widening, and `collectBranchConditionals` being updated in tandem is exactly right.

Pattern is now proven across three node types (expression / attribute / conditional). After #942 and #943 land the DRY consolidation becomes worthwhile; flagging that for a separate clean-up PR.

PR #947, opened 2026-04-21
fix(compiler): wrap-by-default fallback for child-component prop bindings (#942)

### LINKED ISSUE #942: compiler: extend wrap-by-default fallback to child-component prop bindings (#937 follow-up)

kfly8:
## Summary

Follow-up to #937. Carries the Solid-style wrap-by-default fallback landed in #939 (text interpolation), #940 (attributes), and #941 (conditionals) to **prop bindings on child components** — `<Card title={expr} />`, `<Badge variant={expr} />`, etc. Closes #942.

Before this change, the `reactiveChildProps` gate in `collect-elements.ts` fired on `hasPropsRef || needsEffectWrapper(...)`. The latter is an allow-list that only recognises known signal getters, memos, and prop parameter names. Any child-component prop expression the analyzer couldn't statically prove reactive — e.g. `<Card title={formatTitle(page)} />` where `formatTitle` is imported — was silently dropped. SSR evaluated the prop once; the client never re-evaluated it, so the child component's DOM stayed frozen.

## Changes

`packages/jsx/src/ir-to-client-js/collect-elements.ts`:

The gate now ORs three conditions:

- `hasPropsRef` — direct `props.xxx` access (existing path).
- `hasReactiveExpr` — `needsEffectWrapper`'s allow-list match (existing path).
- `hasAnyCall` — **new regex fallback**: `/\b\w+\s*\(/` on the expanded value, applied **after stripping single- and double-quoted string literals**. The strip avoids false positives on object-literal prop values like `{ color: 'hsl(221 83% 53%)' }` — the `hsl(` inside the literal would otherwise trigger a spurious wrap. Template literals are left intact because `${calls()}` interpolations are live code.

Over-wrap (extra createEffect that subscribes to nothing) is harmless: one closure at init time, sets the child component DOM attribute once. Under-wrap is the silent-drop bug this closes.

`packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts` — 6 regression cases mirroring `runtime-fallback-wrap.test.ts` from #939:

1. Known signal in `<Card title={count()} />` still wraps (regression guard).
2. Unrecognised call in `<Card title={formatTitle(page)} />` now wraps (new behaviour).
3. Pure call `<Badge variant={variantFor(status)} />` with no recognised reactive source wraps (harmless over-wrap).
4. String literal prop `<Card title=\"Static\" />` stays un-wrapped.
5. Bare identifier `<Card title={label} />` without calls or `props.` stays un-wrapped.
6. `<Card title={props.title} />` — existing `hasPropsRef` path still wraps.

## Fixture sweep

Baseline: `site/ui/dist/components/**/*.js` on `main` (post #939/#940/#941). After rebuild, **52 / 181 distinct components differ (~28.7 %)**, just under the 30 % rollback threshold from #937.

Bulk of the diffs trace to a single source pattern: \`<CopyButton code={plainJsxTree(tree())} />\` in every playground component. \`tree()\` is a signal but the outer \`plainJsxTree\` call blocked \`needsEffectWrapper\` from recognising the signal read — a genuine silent-drop fix now reactively refreshing the playground's copy-to-clipboard source.

The remaining diffs add harmless over-wraps for inline arrow-function props (\`(v) => v.slice(0, 3)\`) and constructor calls (\`new Date()\`). The child component receives each prop stably via \`initChild()\`; the extra DOM \`setAttribute\` on the component element is a one-shot at hydration and matches the SSR-rendered attribute.

Initial attempt without the string-literal strip produced 60 / 181 = 33 %, just over threshold, because chart demos use inline \`config={{ color: 'hsl(221 83% 53%)' }}\` — the \`hsl(\` inside the string was matching the naive regex. Stripping single/double-quoted strings before the regex drops those spurious matches.

## Scope (non-goals)

- Did **not** change `forwardProps()` semantics or spread-prop reactivity.
- Did **not** change the JSX-children getter path.
- Did **not** deduplicate the now four-site `hasAnyCall` regex into a shared helper — same DRY discussion as #941, worth revisiting after #943 lands.

## Test plan

- [x] `bun test packages/jsx` — 680 tests pass (6 new, 674 pre-existing).
- [x] `bun test packages/{dom,cli,client,hono,go-template,adapter-tests}` — 802 tests pass.
- [x] Fixture sweep: 52 / 181 components gain correct new \`createEffect\` for child-component prop updates.

## References

- Sibling follow-ups: #940 (attributes, merged), #941 (conditionals, merged), #943 (loops)

### DIFF
--- a/packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts
@@ -1,0 +1,189 @@
+/**
+ * Regression tests for #942: Solid-style wrap-by-default fallback for
+ * prop bindings on child components — `<Card title={expr} />`. Follow-up
+ * to #937 (architecture) and #939 (text interpolation pilot).
+ *
+ * Before this change, `collect-elements.ts` gated `reactiveChildProps`
+ * on `hasPropsRef || needsEffectWrapper(...)`. The latter is an
+ * allow-list that only recognises known signal getters, memos, and
+ * prop parameter names. Any child-component prop expression the
+ * analyzer couldn't prove reactive — e.g. `<Card title={formatTitle(
+ * page)} />` where `formatTitle` is imported — was silently dropped.
+ * At SSR the prop was evaluated once; the child component's DOM stayed
+ * frozen at its initial render on the client.
+ *
+ * The fix adds a third branch to the OR gate: `/\b\w+\s*\(/.test(
+ * expandedValue)`. Over-wrap (extra createEffect that subscribes to
+ * nothing) is harmless — one closure at init time. Under-wrap is the
+ * silent-drop bug we're closing.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { compileJSXSync } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
+
+const adapter = new TestAdapter()
+
+function getClientJs(source: string, filename: string): string {
+  const result = compileJSXSync(source, filename, { adapter })
+  expect(result.errors.filter(e => e.severity === 'error')).toHaveLength(0)
+  const clientJs = result.files.find(f => f.type === 'clientJs')
+  expect(clientJs).toBeDefined()
+  return clientJs!.content
+}
+
+describe('Solid-style wrap-by-default fallback for child-component props (#942)', () => {
+  test('known signal getter in child prop still wraps (regression guard)', () => {
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { Card } from './Card'
+
+      export function Dashboard() {
+        const [count, setCount] = createSignal(0)
+        return (
+          <div onClick={() => setCount(c => c + 1)}>
+            <Card title={count()} />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Dashboard.tsx')
+    // Reactive child props emit a `// Reactive child component props`
+    // block containing a createEffect that refreshes the child element's
+    // attributes when the signal changes.
+    expect(clientJs).toContain('Reactive child component props')
+    expect(clientJs).toContain('count()')
+  })
+
+  test('unrecognised call in child prop now wraps (new behaviour)', () => {
+    // `formatTitle` is an imported helper the analyzer can't prove
+    // reactive; `page` is a local const. Before the fix, this was
+    // silently dropped from reactiveChildProps. With wrap-by-default the
+    // function-call shape alone triggers the fallback.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { formatTitle } from './format'
+      import { Card } from './Card'
+
+      export function Dashboard() {
+        const [, setFoo] = createSignal(0)
+        const page = 'home'
+        return (
+          <div onClick={() => setFoo(1)}>
+            <Card title={formatTitle(page)} />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Dashboard.tsx')
+    expect(clientJs).toContain('Reactive child component props')
+    expect(clientJs).toContain('formatTitle(')
+  })
+
+  test('call with no recognised reactive source wraps (harmless over-wrap)', () => {
+    // `variantFor(status)` is a pure call: neither signal, memo, nor
+    // prop appears. Under wrap-by-default we emit a createEffect that
+    // subscribes to nothing and runs exactly once. Cheaper than the
+    // silent-drop risk.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { variantFor } from './variants'
+      import { Badge } from './Badge'
+
+      export function Label() {
+        const [, setFoo] = createSignal(0)
+        const status = 'active'
+        return (
+          <div onClick={() => setFoo(1)}>
+            <Badge variant={variantFor(status)} />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Label.tsx')
+    expect(clientJs).toContain('Reactive child component props')
+    expect(clientJs).toContain('variantFor(')
+  })
+
+  test('string literal prop stays un-wrapped', () => {
+    // Static literal props have attr.dynamic === false and take the
+    // `prop.isLiteral` branch of the if/else — they never reach the
+    // reactiveChildProps gate at all.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { Card } from './Card'
+
+      export function Dashboard() {
+        const [, setFoo] = createSignal(0)
+        return (
+          <div onClick={() => setFoo(1)}>
+            <Card title="Static" />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Dashboard.tsx')
+    expect(clientJs).not.toContain('Reactive child component props')
+  })
+
+  test('bare identifier child prop (no calls, no props ref) stays un-wrapped', () => {
+    // `label` is a local const with no function calls and no `props.`
+    // reference. None of the three OR branches fires, so no
+    // createEffect is emitted for the child prop binding.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { Card } from './Card'
+
+      export function Dashboard() {
+        const [, setFoo] = createSignal(0)
+        const label = 'hi'
+        return (
+          <div onClick={() => setFoo(1)}>
+            <Card title={label} />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Dashboard.tsx')
+    expect(clientJs).not.toContain('Reactive child component props')
+  })
+
+  test('props.xxx access in child prop still wraps (hasPropsRef regression guard)', () => {
+    // The existing `hasPropsRef` branch remains — `props.title` is a
+    // direct prop reference even though it contains no function call.
+    // Ensures the widening didn't accidentally displace the existing
+    // path.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { Card } from './Card'
+
+      export function Dashboard(props: { title: string }) {
+        const [, setFoo] = createSignal(0)
+        return (
+          <div onClick={() => setFoo(1)}>
+            <Card title={props.title} />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Dashboard.tsx')
+    expect(clientJs).toContain('Reactive child component props')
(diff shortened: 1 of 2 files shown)


kfly8 (commented):
## Overview

Fourth application of the #937 wrap-by-default pattern, now on child-component prop bindings (`<Card title={expr} />`). +212/-1, one source file changed, one test file added, all CI green. Closes #942.

Adds a third OR branch to the `reactiveChildProps` gate in `collect-elements.ts`:

- `hasPropsRef` — existing `props.xxx` direct access.
- `hasReactiveExpr` — existing `needsEffectWrapper` allow-list.
- `hasAnyCall` — **new**: regex `/\b\w+\s*\(/` applied after stripping single- and double-quoted string literals. Template literals are intentionally left intact.

## Strengths

- **Real bug uncovered and fixed.** The fixture sweep surfaced `<CopyButton code={plainJsxTree(tree())} />` across every playground component — `tree()` is a signal but the outer `plainJsxTree()` wrapper hid it from `needsEffectWrapper`'s allow-list. Before this PR the clipboard copy button stayed frozen at the SSR value. Now it tracks the signal correctly.
- **#937 rollback criterion applied in practice.** Author hit 33% (60/181) on the first try, diagnosed the false positive (`hsl(221 83% 53%)` inside object-literal string values triggering the naive regex), narrowed the heuristic by stripping quoted strings, and re-measured to 28.7% (52/181) just under threshold. That's exactly the discipline the threshold was designed to provoke — documented in the PR description.
- **Comment explains the intent and the trade-off.** Specifically calls out that template literals are NOT stripped because `${calls()}` is live code — avoids a future maintainer deleting that asymmetry.
- **Test matrix matches the proven #939/#941 shape** — known reactive, unrecognised call, pure call over-wrap, static literal, bare identifier, `props.xxx` regression guard.
- **Scope disciplined.** Non-goals enumerated: `forwardProps` semantics, JSX children getter path, DRY consolidation.

## Concerns (non-blocking but worth raising)

### 1. Heuristic implementation diverges from #939/#941

The three prior PRs widened the gate using AST-derived `hasFunctionCalls` / `callsReactiveGetters` flags stored on `IRExpression` and `IRConditional`. This PR uses a source-text regex instead. The PR justifies the deviation — "Phase 2 has no AST access here" — and that's correct for the current IR shape, but it means the heuristic is implemented **two different ways for the same conceptual rule**:

- `IRExpression` / `IRConditional`: flag-on-IR.
- Child-component prop value: regex on the expanded source string.

Debugging \"why did this expression wrap?\" a month from now requires remembering which path the expression took. Two options for consolidation, ideally before #943 adds a 5th site:

- **Option A (cheaper):** extend the child-prop IR shape to carry the two flags like `IRExpression` does. Phase 1 sees the AST — compute and store, Phase 2 just reads.
- **Option B (deferred):** track as a follow-up once the pattern has settled across all 4–5 sites.

Not blocking this PR. Worth an explicit comment in #943 / the DRY consolidation issue so it isn't forgotten.

### 2. String-literal strip has a corner-case glitch (low impact)

The strip runs single-quote regex first, then double-quote. A source expression like `"It's Bob's birthday"` (double-quoted containing unescaped `'`) confuses the single-quote pass — it consumes `'s Bob's` as if it were a `'...'` literal, leaving `"It''s birthday"`. Then the double-quote pass strips `"It''s birthday"` to `""`. End result: no call detected — which is still *correct* for that input (there are no calls), so the bug is latent. But the approach is fragile:

- Swap the order (strip double-quoted first, then single-quoted) and the failure mode flips.
- Adding `\`template with '${x}' inside\`` to the corpus could surprise future readers.

A small robustness improvement: combine the two strips into one alternation regex that understands both quote kinds:

```ts
const withoutStringLiterals = expandedValue.replace(
  /"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'/g,
  ''
)
```

Single pass, no order dependency. Low priority — the current code is correct for the fixture sweep — but since the same regex will likely be copied if #943 picks up a similar path, cleaner source to copy is nice.

### 3. Variable naming

`const hasAnyCall = /\b\w+\s*\(/.test(withoutStringLiterals)` reads naturally, but the transformation happened one line above. Something like `hasCallOutsideString` / `hasUnstrippedCall` would make the intent obvious without having to read the preceding `withoutStringLiterals` declaration. Nit.

### 4. `hasPropsRef = expandedValue.includes('props.')` is pre-existing, but worth flagging

Not introduced by this PR — but since the surrounding gate is now being audited: `.includes('props.')` matches any token containing that substring, including a user-named `myprops.foo` or an object literal key. False positives on this branch are harmless (triggers extra wrap) but worth noting if someone widens the gate further.

- **Correctness**: Closes the child-prop silent-drop class. The `plainJsxTree(tree())` fixture hit is the concrete evidence this was a real bug. No regression risk seen in CI.
- **Performance**: 52 components gain one init-time closure per newly-collected child prop. Negligible.
- **Security**: Regex on user source is bounded — no catastrophic backtracking risk for `/\b\w+\s*\(/` (linear in input length). The string-strip regex has the classic `(?:[^'\\]|\\.)*` shape which can backtrack in pathological inputs, but expanded prop values are short. Acceptable.

## Verdict

**Ready to merge.** The real-bug fix for `CopyButton` justifies the PR on its own, and the rollback-threshold discipline is exactly what #937 asked for. Concern #1 (AST-vs-regex divergence) is the strongest case for a follow-up — worth deciding in #943 or a dedicated DRY PR whether to bring child-props onto the IR-flag path.

Pattern is now applied at 4 of the 4 planned #937 rollout sites (text, attr, conditional, child-prop). Only #943 (loops) left; after that, consolidation is the logical next step.

kfly8 (commented):
## Re-review (round 2)

One follow-up commit since the first review: `bce6d455 refactor(compiler): tighten child-prop string-strip heuristic` — +11 / -7 on a single file, CI re-run green.

### Addresses

|-------|--------|
| #2 String-literal strip has order dependency | **Fixed.** Collapsed the two `.replace(...)` passes into one alternation regex `/"(?:[^"\\]|\\.)*"\|'(?:[^'\\]|\\.)*'/g` — matches the suggestion exactly. Also cleaner: replaces with `''` (empty) instead of `"'"` / `""` stub strings, so the `hasCallOutsideString` regex doesn't need to distinguish. Comment updated with the `"It's fine"` mixed-quote example as a pin. |
| #3 `hasAnyCall` naming | **Fixed.** Renamed to `hasCallOutsideString` — reads naturally and makes the preceding `withoutStringLiterals` step visibly part of the pipeline. |
| #1 AST-vs-regex divergence across 4 sites | Unchanged — out of scope for this PR. Correctly deferred to the DRY / #943 discussion. |
| #4 `hasPropsRef = ...includes('props.')` pre-existing brittleness | Unchanged — was flagged as pre-existing, not in scope. |

### Delta verification

- New regex is linear in input: `/"(?:[^"\\]|\\.)*"|'(?:[^'\\]|\\.)*'/g` — same `[^x\\]|\\.` shape as before, single alternation, no nested quantifier risk. Safe.
- Empty replacement (`''`) vs the old stubs (`"''"` / `'""'`) is slightly tighter: previously `url('foo')` became `url('')` (stub string retained), now becomes `url()` — still matches the outer `/\b\w+\s*\(/` regex identically. Equivalent for call detection, cleaner downstream.
- All 680 existing unit tests + 6 new child-prop tests pass on CI. E2E (site-ui / hono / echo) all green.

### Verdict

**Still ready to merge**, now with both nits cleanly addressed. The refactor is small and unambiguous; no new risk introduced. Concern #1 (AST-vs-regex unification) remains the right topic for a follow-up after #943 — flag it there.

Issue #944, opened 2026-04-21
debug: surface wrap-by-default fallback expressions in `barefoot inspect` / `why-update`


kfly8:
Follow-up to #937 / #939. Once the wrap-by-default fallback lands across #940-#943, the compiler will emit `createEffect` for expressions the static analysis can't prove reactive. These are harmless from a correctness standpoint but invisible to the user — and the existing debug tooling (`barefoot inspect`, `barefoot why-update`) will also not list them, because it uses the old pre-#937 gate.

This issue tracks making the fallback-wrapped expressions visible in the inspection CLI, so users who want to optimise can find them without resorting to diffing compiled output.

## Why CLI, not `console.warn`

Decided explicitly in #937 and [`project_priorities` memory](https://github.com/barefootjs/barefootjs/issues/937): dev-time warnings are ignored or cause noise. The visibility channel for \"which expressions got auto-wrapped\" is opt-in CLI tooling, matching the existing `inspect` / `why-update` / `test --debug` pattern.

## The current gap

`packages/jsx/src/debug.ts::collectDomBindings` uses the same `node.reactive`-only gate the emitter used before #939:

- line 538 (text interpolation):
  ```ts
  case 'expression': {
    if (node.reactive && node.slotId) { ... bindings.push({ ..., type: 'text' }) }
  }
  ```
- line 551 (conditional): identical pattern
- line 503 (attribute): `attr.dynamic && deps.length > 0`

After #939, `collect-elements.ts` also collects expressions with `callsReactiveGetters || hasFunctionCalls` — those hit `createEffect` at runtime, but `collectDomBindings` never records them, so `barefoot inspect` / `barefoot why-update` under-report the actual reactive footprint.

### 1. Widen `collectDomBindings` gate to match the emitter

Each `case` in `packages/jsx/src/debug.ts::collectDomBindings` (lines 494-600ish) learns the same `shouldWrap = reactive || callsReactiveGetters || hasFunctionCalls` gate that `collect-elements.ts` now uses. Follow-ups in #940/#941/#943 will need the same widening on their respective AST nodes.

### 2. Add a classification field to `DomBinding`

```ts
// packages/jsx/src/debug.ts
export interface DomBinding {
  kind: 'dom'
  label: string
  slotId: string
  deps: string[]
  type: 'text' | 'event' | 'conditional' | 'loop' | 'attribute'
  /**
   * How this binding got classified:
[...]

PR #948, opened 2026-04-21
fix(compiler): wrap-by-default fallback for .map() loops (#943)

### LINKED ISSUE #943: compiler: extend wrap-by-default fallback to loops (#937 follow-up)

kfly8:
## Summary

Final follow-up in the #937 rollout, completing coverage after #939 (text interpolation), #940 (attributes), #941 (conditionals), and #942 (child-component prop bindings). Closes #943.

Widens the `isStaticArray` decision in `transformMapCall` so any `.map()` whose array source AST contains a function call forces reconciliation — even when `isSignalOrMemoArray` can't recognise the callee as a signal or memo. `getItems().map(...)` where `getItems` is an imported helper previously silent-dropped into the static-render path; the SSR list was frozen on the client. Under wrap-by-default the call shape alone drives reconcileList wiring.

## Changes

`packages/jsx/src/jsx-to-ir.ts::transformMapCall`:

- Track `arrayExpr: ts.Expression` alongside the `array` string. `setArray` updates both; the two remaining direct `array = ctx.getJS(...)` sites (simple `filter().map()` and no-chain fallback) now also assign `arrayExpr`. Initial value is `mapSource`, matching the fallback path.
- `isStaticArray = !isSignalOrMemoArray(array, ctx) && !exprHasFunctionCalls(arrayExpr)`. Unlike #940 / #942, this widening uses **AST** rather than a string regex — the AST node is already in scope here, and `exprHasFunctionCalls` is the same helper the text-interpolation pilot (#939) and conditional follow-up (#941) use. Closes the previous reviewer concern (#947 review Concern 1) about AST-vs-regex divergence across the rollout sites, for this one case at least.

`packages/jsx/src/__tests__/loop-fallback-wrap.test.ts` — 7 regression cases:

1. `{items().map(...)}` where `items` is a known signal still reconciles (regression guard).
2. `{getItems().map(...)}` where `getItems` is an imported helper now reconciles (new behaviour).
3. `{[1, 2, 3].map(...)}` inline literal stays static (optimisation preserved).
4. `{props.items.map(...)}` static prop (no calls) stays static.
5. `{computeItems().filter(t => t.done).map(...)}` — chain with extracted filter, array normalises to `computeItems()`; AST still has call, reconciles.
6. Nested: `{outer().map(o => <li><ul>{o.children.map(c => <li>{c}</li>)}</ul></li>)}` — outer reconciles.
7. Composite: `{listOf().map(x => <Item key={x.id} label={x.label} />)}` — createComponent / reconcile path handles unrecognised-call arrays correctly.

## Fixture sweep

Baseline: `site/ui/dist/components/**/*.js` on `main` (post #939/#940/#941/#942).

| Metric | Value | Threshold |
|--------|-------|-----------|
| Component bundles changed | **4 / 181 (~2.2 %)** | 30 % |
| Total JS byte delta | **+2326 (~0.009 %)** | 5 % |

Both metrics are well under the rollback thresholds. The four diffs:

- \`dashboard-demo\` — \`orders.slice(0, 3).map(...)\`
- \`pie-chart-demo\` — \`Object.entries(chartConfig).map(...)\`
- \`scroll-area-demo\` — \`Array.from({ length: 20 }).map(...)\`
- \`table-demo\` — \`invoices.slice(0, 4).map(...)\`

Each gains a \`mapArray\` wrap of ~580 bytes. The underlying collections are currently non-reactive constants, so the wrap is a harmless future-proof: if any of these is ever refactored to a signal, the chain will track without needing a compiler change.

## Scope (non-goals)

- Did **not** replace the `isStaticArray` optimisation — it still earns its keep for genuinely static arrays (literals, bare-identifier props with no chain). Only the detection of the dynamic case is widened.
- Did **not** redesign filter / sort chain handling.
- Did **not** add opt-out syntax. With all four wrap-by-default follow-ups now landed, consolidation — including the DRY discussion from the #947 round 1 review (Concern 1) — becomes the logical next step.

## Test plan

- [x] `bun test packages/jsx` — 687 tests pass (7 new, 680 pre-existing).
- [x] `bun test packages/adapter-tests` — 184 tests pass (no fixture byte regressions).
- [x] `bun test packages/{dom,cli,client,hono,go-template}` — 625 tests pass.
- [x] Fixture sweep: 4 / 181 components (~2.2 %), +2326 bytes (~0.009 %).

## References

- Sibling follow-ups: #940 (attributes, merged), #941 (conditionals, merged), #942 (child-component prop bindings, merged)

### DIFF
--- a/packages/jsx/src/__tests__/loop-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/loop-fallback-wrap.test.ts
@@ -1,0 +1,210 @@
+/**
+ * Regression tests for #943: Solid-style wrap-by-default fallback for
+ * `.map()` loops. Follow-up to #937 (architecture) and #939 (text
+ * interpolation pilot).
+ *
+ * Before this change, `jsx-to-ir.ts` decided whether a loop needed
+ * reconcileList via `isSignalOrMemoArray(array, ctx)` — a regex
+ * allow-list of known signal getters and memos. Any array source the
+ * analyzer couldn't prove reactive — e.g. `{getItems().map(...)}`
+ * where `getItems` is an imported helper — flowed into the
+ * `isStaticArray` branch. SSR rendered the list once; the client
+ * never reconciled, so the list stayed frozen.
+ *
+ * The fix tracks the AST node alongside the `array` string and widens
+ * the decision: if the analyzer can't prove the array reactive but the
+ * AST contains a function call, force reconciliation. `exprHasFunctionCalls`
+ * is the same AST helper used in the text-interpolation pilot (#939)
+ * and the conditional follow-up (#941), so this loop widening closes
+ * the divergence gap between the allow-list path and the AST-flag path.
+ *
+ * Over-reconciliation of a pure-call array has a real cost (reconcileList
+ * is not a cheap primitive), so the fixture sweep in the accompanying
+ * PR measures byte delta too; but the silent-drop bug this closes is
+ * the expensive one — a frozen client list is a visible correctness
+ * regression.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { compileJSXSync } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
+
+const adapter = new TestAdapter()
+
+function getClientJs(source: string, filename: string): string {
+  const result = compileJSXSync(source, filename, { adapter })
+  expect(result.errors.filter(e => e.severity === 'error')).toHaveLength(0)
+  const clientJs = result.files.find(f => f.type === 'clientJs')
+  expect(clientJs).toBeDefined()
+  return clientJs!.content
+}
+
+describe('Solid-style wrap-by-default fallback for loops (#943)', () => {
+  test('known signal array still reconciles (regression guard)', () => {
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function List() {
+        const [items, setItems] = createSignal<string[]>([])
+        return (
+          <ul onClick={() => setItems(prev => [...prev, 'x'])}>
+            {items().map(item => <li key={item}>{item}</li>)}
+          </ul>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'List.tsx')
+    // Dynamic loops emit reconcileList / reconcileElements via mapArray.
+    expect(clientJs).toMatch(/mapArray|reconcile/)
+    expect(clientJs).toContain('items()')
+  })
+
+  test('unrecognised call array now reconciles (new behaviour)', () => {
+    // `getItems` is an imported helper — the analyzer can't prove the
+    // return value reactive. Before the fix the loop baked its SSR
+    // output and never updated. With wrap-by-default the call shape on
+    // the array expression forces reconciliation.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { getItems } from './data'
+
+      export function List() {
+        const [, setFoo] = createSignal(0)
+        return (
+          <ul onClick={() => setFoo(1)}>
+            {getItems().map(item => <li key={item}>{item}</li>)}
+          </ul>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'List.tsx')
+    expect(clientJs).toMatch(/mapArray|reconcile/)
+    expect(clientJs).toContain('getItems()')
+  })
+
+  test('inline literal array stays static (optimisation preserved)', () => {
+    // `[1, 2, 3]` is an ArrayLiteralExpression with no calls — the
+    // existing static-render path is the right call. No reconcileList /
+    // mapArray should appear.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function List() {
+        const [, setFoo] = createSignal(0)
+        return (
+          <ul onClick={() => setFoo(1)}>
+            {[1, 2, 3].map(n => <li key={n}>{n}</li>)}
+          </ul>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'List.tsx')
+    expect(clientJs).not.toMatch(/\bmapArray\s*\(/)
+    expect(clientJs).not.toMatch(/\breconcileList\s*\(/)
+  })
+
+  test('static prop array stays static', () => {
+    // `items` is a non-destructured prop; the AST is just an
+    // Identifier, no calls. Should still take the static path.
+    const source = `
+      export function List(props: { items: string[] }) {
+        return (
+          <ul>
+            {props.items.map(item => <li key={item}>{item}</li>)}
+          </ul>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'List.tsx')
+    expect(clientJs).not.toMatch(/\bmapArray\s*\(/)
+    expect(clientJs).not.toMatch(/\breconcileList\s*\(/)
+  })
+
+  test('unrecognised-call chain with filter now reconciles', () => {
+    // `computeItems().filter(t => t.done).map(...)`: the filter
+    // predicate is extracted by Phase 1 so `array` normalises to
+    // `computeItems()` — still a CallExpression at the AST level, so
+    // the widened gate forces reconciliation. Without the widening the
+    // extracted-chain path dropped the reactivity signal entirely.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { computeItems } from './data'
+
+      export function DoneList() {
+        const [, setFoo] = createSignal(0)
+        return (
+          <ul onClick={() => setFoo(1)}>
+            {computeItems().filter(t => t.done).map(t => <li key={t.id}>{t.label}</li>)}
+          </ul>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'DoneList.tsx')
+    expect(clientJs).toMatch(/mapArray|reconcile/)
+    expect(clientJs).toContain('computeItems()')
+  })
+
+  test('nested loop with unrecognised outer call reconciles', () => {
+    // Outer `outer()` is an imported helper; inner loop iterates a
+    // field on each outer item. The outer loop must reconcile (new
+    // behaviour); the inner loop is handled by the existing
+    // nested-loop metadata path.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { outer } from './data'
+
+      export function Groups() {
+        const [, setFoo] = createSignal(0)
+        return (
+          <ul onClick={() => setFoo(1)}>
+            {outer().map(o => (
+              <li key={o.id}>
+                <ul>{o.children.map(c => <li key={c}>{c}</li>)}</ul>
+              </li>
+            ))}
+          </ul>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Groups.tsx')
+    expect(clientJs).toMatch(/mapArray|reconcile/)
+    expect(clientJs).toContain('outer()')
+  })
+
+  test('loop with child component on unrecognised-call array reconciles', () => {
+    // Composite rendering path: the loop body is a single component.
+    // Previously the `listOf()` silent-drop meant Items were rendered
+    // once at SSR and never re-created on signal change. The widened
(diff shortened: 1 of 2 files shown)


kfly8 (commented):
## Overview

Final piece of the #937 rollout — carries wrap-by-default to `.map()` loops. +286/-4, `jsx-to-ir.ts::transformMapCall` (+34/-4), one new test file (+247), all CI green. Closes #943.

The change tracks `arrayExpr: ts.Expression` alongside the existing `array` string in `transformMapCall`, updates it at all three assignment sites (`setArray`, simple-filter fallback, no-chain fallback), and narrows `isStaticArray`:

```ts
const isStaticArray =
  !isSignalOrMemoArray(array, ctx)
  && (isDestructuredParam || !exprHasFunctionCalls(arrayExpr))
```

— any array whose AST contains a call now forces reconciliation, unless the map callback destructures its item parameter (see concern #1 below).

## Strengths

- **Uses the AST flag directly, not a regex.** `exprHasFunctionCalls(arrayExpr)` is the same helper the text interpolation pilot (#939) and conditional follow-up (#941) use. The PR description explicitly frames this as addressing concern #1 from the #947 round-1 review (AST-vs-regex divergence). For this site at least, the rollout is now back on the uniform AST-flag pattern — #942 remains the only regex-based site, and becomes the natural target of a future consolidation pass.
- **Latent emitter bug discovered, diagnosed, and safely guarded.** Dynamic arrays with destructured map callbacks (`([, cfg]) => ...`) crash at runtime because `mapArray` passes a function-accessor that isn't iterable. Rather than regress `pie-chart-demo.tsx` and friends, the PR skips the widening for destructured params, documents the skip with a test, and calls out the underlying emitter bug in a comment. Fidelity-first: don't break a currently-working SSR path just to theoretically improve it.
- **Two-metric fixture sweep.** 4 / 181 components (~2.2%, against 30% threshold) AND +2326 bytes (~0.009%, against 5% threshold). Since `reconcileList` / `mapArray` wrap is more expensive per-site than a one-shot `createEffect` (text/attr/conditional/prop), measuring bytes in addition to file count is the right discipline. Both metrics pass comfortably.
- **Three assignment sites updated symmetrically.** `setArray` closure, simple-filter branch, and fallback — all three now assign `arrayExpr` in sync with `array`. No risk of the AST node falling out of step with the serialized string.
- **Control-flow default chosen correctly.** `arrayExpr = mapSource` as the declaration default — matches the fallback path at the bottom of the if/else chain, so if the chain evolves to add a new branch that forgets to assign `arrayExpr`, the default is at least safe rather than undefined.
- **Seven tests covering signal / unrecognised call / inline literal / static prop / filter chain / nested loop / composite createComponent path, plus the destructured-param guard.** Matches or exceeds the matrix of the earlier rollout PRs.

## Concerns

### 1. Destructured-param latent emitter bug needs its own follow-up issue

The guard `isDestructuredParam = param.startsWith('[') || param.startsWith('{')` short-circuits reconciliation for array-destructure / object-destructure callbacks. Correct scope call for THIS PR — don't regress. But the underlying bug is broader than #943's widening: a user writing `signalArray().map(([, cfg]) => ...)` where `signalArray` IS a recognised signal will hit the same `mapArray` crash because `isSignalOrMemoArray` returns true and the destructure-param guard doesn't apply on the signal path. The widening hides the bug from `getItems()`-style calls; signal-typed arrays were already exposed.

**Recommendation**: file a follow-up issue titled something like *"emitter: unwrap `item()` accessor for destructured `.map()` callback params"* and link from this PR. Short-term the guard is the right call, but the bug should not disappear into a comment.

### 2. `isDestructuredParam` heuristic may miss parenthesized/type-annotated params

The check is a string-prefix on the serialized `param`. Works for:

- `([, cfg]) => ...` — `param = "[, cfg]"` — starts with `[`. ✓
- `({a, b}) => ...` — `param = "{a, b}"` — starts with `{`. ✓

What about a TS-typed destructure like `([, cfg]: [string, Cfg]) => ...`? If `param` is extracted with annotation intact, it might start with `(` or include the type. Worth a quick test with an explicit type annotation on a destructured param to confirm the prefix check still fires. If it doesn't, the guard silently fails → `mapArray` crash at hydration for type-annotated destructures on dynamic arrays.

Not obvious from the diff how `param` is serialized upstream — adding one test case with a typed destructure would close the gap.

### 3. Broad `mapArray|reconcile` regex in assertions

Tests 1, 2, 5, 6, and the composite test use `toMatch(/mapArray|reconcile/)`. That fires on ANY occurrence of either word, including unrelated pre-existing wiring. Tighter assertions (e.g. `toContain("mapArray(")` scoped to the specific array expression) would catch regressions where one primitive silently replaces another. Low priority; the current assertions are correct for the happy path.

### 4. Nested loop test only asserts outer (6)

Test 6's comment says "the inner loop is handled by the existing nested-loop metadata path" but the assertion only checks `mapArray|reconcile` exists at all. It would pass even if the inner loop's body were dropped. An additional `toContain('o.children')` or similar on the inner's array expression would confirm nested-loop emission still fires end-to-end.

- **Correctness**: Closes the loop-shape silent-drop gate. The latent emitter bug around destructured params is handled defensively — no regression introduced. Every fixture-diff site is a genuinely-reactive-looking array that was previously frozen.
- **Performance**: Each new reconciliation site adds ~580 bytes (`mapArray` wrap + closure). The 4 affected components are non-reactive constants today, so they're over-wrapped — future-proof, but a genuine cost. The 0.009% aggregate byte delta is well inside the project's fidelity-first priority.
- **Security**: Internal compiler change, no user input surface. N/A.

## Verdict

**Ready to merge**, and with this PR the #937 rollout is complete across all four planned sites. The AST-flag approach here closes the divergence concern from the #947 review for three of the four sites; only child-component props (#942) is still on the regex path and becomes the natural scope for the DRY consolidation PR discussed in #937's follow-ups section.

Highest priority of the concerns is filing a separate issue for the destructured-param emitter bug before the context fades — it is not #943's problem to fix, but it is a known-bad-path worth tracking explicitly rather than leaving in a code comment.

kfly8 (commented):
## Re-review (round 2)

Two follow-up commits since the first review:

- `6287007e fix(compiler): skip #943 widening for destructured loop params`
- `4334eeea test(compiler): tighten loop-fallback assertions, pin typed destructure`

Plus a freshly-filed standalone issue **#949** for the underlying emitter bug.

CI re-run, all green.

### Addresses

|-------|--------|
| #1 Destructured-param emitter bug needs a follow-up issue | **Fixed.** #949 filed with a concrete repro; code comment updated from "This is a latent emitter bug that predates #943" to "This is a latent emitter bug **tracked in #949**". The known-bad path is now in the tracker, not only in a code comment. |
| #2 `isDestructuredParam` prefix check may miss typed destructures | **Fixed.** New test `typed destructured map param also skips widening` pins `([, cfg]: [string, Cfg])` → still hits the prefix check. The commentary in the test explains **why** it works (type annotation lives on `firstParam.type`, not `firstParam.name`; `.name.getText()` returns just the binding pattern), turning the guard from "probably works" into "demonstrably works for the shapes we care about". Code comment cross-references the pinning test. |
| #3 Broad `mapArray\|reconcile` regex in assertions | **Fixed.** All 5 affected assertions tightened to `toContain('mapArray(')`. Specific to the primitive we expect; catches regression-via-primitive-swap. |
| #4 Nested loop test only asserts outer | **Fixed.** Test 6 now also asserts `toContain('o.children')` — confirms the inner loop's array expression survives in the emitted template, so a future regression that dropped inner-loop emission would break this test red. |

### Delta verification

- Tightened assertions narrow the test space — no coincidental `reconcile` mention elsewhere in the bundle can hide a real regression now.
- Typed-destructure pinning (new test) exercises the exact realistic case: TS type annotation on the binding pattern. Would have been a theoretical gap in round 1; now locked in.
- Comment → #949 cross-reference creates a symmetric trail: issue describes the bug, code comment in the guard explains the mitigation, test pins the guard's behaviour. Anyone removing the guard has three places to stumble over before accidentally regressing.
- All 687 jsx tests + adapter-tests + e2e still green.

### Verdict

**Still ready to merge.** Every concern from round 1 is addressed — either by a direct code change, a tighter test, or a linked follow-up issue. No new risk introduced by the tightening commits; they only narrow existing behaviour, don't widen it.

With #948 merged, the #937 rollout is complete across all four sites (text / attr / conditional / child-prop / loops — actually five, but who's counting). The natural next issues, in rough priority order:

1. **#937 body itself** — close it; rollout is done.
2. **#949** — emitter fix for destructured `.map()` callback params.
3. **DRY consolidation PR** — bring #942 (child-prop) onto the same AST-flag pattern as #939/#941/#943 so all five sites share a helper.
4. **#944** — `barefoot why-wrap` CLI.

PR #950, opened 2026-04-21
fix(compiler): unwrap accessor for destructured .map() params (#949)

### LINKED ISSUE #949: compiler: emitter crashes on destructured `.map()` callback params for dynamic arrays
## Background

Surfaced while landing #943 (loop wrap-by-default fallback). Called out as Concern 1 in the #948 review — filing as a standalone tracker so the known-bad path doesn't disappear into a code comment.

## The bug

When a `.map()` callback destructures its item parameter and the array flows through the dynamic `mapArray` path, hydration crashes with:

```
Uncaught TypeError: function is not iterable (cannot read property Symbol(Symbol.iterator))
```

Root cause: the emitter in `packages/jsx/src/ir-to-client-js/emit-control-flow.ts` interpolates `elem.param` verbatim into the `mapArray` renderItem arrow head, e.g.

```js
mapArray(() => X, container, null, ([, cfg], __idx, __existing) => { ... })
```

But `mapArray` at `packages/client/src/runtime/map-array.ts` passes the item as a **signal accessor (function)** to `renderItem`, per its contract:

```ts
// renderItem receives item as signal accessor: item() returns current value.
```

Destructuring a function throws. Any binding pattern — array (`([, cfg])`), object (`({ a, b })`), or either with a type annotation — trips the same crash.

## Current state — #943 mitigation

#943 added a narrow guard in `jsx-to-ir.ts::transformMapCall`:

```ts
const isDestructuredParam = param.startsWith('[') || param.startsWith('{')
const isStaticArray =
  !isSignalOrMemoArray(array, ctx)
  && (isDestructuredParam || !exprHasFunctionCalls(arrayExpr))
```

This **skips the #937 wrap-by-default widening** for destructured params, preserving the existing static-render path for cases like `Object.entries(chartConfig).map(([, cfg]) => ...)`. A regression test pins the skip: `packages/jsx/src/__tests__/loop-fallback-wrap.test.ts` → `destructured map param skips widening (latent mapArray bug)`.

**The guard does not fix the underlying bug.** A user writing `signalArray().map(([a, b]) => ...)` where `signalArray` is a recognised signal still hits the crash, because `isSignalOrMemoArray(...)` returns true and the guard doesn't apply on the signal path. #943 only hides the bug for *unrecognised-call* arrays.

## Proposed fix

Teach the emitter to detect destructuring patterns in `loop.param` and unwrap the accessor inside the renderItem body:

```js
// Simple-name param (current, unchanged):
mapArray(() => X, container, null, (item, __idx, __existing) => {
  // template references use item() — template already wraps via wrapLoopParamAsAccessor
})

// Destructured param (new):
mapArray(() => X, container, null, (__item, __idx, __existing) => {
  const [, cfg] = __item()  // unwrap accessor once at renderItem entry
  // template references `cfg` directly as a captured value
})
```

Sites to update in `emit-control-flow.ts`:
- Line 109, 111, 117, 219 (conditional-branch loop emission)
- Line 357 (nested branch inner loops)
- Line 735, 839, 842 (plain element loop reconciliation, with/without reactive effects)
- Line 1189 (nested inner loops)
- Line 1308 (another plain element loop emit)

Each site currently does `(${elem.param}, ${indexParam}, __existing) => ...`. All need the same destructure-vs-simple-name branching.

## Reactivity note (design question)

Captured-once destructuring means fine-grained per-item updates stop working when the signal item is replaced:

```jsx
items().map(([a, b]) => <li>{a}{b}</li>)
```

With the simple fix above, `a` and `b` are captured at first render. `mapArray` calls `setItem` on same-key updates, but the destructured locals never re-read. Template-level interpolations like `${a}` would remain stale.

Options to preserve reactivity:
1. **Re-destructure in a `createEffect`** so `a` and `b` are refreshed every item change. Requires making the destructured locals either signals themselves or re-assigned within reactive scope.
2. **Emit the destructure as getters** on a proxy around `__item()`. Costly and hard to reason about.
3. **Expand the destructure at template-generation time** — rewrite template references like `${a}` to `${__item()[0]}` and `${b}` to `${__item()[1]}`. Cleanest for primitives; trickier for nested `({ user: { name } })`.

Option 3 is closest in spirit to how simple-name params are handled today (template-time rewriting via `wrapLoopParamAsAccessor`). Worth a separate design discussion before implementation.

## Acceptance criteria

- Dynamic-array `.map()` with destructured callback params no longer crashes at hydration.
- `signalArray().map(([a, b]) => ...)` and `signalArray().map(({ x, y }) => ...)` — including typed and parenthesized forms — render correctly and update reactively when the signal array changes.
- #943's destructured-param guard in `transformMapCall` can be removed (or converted to an internal-consistency check).
- Regression test in `loop-fallback-wrap.test.ts` updated to assert the dynamic path is taken, not skipped.

## References

- Discovered: #943 CI failure on `pie-chart.spec.ts > Interactive > switching metric updates the chart`
- Guarded in #943 commit `6287007e`
- Raised in: #948 review (Concern 1)

kfly8:
## Summary

- Fixes the latent `mapArray` emitter crash for destructured `.map()` callback params (`([, cfg]) => ...`, `({ a, b }) => ...`, typed and parenthesized forms) — surfaced during the #937 wrap-by-default rollout and guarded off in #943.
- Emitter (`ir-to-client-js/emit-control-flow.ts`) now renames the renderItem arrow param to a synthetic `__bfItem` accessor and unwraps the destructuring pattern once at body entry. Applied consistently across all 10 `mapArray(...)` emission sites via a shared `destructureLoopParam` helper.
- Removes the `isDestructuredParam` skip in `jsx-to-ir.ts::transformMapCall`; signal/memo and unrecognised-call arrays with destructured callbacks now widen to `mapArray` like any other dynamic loop.

## Design decision

Issue proposed three options for the reactivity of captured destructured bindings. After re-evaluating against the existing `wrapLoopParamAsAccessor` layer:

- **Chose Option 1 (renderItem-entry unwrap)** over Option 3 (template-time rewriting). Option 3 would require making `ctx.loopParams` a path-aware map and extending the regex-based wrapper to rewrite binding references to `__bfItem().path` — a bigger change that is out of scope for this crash fix.
- **Known limitation**: destructured locals are captured at first render. Array-level updates (add / remove / reorder / key change) refresh correctly because each new key invokes a fresh `renderItem`. Same-key `setItem` updates do **not** refresh the destructured locals — tracked in **#951** as the Option 3 follow-up (includes a todo-toggle repro for user self-diagnosis).

See the `destructureLoopParam` doc comment and the updated `transformMapCall` comment for the in-code rationale. The limitation is pinned behaviourally by `map-array.test.ts::destructured locals captured once — frozen on same-key update (known limitation)` so Option 3 (#951) has to consciously flip it.

## Test plan

- [x] `bun test packages/jsx packages/adapter-tests packages/{dom,cli,client,hono,go-template}` — all green
- [x] Unit tests in `packages/jsx/src/__tests__/loop-fallback-wrap.test.ts` — inverted the two pinning tests (8th and 9th) to assert the widening now applies, asserted the emitted `__bfItem` + unwrap shape, and added a new object-pattern case on a signal array
- [x] Added runtime unit tests in `packages/client/__tests__/runtime/map-array.test.ts` (DOM update verification): array-pattern and object-pattern destructure via `__bfItem()` unwrap with add/reorder/remove, plus a pinning test for the same-key limitation (#951)
- [x] Fixture sweep (`site/ui/dist/components`): count 364 → 364 (0% change, well within 30%), bytes 30M → 30M (0% change, within 5%). Only `pie-chart-demo-*.js` hash changed — its Donut demo uses `Object.entries(chartConfig).map(([, cfg]) => ...)`, which now correctly flows through `mapArray`
- [x] E2E (`cd site/ui && bunx playwright test`): pie-chart suite 10/10 green including `Interactive > switching metric updates the chart` (the exact case that drove #943). Full suite: 1385 passed, 2 pre-existing failures in `form-builder.spec.ts` unrelated to #949 (form-builder-demo compiled JS is byte-identical on `main` and this branch — verified via worktree)

## Review round-2 follow-ups

- Renamed synthetic accessor `__loopItem` → `__bfItem` (barefoot-reserved namespace).
- Documented the `destructureLoopParam` heuristic's dependency on `firstParam.name.getText()` shape.
- Added the same-key limitation pinning test so Option 3 flip is conscious.
- Filed **#951** as the Option 3 tracker with a todo-toggle repro; linked from the code comment, the pinning test comment, and this PR description.

## Related

- Closes #949
- Follow-up: #951 (Option 3 — same-key fine-grained reactivity via `__bfItem().path` template rewriting)
- Related: #943 (pinning guard being removed), #948 (review surfaced this follow-up)

### DIFF
--- a/packages/client/__tests__/runtime/map-array.test.ts
+++ b/packages/client/__tests__/runtime/map-array.test.ts
@@ -390,9 +390,9 @@ describe('mapArray', () => {
   // at first render. mapArray skips renderItem for same-key setItem
   // updates, which means the locals freeze — fine-grained reactivity
   // through destructured bindings needs Option 3 (template-time
-  // rewriting to `__bfItem().path`). This test locks that behaviour
-  // in so a future Option 3 PR turns a red test green rather than
-  // silently changing semantics.
+  // rewriting to `__bfItem().path`), tracked in #951. This test locks
+  // the current behaviour in so the Option 3 PR must consciously flip
+  // it from green to red rather than silently changing semantics.
   test('destructured locals captured once — frozen on same-key update (known limitation)', () => {
     const [items, setItems] = createSignal<[string, string][]>([['1', 'A']])
 
--- a/packages/client/__tests__/runtime/map-array.test.ts
+++ b/packages/client/__tests__/runtime/map-array.test.ts
@@ -301,4 +301,86 @@ describe('mapArray', () => {
     expect(renderCount).toBe(1)
     expect(container.children[0].textContent).toBe('B')
   })
+
+  // Regression: #949. The emitter (see destructureLoopParam in
+  // emit-control-flow.ts) now renames the renderItem param to a synthetic
+  // accessor and unwraps once at body entry. This shape — "function-typed
+  // param unwrapped via __loopItem()" — is what the compiled output
+  // produces for `signalArr().map(([a, b]) => ...)`. The test exercises
+  // the contract directly against the runtime so the fix doesn't silently
+  // regress on the client side.
+  test('destructuring via accessor unwrap works for array-pattern params', () => {
+    const [items, setItems] = createSignal<[string, string][]>([
+      ['1', 'A'],
+      ['2', 'B'],
+    ])
+
+    mapArray(
+      items,
+      container,
+      (item, _i) => item[0],
+      // Shape matches compiled output: synthetic __loopItem, unwrap at entry.
+      (__loopItem) => {
+        const [a, b] = __loopItem()
+        const li = document.createElement('li')
+        li.setAttribute('data-a', a)
+        li.textContent = b
+        return li
+      },
+    )
+
+    expect(container.children.length).toBe(2)
+    expect(container.children[0].getAttribute('data-a')).toBe('1')
+    expect(container.children[0].textContent).toBe('A')
+    expect(container.children[1].getAttribute('data-a')).toBe('2')
+    expect(container.children[1].textContent).toBe('B')
+
+    // Array-level update: add + reorder must update DOM.
+    setItems([
+      ['3', 'C'],
+      ['1', 'A'],
+      ['2', 'B'],
+    ])
+
+    expect(container.children.length).toBe(3)
+    expect(container.children[0].getAttribute('data-a')).toBe('3')
+    expect(container.children[1].getAttribute('data-a')).toBe('1')
+    expect(container.children[2].getAttribute('data-a')).toBe('2')
+
+    // Remove one.
+    setItems([['1', 'A']])
+    expect(container.children.length).toBe(1)
+    expect(container.children[0].getAttribute('data-a')).toBe('1')
+  })
+
+  test('destructuring via accessor unwrap works for object-pattern params', () => {
+    const [items, setItems] = createSignal([
+      { id: '1', label: 'A' },
+      { id: '2', label: 'B' },
+    ])
+
+    mapArray(
+      items,
+      container,
+      (item) => item.id,
+      (__loopItem) => {
+        const { id, label } = __loopItem()
+        const li = document.createElement('li')
+        li.setAttribute('data-id', id)
+        li.textContent = label
+        return li
+      },
+    )
+
+    expect(container.children.length).toBe(2)
+    expect(container.children[0].getAttribute('data-id')).toBe('1')
+    expect(container.children[0].textContent).toBe('A')
+
+    setItems([{ id: '2', label: 'B' }, { id: '3', label: 'C' }])
+
+    expect(container.children.length).toBe(2)
+    expect(container.children[0].getAttribute('data-id')).toBe('2')
+    expect(container.children[1].getAttribute('data-id')).toBe('3')
+    expect(container.children[1].textContent).toBe('C')
+  })
 })
--- a/packages/client/__tests__/runtime/map-array.test.ts
+++ b/packages/client/__tests__/runtime/map-array.test.ts
@@ -305,7 +305,7 @@ describe('mapArray', () => {
   // Regression: #949. The emitter (see destructureLoopParam in
   // emit-control-flow.ts) now renames the renderItem param to a synthetic
   // accessor and unwraps once at body entry. This shape — "function-typed
-  // param unwrapped via __loopItem()" — is what the compiled output
+  // param unwrapped via __bfItem()" — is what the compiled output
   // produces for `signalArr().map(([a, b]) => ...)`. The test exercises
   // the contract directly against the runtime so the fix doesn't silently
   // regress on the client side.
@@ -319,9 +319,9 @@ describe('mapArray', () => {
       items,
       container,
       (item, _i) => item[0],
-      // Shape matches compiled output: synthetic __loopItem, unwrap at entry.
-      (__loopItem) => {
-        const [a, b] = __loopItem()
+      // Shape matches compiled output: synthetic __bfItem, unwrap at entry.
+      (__bfItem) => {
+        const [a, b] = __bfItem()
         const li = document.createElement('li')
         li.setAttribute('data-a', a)
         li.textContent = b
@@ -363,8 +363,8 @@ describe('mapArray', () => {
       items,
       container,
       (item) => item.id,
-      (__loopItem) => {
-        const { id, label } = __loopItem()
+      (__bfItem) => {
+        const { id, label } = __bfItem()
         const li = document.createElement('li')
         li.setAttribute('data-id', id)
         li.textContent = label
@@ -382,5 +382,52 @@ describe('mapArray', () => {
     expect(container.children[0].getAttribute('data-id')).toBe('2')
     expect(container.children[1].getAttribute('data-id')).toBe('3')
     expect(container.children[1].textContent).toBe('C')
+  })
+
+  // Pinning test for the #949 known limitation. The fix (see
+  // destructureLoopParam in emit-control-flow.ts) unwraps the accessor
+  // once at renderItem body entry, so destructured locals are captured
+  // at first render. mapArray skips renderItem for same-key setItem
+  // updates, which means the locals freeze — fine-grained reactivity
+  // through destructured bindings needs Option 3 (template-time
+  // rewriting to `__bfItem().path`). This test locks that behaviour
+  // in so a future Option 3 PR turns a red test green rather than
+  // silently changing semantics.
+  test('destructured locals captured once — frozen on same-key update (known limitation)', () => {
+    const [items, setItems] = createSignal<[string, string][]>([['1', 'A']])
+
+    mapArray(
+      items,
+      container,
+      (item) => item[0],
+      (__bfItem) => {
+        const [, b] = __bfItem()
+        const li = document.createElement('li')
+        // Read `b` once — mirrors captured-semantics in compiled output.
+        li.setAttribute('data-b', b)
+        li.textContent = b
+        return li
+      },
+    )
+
+    const firstEl = container.children[0]
+    expect(firstEl.textContent).toBe('A')
+    expect(firstEl.getAttribute('data-b')).toBe('A')
+
+    // Same key '1', new inner value — mapArray fires setItem and
+    // reuses the DOM node, but `b` was captured on first render so
+    // the DOM attribute / text stay at the original value.
+    setItems([['1', 'B']])
+
+    expect(container.children[0]).toBe(firstEl)
+    expect(firstEl.textContent).toBe('A')
+    expect(firstEl.getAttribute('data-b')).toBe('A')
+
+    // Array-level update (different key) produces a fresh renderItem
+    // call, so `b` is re-captured from the new value — this path works.
+    setItems([['2', 'C']])
+
+    expect(container.children[0].getAttribute('data-b')).toBe('C')
+    expect(container.children[0].textContent).toBe('C')
   })
 })
--- a/packages/jsx/src/__tests__/loop-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/loop-fallback-wrap.test.ts
@@ -187,18 +187,15 @@ describe('Solid-style wrap-by-default fallback for loops (#943)', () => {
     expect(clientJs).toContain('o.children')
   })
 
-  test('destructured map param skips widening (latent mapArray bug)', () => {
+  test('destructured map param now reconciles (#949 emitter fix)', () => {
     // `mapArray` passes the item as a signal accessor (a function) to the
-    // renderItem callback; destructuring a function throws
-    // "function is not iterable" at runtime. The emitter currently
-    // interpolates `elem.param` verbatim into the renderItem arrow head,
-    // so `([, cfg]) => ...` on a dynamic array crashes at hydration.
-    //
-    // Until the emitter is taught to unwrap `item()` for destructured
-    // params, the #943 widening skips these cases to avoid regressing
-    // previously-working SSR-only components. The call shape is present
-    // but the array stays on the static path — a known trade-off, not a
-    // silent-drop (the frozen SSR output is the existing behaviour).
+    // renderItem callback; destructuring a function would throw
+    // "function is not iterable" at runtime. Before #949 the emitter
+    // interpolated `elem.param` verbatim into the renderItem arrow head,
+    // so this path was excluded from the #943 widening to avoid the
+    // crash. With #949 the emitter renames the param to a synthetic
+    // accessor and unwraps once at body entry, so destructured-param
+    // callbacks are safe to reconcile like any other dynamic array.
     const source = `
       'use client'
       import { createSignal } from '@barefootjs/client'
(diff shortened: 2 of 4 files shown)


kfly8 (commented):
## Overview

Fixes the latent `mapArray` emitter crash for destructured `.map()` callback params flagged in #949 and worked around in #943. +206/-49, 4 files (2 source, 2 test), all CI green including the benchmark workflow. Closes #949.

Core change: **shared `destructureLoopParam` helper in `emit-control-flow.ts`** that rewrites `([, cfg]) => ...` / `({ a, b }) => ...` into `__loopItem => { const [, cfg] = __loopItem(); ... }`. Applied consistently across **all 10 `mapArray(...)` emission sites** (branch / composite / plain / nested / reconciliation paths). The `isDestructuredParam` bypass added in #943 is removed — dynamic arrays with destructured callbacks now take the same widened path as any other loop.

## Strengths

- **Three options from #949 re-evaluated, Option 1 chosen with documented rationale.** PR description + code comment in `destructureLoopParam` explain why renderItem-entry unwrap was picked over template-time rewriting (Option 3). Scope is the crash fix, not a redesign of `ctx.loopParams`.
- **Single helper covers 10 emission sites.** One conceptual change, applied consistently. The kind of refactor that prevents re-divergence of the emitter — much better than inlining the `(head, unwrap) = ...` at each site.
- **Runtime-level test added, not only emitter-shape test.** `packages/client/__tests__/runtime/map-array.test.ts` now has two new cases that invoke `mapArray` directly with the exact shape the emitter produces (synthetic `__loopItem` + entry unwrap). If either the emitter or the runtime drifts, the runtime test turns red — catches the "emitter and runtime disagree silently" regression class.
- **Pinning tests inverted cleanly.** The two `loop-fallback-wrap.test.ts` tests that previously asserted "widening skipped" now assert the opposite: `mapArray(` IS emitted, `(__loopItem, ` IS the renderItem head, and `const [, cfg] = __loopItem();` appears in the body. A new object-pattern test on a signal array (`{ label, value }`) covers the dual of the array-pattern case.
- **Fixture sweep: 0% delta.** 364 → 364 files, 30M → 30M bytes; only `pie-chart-demo-*.js` hash changes (the exact component #943 was worked around for). No byte inflation in other compiled outputs.
- **E2E verified locally against the motivating case.** pie-chart suite 10/10 including `Interactive > switching metric updates the chart` — the specific failure scenario that drove #943. Author ran this before submitting, called it out in the PR body.
- **`jsx-to-ir.ts::transformMapCall` simplified back to clean form.** The `isDestructuredParam || !exprHasFunctionCalls(...)` compound is removed; `isStaticArray = !isSignalOrMemoArray && !exprHasFunctionCalls` again. Comment updated to reference #949 + the emitter helper.
- **Known limitation documented honestly, not buried.** Same-key `setItem` updates do not refresh destructured locals (because mapArray skips renderItem on same key). PR description + helper doc comment both call this out and point to Option 3 as the follow-up. No silent gap.

## Concerns (non-blocking)

### 1. Same-key `setItem` limitation has no pinning test

The PR correctly documents that destructured locals freeze under same-key updates — array-level updates (add / reorder / remove / key change) refresh, in-place mutations do not. But there's no test that pins this behaviour. If a future refactor accidentally fixes it (good!) or accidentally breaks array-level updates while leaving the limitation wording in place (bad!), nothing will fail red.

Suggestion: add one `test()` in `map-array.test.ts` that mutates an item in-place (same key) and asserts that destructured locals stay at their initial values. This gives Option 3, when it lands, a **pre-existing red test that turns green** — a standard way to pin scope. Alternatively, a `test.todo(...)` with the same intent communicates the gap to future readers.

### 2. `destructureLoopParam` heuristic is string-prefix on `param`

```ts
if (param.startsWith('[') || param.startsWith('{')) { ... }
```

Works for the shapes the serializer currently produces. Relies on `firstParam.name.getText()` stripping parens, type annotations, and leading whitespace. If the upstream serializer ever changes (e.g., retains parens for `(a)` or returns `  [a, b]` with leading whitespace from trivia), the helper silently stops detecting destructures and the crash re-emerges.

Suggestion: one-liner guard comment inside the helper saying `// relies on param being the trimmed output of firstParam.name.getText() — no leading parens / whitespace`. Belt-and-braces so the next reader doesn't refactor the upstream without seeing the dependency.

### 3. `__loopItem` could collide with a user param

If a user happens to write `.map(__loopItem => __loopItem.id)`, the synthetic rename produces `const __loopItem = __loopItem()` — self-referential, ReferenceError at hydration. Extremely unlikely, but a namespaced sentinel (`__bf_loopItem`, `__bfItem`, or similar) would make the collision mathematically impossible rather than just improbable. Low priority.

### 4. Benchmark workflow passed but no delta reported

PR mentions the benchmark workflow ran green, which is good — but the PR body doesn't surface a before/after number. For a change that widens a hot path (every destructured map callback now goes through `mapArray`), a one-line "benchmark delta: +X%" or "no measurable change" would close the loop. Optional.

- **Correctness**: Closes the #943 workaround gap cleanly. Known limitation (same-key refresh) is narrower than the bug being fixed and matches what Solid's own runtime-tracking contract would produce for literal-destructured bindings. No new risk.
- **Performance**: Each destructured-callback loop gains one accessor invocation per renderItem call. Negligible compared to the DOM work the same renderItem already does. Benchmark workflow confirmed no regression at the CI budget level.
- **Security**: Internal emitter change, no user input surface. N/A.

## Verdict

**Ready to merge.** The 10-site consistency via one helper, the runtime-layer test coverage, and the honest limitation documentation are exactly the shape a bug fix for an emitter-runtime contract drift should take. Concerns 1–4 are nits and follow-up tracking — none blocks.

With #950 merged, the #937 rollout's last known gap is closed. Next priorities per earlier discussion:
- DRY consolidation PR (bring #942's regex path onto the AST-flag helper used by #939 / #941 / #943).
- #944 (`barefoot why-wrap` CLI).
- Option 3 follow-up for same-key fine-grained reactivity through destructured bindings (if / when needed).

kfly8:
Thanks for the review. Pushed commit `f9b896b` addressing all four concerns:

**Concern 3 — namespaced sentinel (done)**
Renamed `__loopItem` → `__bfItem`. The `__bf`/`__bf_` namespace is already reserved by the project (`bf-s`, `bf-key`, `bf-loop`, `BF_KEY`, etc.), so a user-written `.map(__bfItem => ...)` collision would be a naming-convention violation on their side rather than a realistic user input.

**Concern 2 — heuristic dependency doc (done)**
Added a paragraph to the `destructureLoopParam` doc comment calling out the dependency on `firstParam.name.getText()` returning a trimmed binding pattern (no leading parens / whitespace / type annotations). If the upstream serializer ever drifts, the prefix check must be widened in lockstep.

**Concern 1 — pinning test for the same-key limitation (done)**
Added `destructured locals captured once — frozen on same-key update (known limitation)` in `map-array.test.ts`. It uses the exact shape the emitter produces (synthetic `__bfItem` + entry unwrap) and asserts that:
- same-key `setItem([['1', 'B']])` leaves DOM at `'A'` (frozen — current behaviour)
- key-change `setItem([['2', 'C']])` refreshes to `'C'` (fresh renderItem call)

When Option 3 lands it turns the first assertion red — a standard way to pin scope.

**Concern 4 — benchmark delta (FYI)**
The benchmark workflow only runs when `packages/client/**` changes, so there's no matching main-branch run to diff against. Two runs on this PR (pre- and post-review-fix):

| Operation | pre-fix | post-fix | SolidJS |
|---|---|---|---|
| create | 4.600 ms | 4.200 ms | 4.300 ms |
| partial update | 0.075 ms | 0.070 ms | 0.105 ms |
| clear | 3.300 ms | 3.600 ms | 3.900 ms |

BarefootJS tracks SolidJS within ~5% across both runs; variance between runs is within the usual CI noise band. No evidence of a widening-driven regression.

CI green on `f9b896b`. Ready for a second look whenever you have time.

kfly8 (commented):
## Re-review (round 2)

One follow-up commit: `f9b896b6 fix(compiler): address PR #950 review — namespaced sentinel, docs, pinning test`. CI green including benchmark.

### Addresses

|-------|--------|
| **#1 Same-key setItem limitation needs pinning** | **Fixed.** New runtime test `destructured locals captured once — frozen on same-key update (known limitation)`. Asserts (a) same-key setItem leaves DOM unchanged, (b) different-key setItem produces fresh renderItem + refreshed DOM. This is a "passing test of current limitation" — arguably better than the `test.todo` I originally suggested, because Option 3 will have to **consciously flip it from green to red** rather than silently fixing it. The test comment points forward to Option 3 explicitly. |
| **#2 Upstream-contract dependency undocumented** | **Fixed.** Doc comment on `destructureLoopParam` now calls out: "Detection relies on `param` being the trimmed output of `firstParam.name.getText()` … no leading parens, no type annotations, no trivia whitespace. If that upstream contract changes, the prefix check must be widened in lockstep or the #949 crash resurfaces silently." The dependency is now discoverable without archaeology. |
| **#3 `__loopItem` collision risk** | **Fixed.** Renamed to `__bfItem` with a new doc line: "The `__bf_` prefix is a barefoot-reserved namespace — avoids any chance of collision if the user writes `.map(item => ...)` with a matching name." All tests updated in lockstep. |
| #4 Benchmark delta not reported | Unchanged — labeled optional in round 1, still fine. |

### New concern surfaced (blocker for **closing #937**, not for merging #950)

The limitation now has a pinning test and thorough code comments, but there is **still no standalone issue for Option 3** (same-key fine-grained reactivity via `__bfItem().path` template rewriting).

Without a tracker, the limitation lives in:
- the `destructureLoopParam` doc comment,
- the PR description,
- the runtime test's commentary.

All three are great for someone already reading #950, but a user who hits a stale-DOM bug in a todo toggle won't search them. For the #937 rollout story — which explicitly decided *fidelity > client-JS size* — the minimum discoverable path is a GitHub issue titled something like:

> compiler: same-key `.map()` `setItem` updates don't refresh destructured locals (Option 3 from #949)

with a tiny repro (todo-toggle pattern) and a link from the `destructureLoopParam` doc. That lets external users self-diagnose without reading the emitter source.

### Strengths of the fix round

- **Pinning test documents limitation behaviourally, not just textually.** If a refactor accidentally fixes the limitation, the test turns red and the fix is conscious. If it accidentally breaks array-level updates, the second half of the test turns red too. Both failure modes caught.
- **Barefoot-reserved namespace** (`__bf_`) is the right habit for compiler-generated identifiers and should probably spread to other synthetic names in the emitter over time.
- **Upstream-contract comment** is exactly the kind of tripwire note that prevents the silent-regression pattern this whole #937 arc has been closing.

### Verdict

**Ready to merge** as soon as the Option 3 tracker issue is filed and linked from (a) the code comment, (b) the pinning test's comment, and (c) the PR description. Files don't need further source changes — the three doc-text updates to point at the tracker are trivial, and can happen before or in a quick follow-up commit.

After that, this closes #949 cleanly. The next rollout step (per earlier plan) becomes the DRY consolidation PR to bring #942 onto the AST-flag pattern.

kfly8:
Round-2 review addressed. Filed **#951** as the Option 3 tracker (*compiler: same-key .map() setItem updates don't refresh destructured locals (Option 3 from #949)*) with:

- a todo-toggle repro so users hitting stale-DOM on same-key mutations can self-diagnose by searching the issue tracker,
- an implementation sketch covering the `ctx.loopParams` path-map extension and the `wrapLoopParamAsAccessor` sibling rewrite,
- rest-element (`[a, ...rest]`) edge-case call-out (reject at compile time vs fall back to #950's unwrap-once),
- acceptance criteria that pin the conscious-flip of the `map-array.test.ts` limitation test.

Linked #951 from all three discoverability surfaces:

1. **Code comment** — `destructureLoopParam` doc in `emit-control-flow.ts`: *"tracked in #951 (Option 3 follow-up)"*
2. **Pinning test comment** — `map-array.test.ts`: *"needs Option 3 (template-time rewriting to `__bfItem().path`), tracked in #951"*
3. **PR description** — Design decision block + dedicated Follow-up line

Commit `7507116a` is doc-only (2 files, +5/−5). CI will rerun but no source logic changed from `f9b896b6`. Ready to merge once green.

kfly8 (commented):
## Re-review (round 3)

One new commit: `7507116a docs(compiler): link #951 from #949 limitation comments` — plus the corresponding **#951** issue filed. CI green.

### Addresses

Round-2 concern was the missing standalone Option 3 tracker. Fully resolved:

- **#951 filed** with exactly the content the round-2 review asked for: a concrete todo-toggle repro, behavioural delta between array-level and same-key updates, and a four-step Option 3 implementation sketch (binding-path map → extended `ctx.loopParams` → accessor-aware rewriter → drop the entry unwrap). External users hitting the stale-DOM symptom now have a searchable tracker.
- `destructureLoopParam` doc comment now ends with `tracked in #951 (Option 3 follow-up)` instead of the dangling "Option 3 follow-up" phrase.
- Runtime pinning test's commentary also links #951 and sharpens the language: "the Option 3 PR must consciously flip it from green to red rather than silently changing semantics."

Three reference points — code comment, test comment, standalone issue — are now mutually linked, so anyone landing on one can traverse to the others.

### Strengths of the documentation commit

- Scope is exactly what it claims: pure doc, -2/+2 in two files. No risk.
- #951's implementation sketch calls out the concrete data-structure change it needs (`ctx.loopParams` from `Set<string>` to a map with binding paths) — future implementer doesn't start from scratch.
- The "must consciously flip green→red" phrasing in the pinning test makes the intended workflow explicit: Option 3 PR is expected to edit this test, not just change the emitter.

### Verdict

**Ready to merge.** All three round-1 concerns (pinning test, upstream-contract doc, namespaced sentinel) and the round-2 concern (Option 3 tracker + cross-links) are resolved. CI green including benchmark. No new risk introduced by the doc commit.

After merge, the #937 rollout arc is functionally complete: all five sites converted (text / attr / conditional / child-prop / loop), the worst latent emitter bug exposed by the rollout is fixed here, and the one remaining fidelity gap has a dedicated tracker. The remaining follow-ups in priority order are:

1. Close #937 once this merges.
2. DRY consolidation PR (bring #942's regex path onto the AST-flag helper).
3. #944 (`barefoot why-wrap` CLI).
4. #951 (Option 3 — only if a real user report makes it load-bearing).

Issue #951, opened 2026-04-21
compiler: same-key .map() setItem updates don't refresh destructured locals (Option 3 from #949)


kfly8:
## Context

Follow-up to #949 (closed by #950). The emitter fix in #950 unwraps the item accessor once at `renderItem` body entry so destructured `.map()` callbacks (`([a, b]) => ...`, `({ x, y }) => ...`) stop crashing with "function is not iterable". Array-level updates (add / remove / reorder / key change) work correctly because `mapArray` invokes `renderItem` again on each new key.

**Known limitation (pinned by test)**: when a same-key `setItem` update fires, the destructured locals are frozen at their first-render values because `mapArray` intentionally reuses the DOM node and skips a new `renderItem` call. Fine-grained effects that read `__bfItem()` directly still refresh; effects that only read the destructured locals do not.

See the pinning test: `packages/client/__tests__/runtime/map-array.test.ts` → `destructured locals captured once — frozen on same-key update (known limitation)`.

## Repro

```tsx
'use client'
import { createSignal } from '@barefootjs/client'

type Todo = { id: string; label: string; done: boolean }

export function TodoList() {
  const [todos, setTodos] = createSignal<Todo[]>([
    { id: '1', label: 'First', done: false },
  ])

  const toggle = (id: string) =>
    setTodos(prev =>
      prev.map(t => (t.id === id ? { ...t, done: !t.done } : t)),
    )

  return (
    <ul>
      {todos().map(({ id, label, done }) => (
        <li key={id} class={done ? 'done' : ''}>
          <button onClick={() => toggle(id)}>{label}</button>
        </li>
      ))}
    </ul>
  )
}
```

Expected: clicking the button toggles `done`, and the `class="done"` (or absence) updates on the `<li>`.

Actual: the `<li>`'s class attribute stays at its first-render value. `mapArray` updates the per-item signal via `setItem`, but the destructured `done` local was captured at the first `renderItem` call and never re-read — so any template interpolation that only reads `done` (not `__bfItem().done`) is stale.

## Proposed fix — Option 3 from #949

Rewrite destructured binding references to `__bfItem().path` at template-generation time, symmetrically with how simple-name params are handled today via `wrapLoopParamAsAccessor`.

Sketch:

1. In `jsx-to-ir.ts::transformMapCall`, when the callback param is a binding pattern, walk it to build a `Map<bindingName, pathExpr>`:
   - `[, cfg]` → `cfg` ↦ `[1]`
[...]

PR #952, opened 2026-04-21
refactor(compiler): unify child-prop wrap-by-default on AST flags (#942 DRY)

kfly8:
## Summary

- Close out the last regex-based outlier in the #937 Solid-style wrap-by-default rollout: child-component prop bindings (#942) now gate on `IRProp.callsReactiveGetters` / `IRProp.hasFunctionCalls`, the same AST-derived flags #939 / #941 / #943 already use on `IRExpression` and `IRConditional`.
- Remove the `expandedValue` regex (`/\b\w+\s*\(/` with string-literal stripping) from `collect-elements.ts`. AST flags are computed Phase 1 on the attribute's source expression, so they can't false-match call-like substrings inside string literals (e.g. `{ color: 'hsl(221 83% 53%)' }`) and don't depend on the expansion order of local constants.
- Add a pinning test that an object-literal prop containing an `hsl(...)` string stays un-wrapped — a structural guarantee from the AST that the regex could only approximate.

## Context

Noted as the DRY follow-up on #937 and as concern 1 of the #948 PR review: 4 of the 5 wrap-by-default sites were already on AST flags; only #942 still diverged because Phase 2 (`collect-elements.ts`) has no AST access — the prop value has collapsed to `IRProp.value: string` by then. This PR carries the flags on `IRProp` itself so Phase 2 can consult them structurally.

## Fixture sweep (site/ui/dist/components)

181 files total, 4 changed (2.2%), -1771 bytes overall. Every diff is a reduction in redundant attribute re-sets inside an existing parent `createEffect`, where a non-reactive local-const value (e.g. `const today = new Date(); <Calendar fromDate={today} />`) was previously regex-caught after expansion. `initChild`-time initialization of those props is unchanged.

- `calendar-demo`: -601 bytes — `fromDate={today}`, `toDate={maxDate}`, `disabled={isWeekend}` no longer force re-binding on every `date()` tick
- `calendar-usage-demo`: -483 bytes — same pattern
- `command-demo`: -405 bytes — local-const arrow function prop
- `date-picker-demo`: -282 bytes — `formatDate={fmtDate}` where `fmtDate` is a local arrow const

## Test plan

- [x] `bun test packages/jsx` — 691 pass (existing 6 child-prop-fallback-wrap cases + new hsl-pinning case all green)
- [x] `bun test --cwd . packages/` — 1647 pass; the remaining 1 fail / 1 error is the pre-existing `packages/xyflow/e2e/flow.spec.ts` Playwright issue, present on main
- [x] `bun run build` — clean (pre-existing `@barefootjs/test` TS error on main is unchanged)
- [x] Fixture sweep vs main — 4/181 files changed, all reductions, all justified

### DIFF
--- a/packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts
@@ -3,19 +3,26 @@
  * prop bindings on child components — `<Card title={expr} />`. Follow-up
  * to #937 (architecture) and #939 (text interpolation pilot).
  *
- * Before this change, `collect-elements.ts` gated `reactiveChildProps`
- * on `hasPropsRef || needsEffectWrapper(...)`. The latter is an
- * allow-list that only recognises known signal getters, memos, and
- * prop parameter names. Any child-component prop expression the
- * analyzer couldn't prove reactive — e.g. `<Card title={formatTitle(
- * page)} />` where `formatTitle` is imported — was silently dropped.
- * At SSR the prop was evaluated once; the child component's DOM stayed
- * frozen at its initial render on the client.
+ * Before #942, `collect-elements.ts` gated `reactiveChildProps` on
+ * `hasPropsRef || needsEffectWrapper(...)`. The latter is an allow-list
+ * that only recognises known signal getters, memos, and prop parameter
+ * names. Any child-component prop expression the analyzer couldn't
+ * prove reactive — e.g. `<Card title={formatTitle(page)} />` where
+ * `formatTitle` is imported — was silently dropped. At SSR the prop was
+ * evaluated once; the child component's DOM stayed frozen at its
+ * initial render on the client.
  *
- * The fix adds a third branch to the OR gate: `/\b\w+\s*\(/.test(
- * expandedValue)`. Over-wrap (extra createEffect that subscribes to
- * nothing) is harmless — one closure at init time. Under-wrap is the
- * silent-drop bug we're closing.
+ * #942 originally closed this by adding a third branch to the OR gate:
+ * `/\b\w+\s*\(/.test(expandedValue)` with string-literal stripping to
+ * avoid false matches on `{ color: 'hsl(...)' }`. #952 then
+ * DRY-consolidated #942 into the same AST-flag shape that #939 / #941
+ * / #943 already used: the gate now reads
+ * `prop.callsReactiveGetters || prop.hasFunctionCalls` and both flags
+ * are computed Phase 1 over the prop's source AST expression — no
+ * post-expansion string scan, no string-strip workaround. Over-wrap
+ * (extra createEffect that subscribes to nothing) stays harmless;
+ * under-wrap (silent drop of a reactive read in a child prop) stays
+ * the class of bug we're closing.
  */
 
 import { describe, test, expect } from 'bun:test'
@@ -187,6 +194,44 @@ describe('Solid-style wrap-by-default fallback for child-component props (#942)'
     `
 
     const clientJs = getClientJs(source, 'Palette.tsx')
+    expect(clientJs).not.toContain('Reactive child component props')
+  })
+
+  test('local-const arrow binding stays un-wrapped (#952 source-level vs post-expansion guard)', () => {
+    // #952 replaced the post-expansion regex gate with AST flags computed
+    // on the prop's source expression. This changes behaviour for a
+    // specific shape that the fixture sweep surfaced empirically but no
+    // unit test locked in: a child-component prop whose value is a bare
+    // identifier bound to a local const whose initializer contains a
+    // call or constructor.
+    //
+    // Source: `<DatePicker formatDate={fmtDate} />` where
+    // `const fmtDate = (d) => d.toLocaleDateString(...)`. The #942 regex
+    // expanded `fmtDate` first, then matched `toLocaleDateString(` on
+    // the expansion — forcing wrap. The #952 AST flags see only the
+    // source identifier `fmtDate` (no CallExpression), so
+    // `hasFunctionCalls` is false and the prop stays un-wrapped. That is
+    // the correct semantic under #937: wrap based on what the prop's
+    // source expression *does*, not on what its transitive inlining
+    // happens to contain. The child component still receives the
+    // function value once via `initChild`'s `get formatDate()` property.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { DatePicker } from './DatePicker'
+
+      export function Form() {
+        const [, setFoo] = createSignal(0)
+        const fmtDate = (d: Date) => d.toLocaleDateString('en-US', { month: 'short' })
+        return (
+          <div onClick={() => setFoo(1)}>
+            <DatePicker formatDate={fmtDate} />
+          </div>
+        )
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Form.tsx')
     expect(clientJs).not.toContain('Reactive child component props')
   })
 
--- a/packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/child-prop-fallback-wrap.test.ts
@@ -158,6 +158,38 @@ describe('Solid-style wrap-by-default fallback for child-component props (#942)'
     expect(clientJs).not.toContain('Reactive child component props')
   })
 
+  test('string-literal-only function-like pattern stays un-wrapped (AST-flag structural check, #942 DRY)', () => {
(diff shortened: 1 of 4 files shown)


kfly8 (commented):
## Overview

Closes the last regex-based outlier in the #937 Solid-style wrap-by-default rollout. Adds `callsReactiveGetters` / `hasFunctionCalls` to `IRProp` (mirroring `IRExpression` / `IRConditional`), computes them in Phase 1 via the existing `exprCallsReactiveGetters` / `exprHasFunctionCalls` helpers, and swaps the Phase-2 string regex (`/\b\w+\s*\(/` over string-stripped `expandedValue`) for an OR on those flags. +67/-26, 4 files, CI green.

## Strengths

- **Naming / shape parity is exact.** `IRExpression.callsReactiveGetters`, `IRConditional.callsReactiveGetters`, `IRProp.callsReactiveGetters` — same names, same `boolean | undefined` encoding, same doc-comment wording (`types.ts:120-123`, `types.ts:139-142`, `types.ts:395-398`). A future reader grepping for either flag finds all five sites at once.
- **Helpers are reused verbatim.** `processComponentProps` calls the same `exprCallsReactiveGetters(...)` / `exprHasFunctionCalls(...)` that #939 / #941 / #943 consume (`jsx-to-ir.ts:73-108`). No new traversal to maintain, no new false-positive surface.
- **Structural pinning test chosen well.** `hsl(221 83% 53%)` inside an object-literal prop is exactly the shape that the previous "strip-strings-then-regex" approach had to opt into defensively. The AST flag doesn't need that opt-in — the StringLiteral is structurally not a CallExpression, so `hasFunctionCalls` is false by construction. The test comment makes the structural-vs-regex distinction explicit.
- **Existing 6 tests keep their assertion surface.** `count()`, `formatTitle(page)`, `variantFor(status)`, `props.title` all still wrap; bare identifier `label` and string literal `"Static"` still don't. All flow through the new `prop.hasFunctionCalls` / `prop.callsReactiveGetters` OR branches without test changes — good signal that the behavior shape is preserved where it should be.
- **Fixture sweep supports the behavioural change honestly.** 4/181 files change (2.2%), all reductions, -1771 bytes overall. Each delta is a case where the old regex fired on a post-expansion call (`today`, `maxDate`, `isWeekend`, `fmtDate`) but the AST flag on the original source expression — a bare identifier — does not. That's the correct semantic under the #937 spec: wrap based on what the prop's **source expression** does, not on what its transitive inlining happens to contain.
- **`hasPropsRef` deliberately retained with documented rationale** (`collect-elements.ts:433-439`). `props.xxx` access is a `PropertyAccessExpression`, not a `CallExpression`, so it wouldn't be caught by either AST flag. The comment flags this explicitly so the next pass knows what's left.

## Concerns (non-blocking)

### 1. Behaviour shift past the 4 fixture files isn't pinned

The fixture sweep found 4 compiled outputs where the old regex wrapped and the new AST flags don't. That's the intended improvement — but there's no unit test pinning a non-wrap for that pattern (e.g. `const fmtDate = (d) => format(d); <DatePicker formatDate={fmtDate} />`). Today's behaviour is verified by reading the byte delta; tomorrow's regression is not guarded.

Suggestion: add one test along the lines of `"local-const arrow function prop stays un-wrapped (AST-flag source-level check)"` — exact mirror of the hsl test but for the `fmtDate` shape. That makes the expansion-doesn't-leak-through behaviour a structural invariant, not just an empirical observation from the sweep. This also documents to future #944 / why-wrap implementers *what wrap-by-default does not cover*.

### 2. Top-of-file docstring on `child-prop-fallback-wrap.test.ts` is now stale

Lines 15-16 still describe the fix as "adds a third branch to the OR gate: `/\b\w+\s*\(/.test(expandedValue)`". After this PR the third and fourth branches are `prop.callsReactiveGetters || prop.hasFunctionCalls`, and the regex is gone. A reader searching for that regex will find it in the comment but not in the code.

Suggestion: update the header docstring to say something like *"Originally gated on `/\b\w+\s*\(/` over stripped text (#942); DRY-consolidated in #952 to use `IRProp.callsReactiveGetters` / `IRProp.hasFunctionCalls` AST flags, matching #939 / #941 / #943."* One paragraph, keeps the history discoverable.

### 3. `hasPropsRef` is the last non-AST check

Not a concern for this PR — the comment already flags it — but worth naming as the follow-up target if the "no Phase-2 string heuristics" goal ever becomes explicit. `expandedValue.includes('props.')` will still false-match on e.g. `<Card x={'props.title is nice'} />` (expanded to `'props.title is nice'`). Harmless (over-wrap), but it's the same class of bug the #952 commit sets out to eliminate. A future extension could add `IRProp.accessesProps` computed via `ts.isPropertyAccessExpression` walk. Not here, not now.

- **Correctness**: Widened coverage for call-shaped props is preserved exactly; contracted coverage (post-expansion calls that don't appear in source) is reported as byte reductions, all justified. No silent-drop class reopened — the wrap-by-default AST flags are strictly more conservative for source-level shapes than the previous post-expansion regex.
- **Performance**: AST flags are computed once per prop during `processComponentProps`, replacing a regex + string replace per prop during `collectElements`. Net slightly cheaper. Compile time unaffected at CI budget (1647 pass).
- **Security**: Internal compiler refactor, no user input surface. N/A.

## Verdict

**Ready to merge.** This is the natural closing move on the #937 rollout arc — all five sites now use the same AST-flag shape, making future audits (and the eventual #944 `why-wrap` CLI) straightforward: one pair of helpers, one pair of flags, five emission sites. Concerns 1–3 are follow-up nits.

After merge, the wrap-by-default rollout is architecturally uniform. Remaining queue per earlier planning:

1. #944 (`barefoot why-wrap` CLI) — now has a uniform flag shape to introspect.
2. #951 (Option 3, same-key destructured loop refresh) — load-bearing only if a real user report surfaces.
3. Optional: `IRProp.accessesProps` to retire `hasPropsRef`'s string check (per concern 3).

kfly8 (commented):
## Re-review (round 2)

One follow-up commit: `d448ce6f test(compiler): address PR #952 review — pin local-const expansion, refresh docstring`. CI green.

### Addresses

|-------|--------|
| **#1 Behaviour shift past the 4 fixture files isn't pinned** | **Fixed.** New test `local-const arrow binding stays un-wrapped (#952 source-level vs post-expansion guard)` uses exactly the shape I suggested: `<DatePicker formatDate={fmtDate} />` with `const fmtDate = (d: Date) => d.toLocaleDateString(...)`. Asserts the prop is NOT in `reactiveChildProps`. This is the clean counter-pin to the hsl() test — one locks in the "string-in-object → structurally not a call" invariant, the other locks in the "bare identifier → AST sees no call, does not follow expansion" invariant. Both failure modes a future refactor could silently re-introduce are now caught at test-time. |
| **#2 Top-of-file docstring stale** | **Fixed.** Docstring now narrates the full #942 → #952 progression explicitly: (a) the original regex-plus-strip fix, (b) the DRY consolidation onto `prop.callsReactiveGetters || prop.hasFunctionCalls`, (c) "no post-expansion string scan, no string-strip workaround". The old regex phrase is preserved as a historical note so a reader grepping for it still lands on the commit arc, but it's no longer described as the live gate. |
| #3 `hasPropsRef` string-check | Explicitly flagged in the commit message as follow-up — not in scope. Correct call; rolling it into this PR would have expanded the blast radius beyond the #942 DRY goal. |

### Delta verification

- New test's comment makes the semantic shift explicit: "wrap based on what the prop's source expression *does*, not on what its transitive inlining happens to contain." That's the #937 spec verbatim applied to child props. Future readers understand **why** the un-wrap is correct, not just that it is.
- Docstring's phrasing "under-wrap (silent drop of a reactive read in a child prop) stays the class of bug we're closing" preserves the #937 framing — this is still a wrap-by-default fallback site, not a coverage retreat.
- Zero source-code change since round 1; only test + comments. No new risk introduced.

### Verdict

**Ready to merge.** Every concern from round 1 is addressed in the minimum-blast-radius way: concerns 1 and 2 got code-level pins (new test + refreshed docstring), concern 3 got an explicit follow-up label. All 5 wrap-by-default sites now share the same IR-level flag shape, and the structural-vs-regex distinction is documented from both axes (hsl(): call-shaped literal stays false; fmtDate: identifier-with-expansion-to-call stays false).

After merge, the #937 rollout's architectural consolidation is complete. The queue stays:

1. **#944** (`barefoot why-wrap` CLI) — uniform flag shape is now a single introspection target across text / attr / conditional / child-prop / loop.
2. **#951** (Option 3, same-key destructured loop refresh) — load-bearing only if a user report surfaces.
3. Optional future DRY+: retire `hasPropsRef`'s string check via `IRProp.accessesProps` computed from `ts.isPropertyAccessExpression`.

PR #953, opened 2026-04-21
refactor(compiler): unify attribute wrap-by-default on AST flags (#940 DRY)

kfly8:
## Summary

- Close the last regex-based outlier in the #937 Solid-style wrap-by-default rollout: JSX attribute bindings (#940) now gate on `IRAttribute.callsReactiveGetters` / `IRAttribute.hasFunctionCalls`, the same AST-derived flags #939 / #941 / #942 / #943 already use on `IRExpression`, `IRConditional`, and `IRProp` (after #952).
- Remove the `/\b\w+\s*\(/` regex from `collect-elements.ts`. AST flags are computed Phase 1 on the attribute's source JSX expression, so they can't false-match call-like substrings inside string literals (e.g. `style={{ color: 'hsl(221 83% 53%)' }}`) and don't depend on the expansion order of local constants.
- Add a structural pinning test: `style={{ color: 'hsl(221 83% 53%)' }}` stays un-wrapped because `hsl(...)` is a `StringLiteral` inside an `ObjectLiteralExpression`, not a `CallExpression`.

## Context

Prep for #944 (`barefoot why-wrap` CLI). Once #944's `collectDomBindings` widens to surface fallback-wrapped expressions, it benefits from every emit site reading the same flag shape — one gate, one code path. This PR is the last stand-alone DRY refactor before #944; no CLI changes here.

`needsEffectWrapper(expandedValueStr, ctx)` stays as the string-level check against expanded text, matching what #952 kept for `IRProp`. It still catches known signal getters / memos / props inside transitively-inlined local-const references like `const classes = \`btn \${count()}\`; <div class={classes}>`. The AST flags live on the source expression; the expanded check lives on post-expansion text. Both branches are needed for full coverage.

## Behaviour change

Local-const identifiers whose initializers contain calls but no reactive reads (e.g. `const classes = \`\${base} \${fmt(v)}\`; <div class={classes}>`) no longer wrap. The source AST sees only the identifier `classes` (no `CallExpression`), and the expansion has no reactive source — so the attribute stays frozen in SSR, which is correct: nothing to track.

This is the attribute-side analogue of the `formatDate={fmtDate}` shift in #952. The existing `spread-expanded prop whose expansion contains a call wraps` test locked in the old post-expansion wrap; renamed to `local-const identifier attribute stays un-wrapped` and flipped to assert the new semantic, consistent with #952's `local-const arrow binding stays un-wrapped` child-prop pinning test.

## Fixture sweep (site/ui/dist/components)

508 files total, **0 changed** (total bytes identical main vs PR: 28078189). No site/ui component currently hits the behaviour-change shape (local-const template-literal-with-call used as an HTML attribute value, without any reactive source). The regression surface is covered by unit tests.

## Test plan

- [x] `bun test packages/jsx` — 693 pass (all 5 fallback-wrap suites + new hsl pinning + renamed local-const test)
- [x] `bun test --cwd . packages/` — 1649 pass; remaining 1 fail / 1 error is the pre-existing `packages/xyflow/e2e/flow.spec.ts` Playwright `beforeEach` issue present on main (same state as #952)
- [x] `bun run build` — clean
- [x] Fixture sweep — 0 files changed, total bytes identical

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/jsx/src/__tests__/attribute-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/attribute-fallback-wrap.test.ts
@@ -11,11 +11,14 @@
  * attribute was evaluated once; the client never re-evaluated it, so the
  * attribute stayed frozen.
  *
- * The fix widens the gate to also wrap when the expanded expression contains
- * a function call (`/\b\w+\s*\(/`). Over-wrapping a pure call that
- * subscribes to nothing is harmless (one closure at init time, no rerun);
- * under-wrapping a reactive read is the bug class we're closing. Pure static
- * literals and bare identifiers without calls stay un-wrapped.
+ * #940 originally closed this by adding a regex fallback
+ * (`/\b\w+\s*\(/`) with quote-strip on the expanded value. The DRY
+ * consolidation (follow-up to #952 for `IRProp`) replaced that
+ * post-expansion regex with AST flags computed Phase 1 on the source
+ * JSX expression — matching the shape #939 / #941 / #942 / #943 already
+ * use. Over-wrap (extra createEffect that subscribes to nothing) stays
+ * harmless; under-wrap (silent drop of a reactive attribute read) stays
+ * the class of bug we're closing.
  */
 
 import { describe, test, expect } from 'bun:test'
@@ -134,12 +137,27 @@ describe('Solid-style wrap-by-default fallback for attributes (#940)', () => {
     expect(clientJs).not.toMatch(/className\s*=\s*[^;]*cls/)
   })
 
-  test('spread-expanded prop whose expansion contains a call wraps', () => {
-    // `classes` is a local const whose value contains a call
-    // (`format(variant)`). expandConstantForReactivity inlines the reference
-    // before the gate runs, so the expanded string matches the call regex
-    // and the attribute wraps. This protects the same silent-drop case for
-    // code that factors class construction into a local variable.
+  test('local-const identifier attribute stays un-wrapped (DRY source-level vs post-expansion guard)', () => {
+    // DRY consolidation replaced the post-expansion regex gate with AST
+    // flags computed on the attribute's source expression. This changes
+    // behaviour for a specific shape: a local-const identifier whose
+    // initializer contains a call, used as an attribute value.
+    //
+    // Source: `class={classes}` where
+    // `const classes = \`\${base} \${format(variant)}\``. The old regex
+    // expanded `classes` first, then matched `format(` on the expansion
+    // — forcing wrap. The AST flags see only the source identifier
+    // `classes` (no CallExpression), so `hasFunctionCalls` is false.
+    // `needsEffectWrapper` on the expanded text also stays false
+    // because `base`, `format`, and `variant` aren't registered as
+    // reactive (no signals/memos/props match). The attribute stays
+    // un-wrapped.
+    //
+    // That is the correct semantic under #937: wrap based on what the
+    // attribute's source expression *does*, not on what its transitive
+    // inlining happens to contain. The SSR-rendered attribute is
+    // already frozen into the DOM and none of the transitive reads are
+    // reactive, so the previous wrap was dead weight.
     const source = `
       'use client'
       import { createSignal } from '@barefootjs/client'
@@ -155,7 +173,33 @@ describe('Solid-style wrap-by-default fallback for attributes (#940)', () => {
     `
 
     const clientJs = getClientJs(source, 'Tag.tsx')
-    expect(clientJs).toContain('createEffect')
-    expect(clientJs).toContain('format(')
+    expect(clientJs).not.toContain('createEffect')
+  })
+
+  test('string-literal-only function-like pattern stays un-wrapped (AST-flag structural check)', () => {
+    // Before DRY consolidation, collect-elements.ts scanned the expanded
+    // attribute value with /\b\w+\s*\(/ after stripping quoted strings.
+    // A structural regex over stripped text is still a regex — any future
+    // regression in the strip step would silently re-introduce hsl / rgb
+    // / url / etc. false positives. The AST flag approach can't be fooled
+    // this way: `{ color: 'hsl(221 83% 53%)' }` is an object literal with
+    // a StringLiteral value, not a CallExpression, so `hasFunctionCalls`
+    // is structurally false.
+    //
+    // The enclosing expression has no reactive source (no signal, no
+    // memo, no \`props.\`), so the attribute must NOT appear in
+    // reactiveAttrs.
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      export function Palette() {
+        const [, setFoo] = createSignal(0)
+        return <div style={{ color: 'hsl(221 83% 53%)' }} onClick={() => setFoo(1)}>x</div>
+      }
+    `
+
+    const clientJs = getClientJs(source, 'Palette.tsx')
+    expect(clientJs).not.toContain('createEffect')
   })
 })
--- a/packages/jsx/src/__tests__/attribute-fallback-wrap.test.ts
+++ b/packages/jsx/src/__tests__/attribute-fallback-wrap.test.ts
@@ -137,7 +137,7 @@ describe('Solid-style wrap-by-default fallback for attributes (#940)', () => {
     expect(clientJs).not.toMatch(/className\s*=\s*[^;]*cls/)
   })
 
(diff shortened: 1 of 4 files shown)


kfly8 (commented):
## Overview

Closes the last regex-based outlier in the #937 Solid-style wrap-by-default rollout. Adds `callsReactiveGetters` / `hasFunctionCalls` to `IRAttribute` (mirroring `IRExpression` / `IRConditional` / `IRProp`), computes them in Phase 1 via the existing `exprCallsReactiveGetters` / `exprHasFunctionCalls` helpers, and swaps the Phase-2 `/\b\w+\s*\(/` regex on expanded attribute text for an OR on those flags. `needsEffectWrapper` stays as the string-level check for expansion-through-local-const cases, matching what #952 kept for `IRProp`. +101/-25, 3 files, CI green, 0 fixture-byte delta.

This is PR-A as scoped in the prompt for #944 — the last DRY consolidation before the `why-wrap` CLI lands on a uniform five-site flag shape.

## Strengths

- **Exact shape parity with #952.** Same flag names (`callsReactiveGetters` / `hasFunctionCalls`), same `boolean | undefined` encoding, same doc-comment wording, same `exprCallsReactiveGetters(...)` / `exprHasFunctionCalls(...)` helper reuse (`jsx-to-ir.ts:1885-1889`, compare `jsx-to-ir.ts:2155-2158` for IRProp). All five emission sites now read one pair of flags — anyone grepping for either symbol finds the full wrap-by-default surface.
- **Gate structure is the `IRProp`-in-#952 analogue, intentionally.** `needsEffectWrapper(...) || attr.callsReactiveGetters || attr.hasFunctionCalls` leaves the expansion-based reactivity check in place for `const classes = \`btn \${count()}\`; <div class={classes}>` — source AST sees only the identifier, but `needsEffectWrapper` on expanded text catches the signal read. The comment at `collect-elements.ts:577-586` documents this split explicitly, and a reader who only looks at one half won't accidentally remove the other.
- **Behaviour-change test was flipped cleanly.** The old `spread-expanded prop whose expansion contains a call wraps` asserted `toContain('createEffect')` + `toContain('format(')`. It now asserts `not.toContain('createEffect')` under the new name `local-const identifier attribute stays un-wrapped (DRY source-level vs post-expansion guard)`. A reviewer can see the semantic inversion in a single test hunk; the test name carries its own rationale. This is the exact mirror of the #952 `fmtDate` flip.
- **Structural pinning added for the hsl() shape.** New test `string-literal-only function-like pattern stays un-wrapped` pins that `style={{ color: 'hsl(221 83% 53%)' }}` cannot regress into false-positive wrapping — the StringLiteral inside ObjectLiteralExpression is structurally not a CallExpression. Same rationale as #952's hsl test, now covering attributes too.
- **Header docstring on `attribute-fallback-wrap.test.ts` refreshed to narrate #940 → DRY consolidation.** Historical regex phrase is preserved so a reader can still trace the commit arc; the live gate description is updated to AST flags. No stale references to `/\b\w+\s*\(/` as the current mechanism.
- **0 / 508 fixture files changed, byte-identical (28,078,189 bytes).** Honest explanation in the PR body: no site/ui component currently uses the `const classes = \`…${call()}\`; <div class={classes}>` shape with no reactive source. The behaviour delta is reserved for the unit test — a justified "tests only" surface when the fixture corpus doesn't exercise the changed code path.
- **PR body calls out the behaviour shift up front.** The "Behaviour change" section explicitly names this as the attribute-side analogue of #952's `formatDate={fmtDate}` flip. That framing helps anyone cross-referencing the two PRs identify them as one conceptual change applied to two IR nodes.

## Concerns (non-blocking)

### 1. New test names don't carry the PR number, breaking the #952 convention

#952 named its two pinning tests `(#952 source-level vs post-expansion guard)` and `(AST-flag structural check, #942 DRY)` — embedding the PR number in the assertion name. Makes test-log greps like `grep "#952"` surface the originating change.

This PR's tests use `(DRY source-level vs post-expansion guard)` and `(AST-flag structural check)` — same structure but missing the ID. Future archaeology on "why was this test flipped" will have to dig through git blame instead of matching test name to PR.

Suggestion: rename to `(#953 source-level vs post-expansion guard)` and `(#953 AST-flag structural check)` before merging. One-line change, preserves the convention across #952 / #953 as a pair.

### 2. `not.toContain('createEffect')` is broader than necessary on the flipped test

```ts
expect(clientJs).not.toContain('createEffect')
```

Asserts the *entire* compiled module has no `createEffect` anywhere, which is true today for the `Tag` component, but fragile: if a future optimisation wraps something unrelated (e.g. a new optimisation on `{variant}` text interpolation), the test fails for the wrong reason.

Suggestion: tighten to `expect(clientJs).not.toMatch(/createEffect\(\s*\(\)\s*=>[^)]*?classes/)` or similar — assert that `classes` specifically isn't a createEffect target. The positive branch (`local-const identifier attribute stays un-wrapped`) is what this test is really about; the narrower regex pins that without false-failing on unrelated effects.

### 3. Zero fixture-delta could use one confirmatory counter-example

The 0/508 sweep is a defensible outcome — the pattern just isn't used. But one thing the sweep doesn't prove is that the **opposite** direction (cases that previously didn't wrap and now do) is also empty. For a defensive sanity check, it'd be reassuring to have spot-checked one or two components known to use `attr={signalGetter()}` or `attr={memo()}` and confirmed they still wrap exactly as before. This is a paranoia check, not a blocker — the existing test coverage already pins those directions, and CI passed all existing test suites.

If the author did this spot-check locally, a one-line mention in the PR body ("verified `<Button disabled={isLoading()}>` etc. still wrap via AST flags") would close the loop.

- **Correctness**: Widened coverage for call-shaped attribute expressions is preserved exactly via `attr.hasFunctionCalls`; contracted coverage (post-expansion calls not reachable via AST) is intentional and reported as 0-byte fixture delta. The structural-vs-regex distinction is now uniform across five emission sites — same guarantee level, same failure mode, same fix shape if a new IR node type is ever added.
- **Performance**: AST flags computed once per attribute during `processAttributes`, replacing a regex + quote-strip per attribute during `collectElements`. Net slightly cheaper. Compile time unchanged at CI budget (1649 tests passing).
- **Security**: Internal compiler refactor, no user input surface. N/A.

## Verdict

**Ready to merge.** This is the closing move on the DRY arc — five sites, one flag shape, one failure mode, one mental model. After this lands, PR-B (#944 `why-wrap` CLI) can consume a uniform `callsReactiveGetters || hasFunctionCalls` signal across text / attr / conditional / child-prop (and loop once `IRLoop` gets similar flags per the PR-B prompt). Concerns 1–3 are nits — concern 1 is the most worth picking up before merge because it's a one-line convention fix.

Post-merge state:

1. **PR-B (#944)** — `why-wrap` CLI, now implementable against a uniform flag shape. Loop-side AST flags may still need to be added as part of PR-B per the prompt's judgment call.
2. #951 (Option 3 same-key destructured loop refresh) — load-bearing only if a user report surfaces.
3. Optional: retire `hasPropsRef` (IRProp) and `needsEffectWrapper` string-check (IRProp / IRAttribute) behind `IRProp.accessesProps` / `IRAttribute.accessesProps` — would fully eliminate Phase-2 string heuristics. Not here, not now.

kfly8:
Addressed the three non-blocking concerns in [3d21d1fd](https://github.com/barefootjs/barefootjs/commit/3d21d1fd):

**Concern 1 (test names):** renamed to `(#953 source-level vs post-expansion guard)` and `(#953 AST-flag structural check)` so the #952 ↔ #953 pair greps cleanly by PR number.

**Concern 2 (`not.toContain('createEffect')` too broad):** replaced with a tighter pin on the actual regression surface — `not.toContain('Reactive attributes')` (the emitter-stable section header for the reactive-attrs block) plus a DOM-op guard specific to the attribute under test (`class` / `style`). Future unrelated optimisations elsewhere in the compiled module won't false-fail these tests now.

**Concern 3 (opposite-direction spot check):** verified locally. Compiling

```tsx
'use client'
import { createSignal, createMemo } from '@barefootjs/client'
export function SignalAttr() {
  const [isLoading, setLoading] = createSignal(false)
  const label = createMemo(() => isLoading() ? 'Loading...' : 'Ready')
  return <button disabled={isLoading()} aria-label={label()} onClick={() => setLoading(l => !l)}>click</button>
}
```

still emits `createEffect` wrapping `_s0.disabled = !!(isLoading())` and the `aria-label` setter — both the signal-getter case (`disabled={isLoading()}`) and the memo case (`aria-label={label()}`) hit the wrap path as before, via `attr.callsReactiveGetters` firing on the source AST and `needsEffectWrapper` also firing on the expanded text. No direction regressed.

PR #956, opened 2026-04-21
feat(cli): barefoot why-wrap — surface fallback-wrapped expressions (#944)

### LINKED ISSUE #944: debug: surface wrap-by-default fallback expressions in `barefoot inspect` / `why-update`

kfly8:
## Summary

- New subcommand `barefoot why-wrap <component>` lists every DOM binding whose `createEffect` came from the Solid-style wrap-by-default fallback (#937) rather than from statically-proven reactivity — these are the optimisation candidates that were previously invisible.
- `DomBinding` gains `classification: 'reactive' | 'fallback'` and `expression?: string` so `barefoot inspect` / `why-update` / `why-wrap` all see the same reactive footprint the runtime sees, and CLI output can print the actual expression. `collectDomBindings` is widened across text (#939), attribute (#940), conditional (#941), loop (#943), **and child-component prop (#942/#952)** to match every emitter's wrap decision.
- `IRLoop` gains `callsReactiveGetters` / `hasFunctionCalls` (mirroring #939/#941/#942/#952-on-`IRProp` and #953-on-`IRAttribute`) so loop bindings classify the same way as the other four sites.
- `formatComponentGraph` prefixes fallback bindings with `~`. `--json` carries `classification` and `expression` verbatim for tooling.

Closes #944.

## Why CLI, not warnings

Decided in #937 and the `project_priorities` memory: dev-time warnings are ignored or become noise. Opt-in CLI is the visibility channel — matches the shape of existing `inspect` / `why-update` / `test --debug` tooling. `why-wrap` surfaces the information; the user decides what to rewrite.

## Manual verification

Running against `calendar` (touches text / attribute / conditional across both sub-months):

```
$ bun run barefoot why-wrap calendar
Calendar — 8 fallback-wrapped expression(s)
  attribute "className"  ~ numMonths() > 1 flex gap-4
  attribute "className"  ~ getDayClasses(day, isSelected, rangePos)
  attribute "data-date"  ~ toISODateString(day.date)
  text "s8"              ~ day.date.getDate()
  conditional "s12"      ~ numMonths() >= 2
  attribute "className"  ~ getDayClasses(day, isSelected, rangePos)
  attribute "data-date"  ~ toISODateString(day.date)
  text "s21"             ~ day.date.getDate()

Fallback wraps run one createEffect per expression. Each subscribes to
whatever signals it happens to read at runtime (possibly none).
Rewrite the expression as a createMemo to make the dependency static,
or inline a known-reactive source so the emitter can prove it.
```

Child-component prop case verified via a synthetic `<Card title={formatTitle(page)} />` component — output prints `attribute "Card.title"  ~ formatTitle(page)`, confirming the #942 motivating example now surfaces end-to-end.

`barefoot inspect calendar` shows `~` markers on the same bindings. `--json` emits `classification` and `expression` on every `domBindings` entry. `why-wrap --json` filters to fallbacks only.

## Commit-level breakdown

1. `feat(compiler): classify DomBindings as reactive vs fallback` — `DomBinding.classification` field, `collectDomBindings` gate widening across text / attribute / conditional / loop, `IRLoop` flags + `jsx-to-ir` population, `formatComponentGraph` marker, `graphToJSON` classification.
2. `test(compiler): pin DomBinding reactive/fallback classification` — 8 `debug.test.ts` cases covering text/attribute/conditional/loop × reactive/fallback, plus `formatComponentGraph` marker pinning and `graphToJSON` shape pinning.
3. `feat(cli): barefoot why-wrap` — command + dispatcher wiring + 5-case test suite.
4. `docs: document why-wrap and the ~ fallback marker` — `CLAUDE.md` + `docs/core/core-concepts/ai-native.md`.
5. `fix(debug): address PR #956 review` — adds the missing `case 'component'` branch for child-component props (concern 1), `DomBinding.expression?` with CLI output overhaul (concern 2), `await expect(...).rejects.toThrow()` on error tests (concern 3), swap non-existent `AnalyticsDemo` for real `calendar` in docs (concern 5). 4 new debug.test.ts cases lock in the new contract.

## Test plan

- [x] `bun test packages/jsx` — 706 pass (13 new cases in `debug.test.ts`)
- [x] `bun test packages/cli` — 506 pass (5 cases in `why-wrap.test.ts`, all with proper `await`)
- [x] `bun test --cwd . packages/` — 1667 pass; remaining 1 fail / 1 error is the pre-existing `packages/xyflow/e2e/flow.spec.ts` Playwright issue on main
- [x] `bun run build` — clean
- [x] Manual: `barefoot why-wrap calendar` shows 8 fallbacks with actual expression text
- [x] Manual: `barefoot why-wrap /tmp/child-prop-demo.tsx` surfaces `Card.title ~ formatTitle(page)` (concern 1 regression guard)
- [x] Manual: `barefoot why-wrap calendar --json` returns valid JSON with `classification` and `expression` on every entry
- [x] Manual: `barefoot inspect calendar` shows `~` marker on fallback bindings

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -55,8 +55,9 @@ Use the `barefoot` CLI (`bun run barefoot`) to look up component APIs, framework
 - `barefoot search <query>` — Find components and docs by name/category/tags
 - `barefoot ui <component>` — Component reference (props, examples, a11y)
 - `barefoot core <topic>` — Core docs (signals, compiler constraints, error codes, etc.)
-- `barefoot inspect <component>` — Show signal dependency graph (signals, memos, DOM bindings)
+- `barefoot inspect <component>` — Show signal dependency graph (signals, memos, DOM bindings). Bindings wrapped by the Solid-style fallback (#937) are prefixed with `~`.
 - `barefoot why-update <component> <signal>` — Trace update propagation path from a signal to DOM
+- `barefoot why-wrap <component>` — List fallback-wrapped expressions (#937). Each is a candidate for `createMemo` refactor — the runtime effect subscribes to whatever it reads at runtime, possibly nothing.
 - `barefoot test --debug <component>` — Show signal initialization trace and effect bindings
 
 Before editing a stateful component (`"use client"`), run `barefoot inspect` first to understand its reactive structure. All inspection commands support `--json` for machine-readable output.
--- a/docs/core/core-concepts/ai-native.md
+++ b/docs/core/core-concepts/ai-native.md
@@ -38,6 +38,7 @@ barefoot test:template Button                        # Generate IR test from exi
 # Inspect reactive structure
 barefoot inspect Counter             # Signal dependency graph
 barefoot why-update Counter count    # Trace update path: signal → DOM
+barefoot why-wrap AnalyticsDemo      # List Solid-style wrap-by-default fallback bindings
 ```
 
 Both humans and AI agents use these commands to generate and debug components without reading source files.
--- a/docs/core/core-concepts/ai-native.md
+++ b/docs/core/core-concepts/ai-native.md
@@ -38,7 +38,7 @@ barefoot test:template Button                        # Generate IR test from exi
 # Inspect reactive structure
 barefoot inspect Counter             # Signal dependency graph
 barefoot why-update Counter count    # Trace update path: signal → DOM
-barefoot why-wrap AnalyticsDemo      # List Solid-style wrap-by-default fallback bindings
+barefoot why-wrap calendar           # List Solid-style wrap-by-default fallback bindings
 ```
 
 Both humans and AI agents use these commands to generate and debug components without reading source files.
--- a/packages/cli/src/__tests__/why-wrap.test.ts
+++ b/packages/cli/src/__tests__/why-wrap.test.ts
@@ -82,6 +82,9 @@ describe('barefoot why-wrap', () => {
       // The text binding's slotId appears in the report line.
       expect(output).toMatch(/text "s\d+"/)
       expect(output).toContain('~')
+      // The actual expression text is printed so users can locate the
+      // binding in their source without running `inspect` separately.
+      expect(output).toContain('formatTitle(page)')
       // Guidance footer — helps the user know what to do next.
       expect(output).toContain('createMemo')
     } finally {
@@ -112,7 +115,7 @@ describe('barefoot why-wrap', () => {
       const parsed = JSON.parse(output) as {
         componentName: string
         sourceFile: string
-        fallbacks: Array<{ classification: string; type: string; label: string; deps: string[]; slotId: string }>
+        fallbacks: Array<{ classification: string; type: string; label: string; deps: string[]; slotId: string; expression?: string }>
       }
       expect(parsed.componentName).toBe('Tag')
       expect(parsed.fallbacks.length).toBeGreaterThan(0)
@@ -124,6 +127,7 @@ describe('barefoot why-wrap', () => {
       expect(attrFallback).toBeDefined()
       expect(attrFallback!.label).toBe('class')
       expect(attrFallback!.deps).toEqual([])
+      expect(attrFallback!.expression).toBe('format(label)')
     } finally {
       logSpy.mockRestore()
       rmSync(path.dirname(file), { recursive: true, force: true })
@@ -137,7 +141,7 @@ describe('barefoot why-wrap', () => {
     const errSpy = spyOn(console, 'error').mockImplementation(() => {})
     try {
       const { run } = await import('../commands/why-wrap')
-      expect(run(['NoSuchComponent12345'], makeCtx(false))).rejects.toThrow('exit 1')
+      await expect(run(['NoSuchComponent12345'], makeCtx(false))).rejects.toThrow('exit 1')
       expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Cannot find'))
     } finally {
       exitSpy.mockRestore()
@@ -152,7 +156,7 @@ describe('barefoot why-wrap', () => {
     const errSpy = spyOn(console, 'error').mockImplementation(() => {})
     try {
       const { run } = await import('../commands/why-wrap')
-      expect(run([], makeCtx(false))).rejects.toThrow('exit 1')
+      await expect(run([], makeCtx(false))).rejects.toThrow('exit 1')
       expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Component name required'))
       expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Usage:'))
     } finally {
--- a/packages/cli/src/__tests__/why-wrap.test.ts
+++ b/packages/cli/src/__tests__/why-wrap.test.ts
@@ -1,0 +1,163 @@
+/**
+ * Tests for `barefoot why-wrap <component>` — surface fallback-wrapped
+ * expressions emitted by Solid-style wrap-by-default (#937).
+ *
+ * Fallback wraps are harmless but invisible in the source: the compiler
+ * chose `createEffect` because it couldn't statically prove the
+ * expression is (or isn't) reactive. This CLI is the opt-in channel users
+ * rely on to find candidates for `createMemo` refactor; regressing its
+ * filter or its JSON shape silently hides the whole reactive footprint.
+ */
+
+import { describe, test, expect, spyOn } from 'bun:test'
+import { mkdtempSync, writeFileSync, rmSync } from 'fs'
+import { tmpdir } from 'os'
+import path from 'path'
+import type { CliContext } from '../context'
+
+function makeCtx(jsonFlag: boolean): CliContext {
+  // Minimal context — `resolveComponentSource` only needs `root`, `config`,
+  // `projectDir` to resolve a bare name; we always pass an absolute .tsx
+  // path below, so those fall back fields are irrelevant for this suite.
+  return {
+    root: process.cwd(),
+    metaDir: '',
+    jsonFlag,
+    config: null,
+    projectDir: null,
+  }
+}
+
+function tmpComponent(source: string, name = 'Demo.tsx'): string {
+  const dir = mkdtempSync(path.join(tmpdir(), 'barefoot-why-wrap-'))
+  const file = path.join(dir, name)
+  writeFileSync(file, source)
+  return file
+}
+
+describe('barefoot why-wrap', () => {
+  test('reactive-only component reports no fallbacks', async () => {
+    // Counter reads a signal — emitter wraps it but classification is
+    // 'reactive'. why-wrap's filter drops it; output is the "none" line.
+    const file = tmpComponent(`
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      export function Counter() {
+        const [count, setCount] = createSignal(0)
+        return <button onClick={() => setCount(c => c + 1)}>{count()}</button>
+      }
+    `)
+    const logSpy = spyOn(console, 'log').mockImplementation(() => {})
+    try {
+      const { run } = await import('../commands/why-wrap')
+      await run([file], makeCtx(false))
+      const output = logSpy.mock.calls.map(c => c[0]).join('\n')
+      expect(output).toContain('no fallback-wrapped expression')
+      expect(output).not.toContain('~')
+    } finally {
+      logSpy.mockRestore()
+      rmSync(path.dirname(file), { recursive: true, force: true })
+    }
+  })
+
+  test('component with fallback-wrapped text binding lists it', async () => {
+    // `formatTitle(page)` is the #939 canonical fallback shape — opaque
+    // call on a non-reactive argument.
+    const file = tmpComponent(`
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { formatTitle } from './format'
+      export function Page() {
+        const [, setFoo] = createSignal(0)
+        const page = 'home'
+        return <h1 onClick={() => setFoo(1)}>{formatTitle(page)}</h1>
+      }
+    `)
+    const logSpy = spyOn(console, 'log').mockImplementation(() => {})
+    try {
+      const { run } = await import('../commands/why-wrap')
+      await run([file], makeCtx(false))
+      const output = logSpy.mock.calls.map(c => c[0]).join('\n')
+      expect(output).toContain('1 fallback-wrapped expression')
+      // The text binding's slotId appears in the report line.
+      expect(output).toMatch(/text "s\d+"/)
+      expect(output).toContain('~')
+      // Guidance footer — helps the user know what to do next.
+      expect(output).toContain('createMemo')
+    } finally {
+      logSpy.mockRestore()
+      rmSync(path.dirname(file), { recursive: true, force: true })
+    }
+  })
+
+  test('--json emits machine-readable shape with classification', async () => {
+    // Editor integrations and CI scripts depend on the JSON shape.
+    // Lock in: componentName, sourceFile, fallbacks[].classification,
+    // fallbacks[].type, fallbacks[].slotId, fallbacks[].deps.
+    const file = tmpComponent(`
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+      import { format } from './fmt'
+      export function Tag() {
+        const [, setFoo] = createSignal(0)
+        const label = 'hi'
+        return <button class={format(label)} onClick={() => setFoo(1)}>x</button>
+      }
+    `)
+    const logSpy = spyOn(console, 'log').mockImplementation(() => {})
+    try {
+      const { run } = await import('../commands/why-wrap')
+      await run([file], makeCtx(true))
+      const output = logSpy.mock.calls.map(c => c[0]).join('\n')
+      const parsed = JSON.parse(output) as {
+        componentName: string
+        sourceFile: string
+        fallbacks: Array<{ classification: string; type: string; label: string; deps: string[]; slotId: string }>
+      }
+      expect(parsed.componentName).toBe('Tag')
+      expect(parsed.fallbacks.length).toBeGreaterThan(0)
+      // All entries must be fallbacks — reactive bindings are filtered out.
+      for (const f of parsed.fallbacks) {
+        expect(f.classification).toBe('fallback')
+      }
+      const attrFallback = parsed.fallbacks.find(f => f.type === 'attribute')
+      expect(attrFallback).toBeDefined()
+      expect(attrFallback!.label).toBe('class')
+      expect(attrFallback!.deps).toEqual([])
+    } finally {
+      logSpy.mockRestore()
+      rmSync(path.dirname(file), { recursive: true, force: true })
+    }
+  })
+
+  test('errors on unknown component', async () => {
+    const exitSpy = spyOn(process, 'exit').mockImplementation(((code?: number) => {
+      throw new Error(`exit ${code}`)
+    }) as never)
+    const errSpy = spyOn(console, 'error').mockImplementation(() => {})
+    try {
+      const { run } = await import('../commands/why-wrap')
+      expect(run(['NoSuchComponent12345'], makeCtx(false))).rejects.toThrow('exit 1')
+      expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Cannot find'))
+    } finally {
+      exitSpy.mockRestore()
+      errSpy.mockRestore()
+    }
+  })
+
+  test('errors with usage when no component argument', async () => {
+    const exitSpy = spyOn(process, 'exit').mockImplementation(((code?: number) => {
+      throw new Error(`exit ${code}`)
+    }) as never)
+    const errSpy = spyOn(console, 'error').mockImplementation(() => {})
+    try {
+      const { run } = await import('../commands/why-wrap')
+      expect(run([], makeCtx(false))).rejects.toThrow('exit 1')
+      expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Component name required'))
+      expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Usage:'))
+    } finally {
+      exitSpy.mockRestore()
+      errSpy.mockRestore()
+    }
+  })
+})
--- a/packages/cli/src/commands/why-wrap.ts
+++ b/packages/cli/src/commands/why-wrap.ts
@@ -47,6 +47,7 @@ export async function run(args: string[], ctx: CliContext): Promise<void> {
         deps: f.deps,
         type: f.type,
         classification: f.classification,
+        ...(f.expression !== undefined && { expression: f.expression }),
       })),
     }, null, 2))
     return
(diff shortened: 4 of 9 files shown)


kfly8 (commented):
## Overview

Ships `barefoot why-wrap <component>` as the opt-in visibility channel for Solid-style wrap-by-default fallbacks (#937), plus the underlying `DomBinding.classification: 'reactive' | 'fallback'` field that powers it. Widens `collectDomBindings` across **text / attribute / conditional / loop** to match each emitter's gate, adds `callsReactiveGetters` / `hasFunctionCalls` to `IRLoop` (mirroring #939/#941/#952/#953). `formatComponentGraph` prefixes fallbacks with `~`; `--json` carries the classification verbatim. +512/-20, 8 files, CI green.

This is PR-B of the #944 plan. PR-A (#953) landed the prerequisite IRAttribute AST flags.

## Strengths

- **IRLoop DRY finishes the AST-flag arc.** `transformMapCall` now computes both flags and reuses the already-needed `exprHasFunctionCalls(arrayExpr)` — one call instead of two — then stores them on `IRLoop` (`jsx-to-ir.ts:1696-1700`). Shape / naming / encoding match `IRProp` (#952), `IRAttribute` (#953), `IRExpression` (#939), `IRConditional` (#941). Four IR node types, one pair of flags: future audits and refactors have a single grep target.
- **Classification logic is consistent across the four collectDomBindings cases.** Same two-step pattern in every case: compute `isReactive` from static-proven deps, compute `isFallback` from the AST flags only when reactive is false, emit the binding only if either fires. No duplication, no drift. The `case 'loop'` version correctly handles three paths (signal-driven array via `extractReactiveDeps` over the array text, signal-driven via the AST flag itself, and opaque-call fallback) — all three get distinct classifications.
- **Event handlers correctly treated as always-reactive.** The old `if (deps.length > 0 || true)` idiom (an always-true conjunction left over from an earlier pass) is cleaned up to an unconditional push with `classification: 'reactive'`. The doc comment makes the reasoning explicit: handlers are bound once, not re-evaluated, so the wrap-by-default gate doesn't apply. This is the right call and the comment will prevent a future "why is this always reactive?" question.
- **`formatComponentGraph` marker uses constant column width.** `'~ '` for fallback, `'  '` for reactive — both 2 chars — so the rest of the line stays column-aligned. Downstream tools that parse the output won't suddenly see shifted columns depending on classification. The diff shows only the marker changes, not the field positions.
- **JSON shape is stable and test-pinned.** `why-wrap.test.ts` locks in `{componentName, sourceFile, fallbacks: [{classification, type, label, slotId, deps}]}`. Editor integrations / CI scripts depend on this shape; the pinning test makes accidental reshaping loud.
- **`debug.test.ts` 9 new cases exhaustively cover the matrix.** text × {reactive, fallback}, attribute × fallback, conditional × fallback, loop × {reactive, fallback}, event handler always-reactive, `graphToJSON` shape, `formatComponentGraph` marker. Each case is focused on one dimension; together they'd turn red on any silent regression of the classification axis.
- **CLI test uses the right test shape.** `mkdtempSync` + `writeFileSync` + absolute-path resolution side-steps the `resolveComponentSource` name-search logic so the tests only exercise `why-wrap`'s own logic. Usage-error and unknown-component paths both covered with `process.exit` mocked to throw.
- **Docs touched in the right two places.** `CLAUDE.md` CLI section mentions both the new command AND the `~` marker change to `inspect` — so AI agents reading CLAUDE.md know both UX shifts. `docs/core/core-concepts/ai-native.md` adds the command to the example block.

## Concerns

### 1. Child-component prop fallbacks (IRProp, the #942/#952 site) are silently missed

Confirmed by reading `types.ts:272-281` + grepping `debug.ts`: `IRComponent` is its own node type (`type: 'component'`, fields `name`, `props: IRProp[]`, `children`, …) produced by `transformComponentElement` / `transformSelfClosingComponent`. `collectDomBindings` has **no `case 'component'`** in its switch. Falls through silently — neither `node.props` gets iterated (missing the fallback classification) nor `node.children` gets recursed into.

Consequences:

- `<Card title={formatTitle(page)} />` — the exact motivating example for the #942 wrap-by-default work — doesn't appear in `why-wrap`'s output. The whole #952 DRY work that put `callsReactiveGetters` / `hasFunctionCalls` on `IRProp` is computed but not consumed here.
- Any fallback reached only through a component subtree (child nodes of `<Dialog>...</Dialog>`) is also invisible, because the switch falls through without recursing into `node.children`.
- The PR summary lists "text (#939), attribute (#940), conditional (#941), and loop (#943)" — 4 of the 5 wrap sites. #942 (child-prop) is absent. The manual verification on `calendar` finds 8 fallbacks; calendar plausibly also passes fallback props to internal sub-components, none of which are listed.

This is the exact bug class that `project_priorities` flags as "fidelity first, then client JS size" — a user running `why-wrap` to audit their component is told there are 0 fallbacks when in fact there are N on child components. That's a silent under-report.

**Suggestion**: add a `case 'component'` in `collectDomBindings`, structured like `case 'element'`:

```ts
case 'component': {
  for (const prop of node.props) {
    if (!prop.dynamic) continue
    const propValue = typeof prop.value === 'string' ? prop.value : ''
    if (!propValue) continue
    const deps = extractReactiveDeps(propValue, signalGetters, memoNames)
    const isReactive = deps.length > 0
    const isFallback = !isReactive && (prop.callsReactiveGetters || prop.hasFunctionCalls)
    if (isReactive || isFallback) {
      bindings.push({
        kind: 'dom',
        label: `${node.name}.${prop.name}`,  // or keep prop.name only
        slotId: node.slotId ?? '?',
        deps,
        type: 'attribute', // or a new 'child-prop' type — see below
        classification: isReactive ? 'reactive' : 'fallback',
      })
    }
  }
  for (const child of node.children) {
    collectDomBindings(child, bindings, signalGetters, memoNames)
  }
  break
}
```

Open design question: reuse `type: 'attribute'` vs introduce `type: 'child-prop'`. The existing `DomBinding.type` union is `'text' | 'event' | 'conditional' | 'loop' | 'attribute'`. Reusing `'attribute'` is cheapest (no new union variant, no CLI formatting changes), but loses the native-attr-vs-component-prop distinction. Adding `'child-prop'` would be a deliberate minor breaking change to consumers. I'd recommend reusing `'attribute'` with label prefix `"ComponentName.propName"` as the compromise.

Add a matching `debug.test.ts` case: `opaque call in child-component prop is fallback` mirroring the existing `opaque call in attribute is fallback` test, using the exact `<Card title={formatTitle(page)} />` shape.

### 2. Per-binding output line repeats the same suffix for every fallback

```
  ${f.type} "${id}"  ~ (wrapped by Solid-style fallback)
    The analyzer couldn't prove this expression's reactivity from a
    signal/memo/prop read, so a createEffect is emitted to be safe.
```

This prints two constant lines per fallback, regardless of which fallback. On the Calendar verification (8 fallbacks), that's 16 identical filler lines. The information a user actually needs is **the expression text** — so they can locate it in their source without running `inspect` separately.

`DomBinding` doesn't currently carry the expression text, only the slotId and label. But `collectDomBindings` has access to `node.expr` (text), `attr.value` (attribute), `node.condition` (conditional), `node.array` (loop) at collection time — none is propagated onto `DomBinding`.

**Suggestion** (can be a follow-up, not a blocker for this PR): add an optional `DomBinding.expression?: string` populated at collection time with the source expression. The `why-wrap` output becomes:

```
  text "s7"              ~ formatTitle(page)
  attribute "className"  ~ getClassName(props.layout)
```

One informative line per fallback, footer printed once. Drops the repeated "The analyzer couldn't prove..." lines entirely — the footer already says it.

This is the single highest-leverage UX improvement and is arguably the whole point of `why-wrap` over `inspect | grep ~`.

### 3. `why-wrap.test.ts` rejection-assertion not awaited

```ts
expect(run(['NoSuchComponent12345'], makeCtx(false))).rejects.toThrow('exit 1')
expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Cannot find'))
```

`expect(promise).rejects.toThrow(...)` returns a promise. Without `await` or `return`, the test function completes before the rejection assertion actually runs. It passes today because `process.exit` (mocked to throw) fires synchronously inside the `run` call, so `errSpy` is already populated by the time the (unawaited) rejection assertion is created. But a future refactor that moves the error path to async would silently break the test's primary assertion.

**Suggestion**:

```ts
await expect(run(['NoSuchComponent12345'], makeCtx(false))).rejects.toThrow('exit 1')
expect(errSpy).toHaveBeenCalledWith(expect.stringContaining('Cannot find'))
```

One-word fix, applies to both error tests (unknown-component + missing-arg).

### 4. Manual verification output in PR body doesn't match the actual format

PR body shows:
```
  attribute "className"  ~ (wrapped by Solid-style fallback)
  attribute "data-date"  ~ (wrapped by Solid-style fallback)
  text "s8"              ~ (wrapped by Solid-style fallback)
```

The actual command output at `why-wrap.ts:263-265` uses `${f.type} "${id}"` with no padding — so the output in practice is:

```
  attribute "className"  ~ (wrapped by Solid-style fallback)
  attribute "data-date"  ~ (wrapped by Solid-style fallback)
  text "s8"  ~ (wrapped by Solid-style fallback)
```

(The PR body output is edited for visual alignment.) Minor; future readers reproducing the example may wonder if something's wrong. If concern #2 is addressed (one informative line per fallback), this concern disappears naturally.

### 5. `AnalyticsDemo` referenced in `docs/core/core-concepts/ai-native.md` but likely not a real component

```
barefoot why-wrap AnalyticsDemo      # List Solid-style wrap-by-default fallback bindings
```

Docs hygiene — a reader copy-pasting this example gets "Cannot find component". Consider replacing with `calendar` (which the PR body already confirms has 8 fallbacks), or a generic placeholder like `MyComponent`.

- **Correctness**: The four widened cases match the emitter's gate exactly. IRComponent is the one case where the debug side lags the emitter — concern 1. No other silent drop identified.
- **Performance**: `collectDomBindings` adds at most one boolean OR per node; `exprCallsReactiveGetters`/`exprHasFunctionCalls` were already computed on IRLoop creation. Negligible cost. `why-wrap` runs only on user invocation; no effect on compile path.
- **Security**: CLI reads a user-specified file path via `resolveComponentSource`. No new surface beyond what `inspect` / `why-update` already expose. N/A.

## Verdict

**One blocker (concern 1) + four nits.** The child-component prop case is the most directly in-scope miss — the #952 PR specifically landed the flags this PR fails to consume, and the gap produces silent under-reporting in a visibility CLI, which is exactly the #944 anti-pattern. Adding `case 'component'` is ~15 lines of code + one test; highly worth fixing before merge so the Calendar verification output truly represents all fallbacks.

Concerns 2–5 are all improvements that don't block:
- **Concern 2** (expression text in output) is the highest-UX-leverage follow-up; can be a separate issue.
- **Concerns 3, 4, 5** are 1-line / doc-only cleanups; worth picking up together with concern 1 since the author is already iterating.

Once concern 1 is closed, this closes the #937 rollout's visibility arc. After merge, the remaining queue in the prompt is:
1. #951 (Option 3 same-key loop refresh) — load-bearing only if a user report surfaces.
2. Optional DRY+: `accessesProps` AST flag to retire `hasPropsRef` string check on `IRProp` / `IRAttribute`.

kfly8:
All five review concerns addressed in [83faaee1](https://github.com/barefootjs/barefootjs/commit/83faaee1). PR body updated to reflect the new output format.

**Concern 1 (blocker — IRComponent case):** Added `case 'component'` to `collectDomBindings` that iterates `node.props` with the same gate as `collect-elements.ts:442` (`hasPropsRef || deps.length > 0` → reactive; AST flags without reactive match → fallback) and recurses into `node.children`. Label uses `\"ComponentName.propName\"` — keeps the existing `DomBinding.type` union stable for consumers.

End-to-end verification:
```
$ barefoot why-wrap /tmp/child-prop-demo.tsx
ChildPropDemo — 1 fallback-wrapped expression(s)
  attribute \"Card.title\"  ~ formatTitle(page)
```
Three new `debug.test.ts` cases pin this: (a) opaque call in child prop → fallback with correct label/expression, (b) `props.xxx` child prop stays reactive (guards the `hasPropsRef` branch), (c) fallback inside a component subtree is still reached via recursion.

**Concern 2 (output repeated filler):** Added optional `DomBinding.expression?: string` populated at collection time across all five sites. `why-wrap` now column-aligns one line per fallback showing the actual expression. Calendar verification went from 24 lines (filler × 8) to 9 lines with expressions visible — e.g. `attribute \"data-date\"  ~ toISODateString(day.date)`. Also added a pinning test (`domBindings carry expression text for non-event sites`) locking the contract across text / attribute / conditional / loop / child-prop, and pinning that event handlers have no expression per the DomBinding doc.

**Concern 3 (rejection assertion not awaited):** `await expect(run(...)).rejects.toThrow(...)` on both error-path tests. The TS `'await' has no effect'` diagnostic is a Bun type imprecision — at runtime `rejects.toThrow` returns a thenable, and the await is what the review asked for to future-proof against an async refactor.

**Concern 4 (PR body output doesn't match reality):** Naturally resolved by concern 2 — PR body now shows the real column-aligned output format copy-pasted from the command.

**Concern 5 (AnalyticsDemo doesn't exist):** Swapped for `calendar` in `docs/core/core-concepts/ai-native.md` — calendar actually produces 8 fallbacks so the doc example demonstrates real output.

kfly8 (commented):
## Re-review (round 2)

One follow-up commit: `83faaee1 fix(debug): address PR #956 review — IRComponent case, expression field, tests`. Five review concerns, one surgical commit, CI green.

### Addresses

|---|---|
| **#1 (blocker) IRComponent / child-prop fallbacks silently missed** | **Fixed.** New `case 'component'` in `collectDomBindings` mirrors the structure of `case 'element'`: iterates `node.props`, recurses into `node.children`, gate matches the emitter at `collect-elements.ts:442` (`hasPropsRef \|\| deps.length > 0` = reactive, `prop.callsReactiveGetters \|\| prop.hasFunctionCalls` without match = fallback). Label encoded as `"ComponentName.propName"` reuses the existing `type: 'attribute'` union — the compromise I suggested, keeps `DomBinding.type` stable for consumers. Edge cases handled correctly: spread (`name.startsWith('...')`) and JSX-as-prop (`prop.jsxChildren`) are both skipped with comments explaining why. **Three dedicated tests** lock all three facets: fallback via child-prop opaque call, reactive via `props.xxx` (hasPropsRef branch), and recursion into a `<Dialog><h1>{formatTitle(page)}</h1></Dialog>` subtree — the second-order gap from the silent fall-through. |
| **#2 Per-fallback filler lines redundant** | **Fixed cleanly.** New optional `DomBinding.expression?: string` populated at collection time with the source text — `node.expr` / `attr.value` / `node.condition` / `node.array` / `prop.value`. Event handlers explicitly opt out, documented on the field. CLI output switches from 3 lines per fallback (type+id line, two identical explanation lines) to 1 column-aligned line per fallback (type+id + expression), with `padEnd(width)` cleanly handling the alignment concern I raised about the PR-body verification. JSON output uses conditional spread so pre-expression JSON consumers still parse. **New `domBindings carry expression text for non-event sites` test** covers all 5 sites in one `Everything` component + pins event-handler opt-out per the doc. |
| **#3 `expect(...).rejects` not awaited** | **Fixed.** Both error-path tests now `await expect(run(...)).rejects.toThrow('exit 1')`. A future async refactor of the error handler won't silently break the assertion ordering. |
| **#4 PR body verification output mismatched actual format** | **Resolved via concern #2.** The new column-aligned single-line output IS now the format the user sees, so the body's verification snippet matches reality. |
| **#5 `AnalyticsDemo` in docs doesn't exist** | **Fixed.** Swapped for `calendar` in `ai-native.md` — a real component the PR body itself verified has 8 fallbacks. |

### Delta verification

- Scope is exactly what it claims: `+~180 lines` across debug.ts (case + field), why-wrap.ts (alignment + expression), and tests. No unrelated drift.
- `case 'component'`'s comment is unusually detailed — names the silent fall-through, quotes the emitter gate line, explains the label-prefix compromise. Anyone landing on this in 6 months has full context without reading the PR thread.
- The `hasPropsRef = propValue.includes('props.')` approximation matches the emitter's own string check (`collect-elements.ts:440`) rather than trying to out-smart it. That's the right call — debug mirrors emitter, same failure mode if any.
- The column-alignment for `why-wrap` output uses a single-pass max over all cells, so it scales cleanly from 1 fallback (width-of-that-cell) to 8 (calendar) to however many a future component has.
- All 13 new or tightened test cases are behavioural pins on real source shapes — no over-specified string matching that'd turn red on cosmetic drift.

### Verdict

**Ready to merge.** The round-1 blocker is closed structurally (not just by a test), and four nits are all picked up in the same commit cleanly rather than deferred. The #937 wrap-by-default rollout now has its full visibility arc — all five emitter sites surface through `why-wrap`, output is informative per-fallback, and the complete component/subtree recursion handles the case that used to vanish.

After merge:

1. #944 closes with the full scope delivered.
2. #951 (Option 3 same-key destructured loop refresh) — load-bearing only if a real user report surfaces.
3. Optional DRY+: retire `hasPropsRef` / `propValue.includes('props.')` string checks on both sides (emitter + debug) via `IRProp.accessesProps` / `IRAttribute.accessesProps` AST flags. Out of scope; low priority.

With this merged, the #937 arc's headline claim — "silent drops are gone, and when the compiler can't prove reactivity it tells you explicitly" — is fully backed by tooling.

Issue #971, opened 2026-04-22
compiler: unify JSX-embeddable expression dispatcher to eliminate silent-drop bug class


kfly8:
## Background

We keep shipping targeted fixes for a recurring failure mode: a JSX-containing expression shape works correctly in one position (e.g. JSX child `{expr}`) but is silently dropped or mis-compiled in another position (e.g. top-level component return). Recent examples:

- #968 — top-level `return cond ? <A/> : null` drops the conditional; SSR always renders the truthy branch. Fixed in #970 by adding `ConditionalExpression` to `analyzer.jsxReturn` and dispatching it to the existing `transformConditional`.
- #937 → #939 → #940 → #941 → #942 → #943 — "wrap-by-default" fallback rolled out per-position (text interpolation → attributes → conditionals → child-component props → loops) because each position had its own reactivity gate.

The #937 family addresses the **reactivity** side of the same root problem. This issue is about the **shape / compilability** side.

## Root cause

There are (at least) **three parallel allow-lists** of "expressions that may contain JSX", one per position, none of which references the others:

1. **Return position** — `analyzer.jsxReturn` type in `packages/jsx/src/analyzer-context.ts:99-105` + the discriminator in `visitComponentBody` (`analyzer.ts:541-562`). Historically limited to `JsxElement | JsxFragment | JsxSelfClosingElement`; #968 widened it to also accept `ConditionalExpression`.
2. **JSX child `{expr}` position** — `transformExpression` in `jsx-to-ir.ts:784-904` dispatches to `transformConditional`, `transformLogicalAnd`, `transformNullishCoalescing`, `transformJsxFunctionCall` (inline JSX helpers, #569), `transformMapCall` (#783), plus fallback to `IRExpression`.
3. **Conditional branch position** — `transformConditionalBranch` in `jsx-to-ir.ts:1114-1180` has its own near-identical list plus parenthesis unwrapping and a `null`/scalar fallback.

Plus a 4th implicit path: the recursive `visitComponentBody` walk, which sets `jsxReturn` to the first JSX descendant it encounters. This is the mechanism that caused #968's specific symptom — the walk descended into the `ConditionalExpression` and registered the `whenTrue` branch as the return, silently replacing the conditional with one of its branches.

Each new expression shape has to be added to N lists. Forgetting one list manifests as a silent mis-render rather than a compile error — the bugs are expensive to find and every new TypeScript syntax (e.g. `SatisfiesExpression`) is a potential future instance.

[...]

PR #981, opened 2026-04-22
fix(toast): add setTimeout fallback so exit state finalizes without transitionend

### LINKED ISSUE #979: flaky: toast.spec.ts — transitionend not firing within 5s in CI
## Summary

\`site/ui/e2e/toast.spec.ts:21\` (\"closes toast when close button is clicked\") flakes in CI. First observed on attempt 1 of run [24773987180, job 72487464613](https://github.com/barefootjs/barefootjs/actions/runs/24773987180/job/72487464613) (push of #977 merge). Retry attempt 2 passed — overall run is green, so the whole CI-main remains green.

## Failure

\`\`\`
1) [chromium] › e2e/toast.spec.ts:21:9 › Toast Reference Page › Default Toast › closes toast when close button is clicked

Error: expect(locator).toHaveAttribute(expected) failed
Locator:  locator('[data-slot=\"toast\"]').first()
Expected: \"hidden\"
Received: \"exiting\"
Timeout:  5000ms

Call log:
  - Expect \"toHaveAttribute\" with timeout 5000ms
  - waiting for locator('[data-slot=\"toast\"]').first()
    9 × locator resolved to <div ... data-state=\"exiting\" ...>
      - unexpected value \"exiting\"
\`\`\`

## Root cause

\`ui/components/ui/toast/index.tsx\` dismisses in two steps:

1. \`createEffect\` reacts to \`props.open === false\`: sets \`data-state=\"exiting\"\` and swaps class to \`stateClasses.exiting\` (\`translate-x-full opacity-0\`).
2. \`transitionend\` listener on the element detects \`data-state === 'exiting'\` and swaps to \`data-state=\"hidden\"\` + \`hidden\` class.

The test polled the element 9 times over 5s and always saw \`data-state=\"exiting\"\` — step 2 never fired. Headless Chromium under CI load can occasionally drop \`transitionend\` for short (~500ms \`duration-slow\`) transitions, especially when the class change is batched with other work in the same frame.

## Not a regression from #971

Toast source last meaningfully changed in \`ba24ef9a Fix border-color layer priority\`, well before the #971 series. The #971 compiler refactors (#973/#974/#976/#977/#978) don't touch runtime effect dispatch or CSS, and the full \`bun test\` scope stayed 1642 pass / 0 fail throughout. Flaky on main CI, not a new regression.

## Proposed fix

Defensive timeout alongside \`transitionend\` so \`transitionend\` no longer gates correctness:

\`\`\`ts
// ui/components/ui/toast/index.tsx, inside handleMount, in the exit branch:
el.dataset.state = 'exiting'
el.className = \`\${stateClasses.exiting} \${toastBaseClasses} \${toastVariantClass} \${className}\`
setTimeout(() => {
  if (el.dataset.state === 'exiting') {
    el.dataset.state = 'hidden'
    el.className = \`hidden \${toastBaseClasses} \${toastVariantClass} \${className}\`
  }
}, 1000) // duration-slow (~500ms) + buffer
\`\`\`

Idempotent with the existing \`transitionend\` listener: whichever fires first wins, the loser's \`if (el.dataset.state === 'exiting')\` guard rejects the no-op.

## Alternatives considered

- Drop \`duration-slow\` from \`toastBaseClasses\` while exiting → visible jump, poor UX.
- Replace \`transitionend\` with a \`requestAnimationFrame\` loop that polls \`getComputedStyle\` — more code, same race.
- Bump the Playwright timeout to 10s — hides the flake, doesn't fix it.

The \`setTimeout\` fallback is the same pattern Radix UI and shadcn/ui use for similar animation-state transitions.

## Scope

- File: \`ui/components/ui/toast/index.tsx\`, \`handleMount\` inner effect (around lines 256–263 at head-of-main).
- Tests: existing E2E should pass unchanged; should reduce CI flake rate to zero.

kfly8:
## Summary

Fixes #979 — `toast.spec.ts:21` ("closes toast when close button is clicked") was flaking in CI because `data-state="exiting" → "hidden"` was gated solely on a `transitionend` event. Under CI load, headless Chromium can drop `transitionend` for short transitions (`duration-slow` = 300ms), leaving the component stuck at `exiting` and causing the 5s Playwright assertion to time out.

**Root cause**: `handleMount` in `ui/components/ui/toast/index.tsx` relied on `transitionend` as the only path to `hidden`. If the event does not fire, the toast stays at `exiting` indefinitely.

**Fix**: add a 1-second `exitTimer` fallback alongside the existing `transitionend` listener. Whichever fires first wins; the loser's `if (el.dataset.state === 'exiting')` guard is already in place and rejects the duplicate write. This is the same belt-and-braces idiom used by Radix UI and shadcn/ui for the same class of problem.

Changes:
- `ui/components/ui/toast/index.tsx` — add `exitTimer` handle, arm it in the exit branch, and clear it in three places (on re-open, inside `transitionend`, and at the start of the else branch) to prevent stale timer snaps.

This is **not a regression from #971** — toast source last changed in `ba24ef9a`, well before the #971 compiler series.

## Test plan

- [x] `bun test ui/components/ui/toast` — 18/18 pass (IR-level tests unchanged)
- [x] `bun run build` — clean build passes
- [ ] CI: `site/ui/e2e/toast.spec.ts` should now pass without flaking (verify on next 2 runs)

### DIFF
--- a/ui/components/ui/toast/index.tsx
+++ b/ui/components/ui/toast/index.tsx
@@ -217,7 +217,8 @@ function Toast(props: ToastProps) {
     let exitTimer: ReturnType<typeof setTimeout> | null = null
 
     // Transition to hidden after exit animation completes
-    el.addEventListener('transitionend', () => {
+    el.addEventListener('transitionend', (e) => {
+      if (e.target !== el) return
       if (el.dataset.state === 'exiting') {
         if (exitTimer) { clearTimeout(exitTimer); exitTimer = null }
         el.dataset.state = 'hidden'
--- a/ui/components/ui/toast/index.tsx
+++ b/ui/components/ui/toast/index.tsx
@@ -214,10 +214,12 @@ function Toast(props: ToastProps) {
   const handleMount = (el: HTMLElement) => {
     const providerEl = el.closest('[data-slot="toast-provider"]') as HTMLElement
     let dismissTimer: ReturnType<typeof setTimeout> | null = null
+    let exitTimer: ReturnType<typeof setTimeout> | null = null
 
     // Transition to hidden after exit animation completes
     el.addEventListener('transitionend', () => {
       if (el.dataset.state === 'exiting') {
+        if (exitTimer) { clearTimeout(exitTimer); exitTimer = null }
         el.dataset.state = 'hidden'
         el.className = `hidden ${toastBaseClasses} ${toastVariantClass} ${className}`
       }
@@ -232,6 +234,7 @@ function Toast(props: ToastProps) {
       if (isOpen) {
         // Clear dismiss timer when entering
         if (dismissTimer) { clearTimeout(dismissTimer); dismissTimer = null }
+        if (exitTimer)    { clearTimeout(exitTimer);    exitTimer    = null }
 
         // Entering state
         el.dataset.state = 'entering'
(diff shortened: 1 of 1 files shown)


kfly8:
Thanks for the review. Addressed the `e.target` bubbling concern in bd89482b — added `if (e.target !== el) return` so events from transitioning descendants are ignored.

Re named-function cleanup: Toast elements stay in the DOM for their lifetime (they toggle between `hidden` / `visible` via `data-state` rather than being inserted/removed), so the listener is intentionally long-lived. Leaving cleanup out of scope for this PR.

PR #998, opened 2026-04-23
fix(xyflow): draw lines at tile edge (x=0, y=0) not tile center

kfly8:
## Problem

Lines placed at `scaledGap / 2` pass through the **center** of each cell. The conventional grid behavior is for lines to sit on **cell boundaries**, not centers.

## Fix

Place lines at the tile edge (`x=0`, `y=0`):

```ts
// lines: vertical line at left edge of tile
M0 0 V${scaledGap}

// cross: horizontal + vertical lines at top-left corner of tile
M0 0 V${scaledGap} M0 0 H${scaledGap}
```

When the pattern tiles, lines appear at `0, gap, 2*gap, ...` — forming cell boundaries rather than cell centers.

### DIFF
--- a/packages/xyflow/src/background.ts
+++ b/packages/xyflow/src/background.ts
@@ -81,11 +81,10 @@ export function initBackground(scope: Element, props: Record<string, unknown>): 
       circle.setAttribute('cy', String(scaledGap / 2))
     } else if (variant === 'lines') {
       const path = patternContent as SVGPathElement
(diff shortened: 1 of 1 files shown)


kfly8:
This is not a barefootjs bug — both @xyflow/react and barefootjs intentionally draw lines at tile center. The edge-based requirement is desk-specific and should be handled at the application level.

Issue #1018, opened 2026-04-24
refactor(compiler): replace multi-pass scope classification in generate-init.ts with a single fixpoint


kfly8:
## Problem

`generate-init.ts` decides whether each local function goes to **module scope** (accessible from the SSR template) or **init scope** (inside `initComponentName()`). This decision currently uses two separate passes that have grown independently and still diverge for new patterns:

**Pass 1** (`generate-init.ts` L221–242)
Builds `componentScopeNames` = `{signal getters/setters, memo names, constant names, props}`, then calls `bodyReferencesComponentScope(fn.body, componentScopeNames)` per function. If the function directly references any of those names → init scope, else → module-level candidate.

**Pass 2** (`generate-init.ts` L245–282, added in #1017)
Collects `initScopeFnNames` from functions already demoted to init scope in pass 1, then runs a fixpoint loop demoting module-level candidates that call init-scope functions. This was needed because pass 1 doesn't check *transitive* function → function chains.

The two passes cover overlapping but subtly different name sets and must be kept in sync manually. Any new pattern that creates a three-hop chain (e.g. `fetchPage → articlesForPage → ARTICLES`) requires yet another tweak.

## Root cause

Pass 2 needed to exist because `componentScopeNames` in pass 1 includes constant names but **not function names**. So `fetchPage → articlesForPage` was invisible to pass 1 even though `articlesForPage` ended up in init scope (because it referenced `ARTICLES`).

Both passes are really doing the same thing: *reachability in a dependency graph from a set of "must-be-in-init-scope" roots*. They should be one.

## Proposed fix

Replace both passes with a single fixpoint over all local declarations:

```
initRequiredNames =
    signal getters/setters
  ∪ memo names
  ∪ props param names / propsObjectName
  ∪ { name | constant is in init scope (neededConstants) }

// Classify module-level candidates via fixpoint
pendingModuleLevel = [fn for fn in localFunctions if fn.isModule]

changed = true
while changed:
  changed = false
  stillModule = []
  for fn in pendingModuleLevel:
    refs = extractIdentifiers(fn.body) - fn.params
    if refs ∩ initRequiredNames ≠ ∅:
      // demote to init scope
      declarations.push({ kind: 'function', info: fn, ... })
      initRequiredNames.add(fn.name)   // ← so callers get demoted next round
      changed = true
    else:
      stillModule.push(fn)
  pendingModuleLevel = stillModule

moduleLevelFunctions = pendingModuleLevel
```

[...]

PR #1019, opened 2026-04-24
refactor(compiler): unify two-pass function scope classification into single fixpoint (#1018)

kfly8:
## Summary

Replaces the two-pass function scope classification in `generateInitFunction` with a single fixpoint, resolving [#1018](https://github.com/barefootjs/barefootjs/issues/1018).

- Seed `initRequiredNames` with reactive roots (signals, memos, props) and constants routed to init scope (`neededConstants`) — module-scope constants (`systemConstructKind`, `initStmtAssigned`) are deliberately excluded so functions referencing only those can still live at module level.
- Walk every module-level candidate: if its body references any name in the set, demote it to init scope and add its own name so transitive callers are demoted in the next iteration. Loop until stable.
- Subtract `fn.params` from the reference set to prevent a shadowing parameter from triggering a false demotion (a latent bug in both previous passes).
- Remove the `bodyReferencesComponentScope` import from `generate-init.ts`. The helper stays in `utils.ts` because `emit-registration.ts` still uses it.

Three-hop chains (e.g. `fetchPage → articlesForPage → ARTICLES`) are now handled by the same mechanism as direct references — no separate pass needed.

## Verification

- `bun run test --filter=packages/jsx` — **1809/1809 pass** (zero regressions)
- `bunx tsc --noEmit` in `packages/jsx` — clean
- Compiled client JS for `infinite-scroll-demo` diffed against pre-refactor: **byte-identical**
- `cd site/ui && bunx playwright test infinite-scroll` — passes when the demo's built-in 12% random network-error rate does not fire on multi-page load (pre-existing flake, unrelated to this refactor)

## Test plan

- [ ] CI: `bun run test --filter=packages/jsx` green
- [ ] CI: typecheck green
- [ ] Run `bunx playwright test infinite-scroll` locally a couple of times — expect occasional flake from the demo's random error injection, not a consistent failure

### DIFF
--- a/packages/jsx/src/ir-to-client-js/generate-init.ts
+++ b/packages/jsx/src/ir-to-client-js/generate-init.ts
@@ -4,7 +4,7 @@
 
 import type { ComponentIR, ConstantInfo, FunctionInfo, IRNode } from '../types'
 import type { ClientJsContext, ConditionalElement } from './types'
-import { varSlotId, bodyReferencesComponentScope, PROPS_PARAM } from './utils'
+import { varSlotId, PROPS_PARAM } from './utils'
 import { collectUsedIdentifiers, collectUsedFunctions, collectIdentifiersFromIRTree, extractIdentifiers } from './identifiers'
 import { valueReferencesReactiveData, getControlledPropName, detectPropsWithPropertyAccess } from './prop-handling'
 import { IMPORT_PLACEHOLDER, MODULE_CONSTANTS_PLACEHOLDER, RUNTIME_MODULE, detectUsedImports, collectUserDomImports, collectExternalImports } from './imports'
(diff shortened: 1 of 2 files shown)


kfly8:
## Update: included the E2E flake fix to unblock CI

CI was failing on `e2e-site-ui (2)` because of a pre-existing flake in the `infinite-scroll` tests, not because of this refactor.

### Evidence that the refactor does not change behavior

1. Checked out `main`'s `packages/jsx/src/ir-to-client-js/generate-init.ts` into the working tree and ran `bun run build` — saved the output.
2. Restored this branch's `generate-init.ts` and ran `bun run build` again.
3. `diff -rq /tmp/main-dist site/ui/dist/components` reports **zero differences across all 326 generated files**. The compiled client JS is byte-identical to what `main` produces.

### Flake source and local reproduction on `main`

`site/ui/components/infinite-scroll-demo.tsx:105` injects a 12% random error rate in `fetchPage`:

```ts
if (Math.random() < 0.12) {
  reject(new Error('Network error — please retry'))
  return
}
```

Running `bunx playwright test infinite-scroll` five times against `main`'s build:

| run | result |
|-----|--------|
| 1 | 20 passed / 1 failed (`scrolling to bottom loads additional articles`) |
| 2 | 21 passed |
| 3 | 20 passed / 1 failed (`end-of-list message appears after all 40 articles are loaded`) |
| 4 | 19 passed / 2 failed (`page indicator advances after loading more`, `end-of-list ...`) |
| 5 | 20 passed / 1 failed (`end-of-list message appears after all 40 articles are loaded`) |

Four of five runs against `main` fail; a different test fails each time. This matches the combined failure rate math: ~55% per run, driven by the 12% rate and the 4-scroll end-of-list assertion (`1 − 0.88⁴ ≈ 40%`).

### Why `main` CI "looks" green

`#1017` (the PR that added this demo + test) passed CI once before merge, but the subsequent merge CI on `main` (`CI — Core`, run `24869663342`) failed on `e2e-site-ui (2)` with the same flake. No PR has been pushed to `main` since, so the most recent `main` data point is already a failure.

### Fix

I included the flake fix commit in this PR (cherry-picked from #1020) so CI can go green here directly. The fix stubs `Math.random` via `page.addInitScript` in `beforeEach` — local 5/5 runs all pass. The existing "Error state" test still works because it captures `Math.random` (now the stub), overrides to force an error, and restores to the stub (not the real random function).

If #1020 is preferred as a standalone fix, the duplicated commit will drop cleanly on rebase.

PR #1044, opened 2026-04-25
fix(compiler): make data-key on loop items unconditional (O-7)

kfly8:
## Summary

The hydrate / insert template emitted \`data-key\` through the generic dynamic-attribute path, which guards every value with \`!= null\`:

\`\`\`
<li \${(item().id) != null ? 'data-key="' + (item().id) + '"' : ''}>
\`\`\`

\`data-key\` is the reconciliation contract for \`mapArray\` — if it's absent on any item, hydration silently mis-aligns SSR DOM with the client signal. The "skip when null" branch was a defensive holdover from a time when \`key\` was optional. With keyed loops now requiring a key (the only non-keyed mapArray path uses \`keyFn=null\` and explicitly does not emit \`data-key\` at all), the null guard hides bugs: a stray \`key={undefined}\` would silently drop the attribute and desync the loop, instead of surfacing as \`data-key="undefined"\` that mapArray's reconciler can fail loudly on.

## Fix

\`templateAttrExpr\` now special-cases \`data-key\` and \`data-key-N\`, emitting an unconditional template-literal interpolation:

| | output |
|---|---|
| before | \`\${(item().id) != null ? 'data-key="' + (item().id) + '"' : ''}\` |
| after  | \`data-key="\${item().id}"\` |

Smaller, no branching, and \`key={undefined}\` now propagates verbatim to the DOM.

## Falling tests

One test pinned to the legacy string-concat shape:

- \`compiler-runtime-contract.test.ts\` → "loop inside conditional: outer=data-key, inner=data-key-1 in SSR template"

Updated to match the new template-literal form. Comment notes the post-O-7 shape so the reason is recorded.

## Test plan

- [x] \`bun test packages/jsx/src/__tests__/\` — 766 / 766 pass
- [x] \`bun test packages/adapter-tests/\` — 214 / 214 pass
- [x] \`bun test packages/client/\` — 212 / 212 pass
- [x] \`bun run --filter '@barefootjs/jsx' build\` — clean exit

### DIFF
--- a/packages/jsx/src/__tests__/compiler-runtime-contract.test.ts
+++ b/packages/jsx/src/__tests__/compiler-runtime-contract.test.ts
@@ -152,10 +152,13 @@ describe('Compiler-Runtime Contract', () => {
       // data-key for outer items and data-key-1 for inner items, matching
       // what the event dispatcher searches for.
       expect(js).not.toContain('data-key-2')
-      // Outer loop items must have data-key in the SSR branch template
-      expect(js).toMatch(/data-key="['"] \+ \(group\.id\)/)
-      // Inner loop items must have data-key-1 in the SSR branch template
-      expect(js).toMatch(/data-key-1="['"] \+ \(item\.id\)/)
+      // Outer loop items must have data-key in the SSR branch template.
+      // Post-O-7 the emission is an unguarded template-literal interpolation
+      // (`data-key="${group.id}"`) instead of a `... != null ? ... : ''`
+      // ternary string concat.
+      expect(js).toMatch(/data-key="\$\{group\.id\}"/)
+      // Inner loop items must have data-key-1 in the SSR branch template.
+      expect(js).toMatch(/data-key-1="\$\{item\.id\}"/)
     })
 
     test('loop inside conditional wires reactive conditionals in mapArray callback', () => {
--- a/packages/jsx/src/ir-to-client-js/html-template.ts
+++ b/packages/jsx/src/ir-to-client-js/html-template.ts
@@ -53,6 +53,13 @@ function templateAttrExpr(attrName: string, valExpr: string, attr: { presenceOrU
   if (attrName === 'style') {
     return `\${((v) => v != null ? 'style="' + v + '"' : '')(styleToCss(${valExpr}))}`
   }
+  // `data-key` / `data-key-N` is a reconciliation contract — every loop item
+  // must carry one. Emit unconditionally; if the user passes `key={undefined}`
+  // we want it to surface as `data-key="undefined"` (and ultimately a runtime
+  // assertion in mapArray) rather than silently fall back to "no key".
+  if (attrName === 'data-key' || attrName.startsWith('data-key-')) {
+    return `${attrName}="\${${valExpr}}"`
+  }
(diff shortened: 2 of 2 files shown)


kfly8:
## Self-review: the "fail loudly" claim in the PR description is incorrect

The PR description said:

> a stray `key={undefined}` would silently drop the attribute and desync the loop, instead of surfacing as `data-key="undefined"` that mapArray's reconciler can fail loudly on.

The second half — *"mapArray's reconciler can fail loudly"* — is wrong. Verified with `tmp/emit-survey/key-undefined-runtime.test.ts`:

```
[KEY=undefined]
  threw: false
  console output: []
  ul.children.length: 1                                  ← inserted 2 items, only 1 rendered
  ul.innerHTML: <li data-key="undefined">a</li>
```

### What actually happens

`packages/client/src/runtime/map-array.ts` L160-181 — the key diff is backed by `Set<string>` + `Map<string, ItemScope>`. When two items share the same key (`'undefined'`):

1. `newKeys.add('undefined')` is called twice → the `Set` collapses them into one
2. On the second item, `scopes.get('undefined')` returns the first item's scope, and `existing.setItem(item)` overwrites the value
3. No throw, no `console.warn`, no `console.error`

So the failure mode is "silent merge of duplicate keys, with array add/update silently broken". Not loud failure.

### No compile-time detection either

`packages/jsx/src/errors.ts` declares `BF023 MISSING_KEY_IN_LIST` and `BF024 MISSING_KEY_IN_NESTED_LIST`, but `grep -r MISSING_KEY_IN_LIST packages/jsx/src --include="*.ts"` returns only the two declaration lines. Nothing actually raises them. `key={undefined}` (statically detectable) and `key={item.maybeId}` (dynamic) both compile silently.

### What this PR actually achieves

| | before | after |
|---|---|---|
| `data-key` template emission | ternary-guarded; skipped when value is null | always `data-key="<value>"` |
| Generated JS size | slightly verbose | shorter |
| DOM for `key={undefined}` | `<li>` (no attribute) | `<li data-key="undefined">` |
| Runtime behavior | silent merge | silent merge (no change) |
| Compile-time warning | none | none (no change) |

So the PR's real value is **cleaner output and a guaranteed `data-key`**, not the "loud failure" the description implied. Follow-up plan in the next comment.

kfly8:
## Re: revising the follow-up plan

The previous comment proposed two follow-ups (a) compile-time `BF023` raise and (b) runtime duplicate-key warn. Re-reading them side by side:

| Failure mode | (a) static `BF023` raise | (b) runtime duplicate warn |
|---|---|---|
| `key` prop completely missing | ✅ | (never fires — no key at all) |
| literal `key={undefined}` | ✅ statically reachable via AST | ✅ if duplicates result |
| `key={item.id}` where the type is `string \| undefined` | ✅ catchable at the type-check boundary | ✅ if runtime undefined → duplicates |
| user data has duplicate ids | ❌ undecidable statically | ✅ |

If (a) is widened to "raise `BF023` when the `key` prop is missing **or** the value type is not non-nullable", what remains for (b) is **only the data-error case** (user passed real data with duplicate ids). Adding a per-reconcile `Set` duplicate check for that single case is hard to justify against the always-on cost — both React and Solid leave data-side id duplication to the user (silent or DevTools-only).

→ **(a) alone is enough.** Skip (b).

PR #1049, opened 2026-04-25
feat(runtime): upsertChild — collapse SSR/CSR mode in composite emit

kfly8:
## Summary

Adds \`upsertChild(parent, name, slotId, props, key?)\` to the client runtime — a single helper that resolves SSR (find existing scope + \`initChild\`) vs CSR (find placeholder + \`createComponent\` + \`replaceWith\`) at runtime. Returns the resolved element (or null when neither shape matches — already initialised on a previous reconcile pass).

This lets the compiler drop the \`mode: 'csr' | 'ssr'\` argument that \`emitComponentAndEventSetup\` (and transitively \`emitInnerLoopSetup\`) had to thread through every composite-loop call site. The hand-rolled \`if (__existing) { initChild(...) } else { createComponent + replaceWith }\` block at every call site collapses to a single \`upsertChild(...)\` call.

## Stacked on #1048 (which is stacked on #1047)

Review and merge #1047 → #1048 first; then GitHub will rebase this PR's base onto main.

## Cascade in the emit layer

| Site | Before | After |
|---|---|---|
| \`emitComponentAndEventSetup\` | takes \`mode\` arg, emits 2 different blocks per comp | no \`mode\`, emits 1 \`upsertChild(...)\` per comp |
| \`emitInnerLoopSetup\` | takes \`mode\` arg, threads it through recursion | no \`mode\` |
| \`emitBranchInnerLoops\` | wraps comp setup in \`if (!__existing) { csr } else { ssr }\` | one call, no split |
| \`stringify/composite-loop.ts\` | \`let __el; if (__existing) { __el = … } else { __el = … clone …}\` | \`const __el = __existing ?? (() => { … clone … })()\` (matches \`stringifyPlainLoop\`'s shape) |

## Concrete diff in generated code

For \`loop-nested-plain\` (composite, no nested comps): the composite renderItem now matches the plain-loop renderItem shape — one-liner \`__el\` initialisation, then inner-loop setup, then return.

For composites with nested comps (e.g. the \`composite-branch-loop\` test fixtures), each nested component goes from a 4-line SSR block + a 4-line CSR block paired inside an \`if (!__existing)\` to a single \`upsertChild(...)\` line — roughly halves the per-component emission.

## Tests touched

Two regression tests (\`composite-branch-loop.test.ts\`) asserted on the literal \`createComponent('Badge'\` / \`createComponent('Button'\` strings the old emitter produced. Updated to assert on \`upsertChild(...)\` + the component name in the new emission, with a comment explaining why.

## Test plan

- [x] \`bun test packages/jsx/src/__tests__/\` — 766 / 766 pass
- [x] \`bun test packages/adapter-tests/\` — 214 / 214 pass
- [x] \`bun test packages/client/\` — 212 / 212 pass
- [x] \`bun run --filter '@barefootjs/jsx' build\` — clean exit
- [x] \`bun run --filter '@barefootjs/client' build\` — clean exit

### DIFF
--- a/packages/client/src/runtime/index.ts
+++ b/packages/client/src/runtime/index.ts
@@ -77,7 +77,7 @@ export { styleToCss } from './style'
 // Runtime helpers
 export { findScope, find, $, $c, $t, qsa } from './query'
 export { hydrate, rehydrateAll } from './hydrate'
-export { registerComponent, getComponentInit, initChild } from './registry'
+export { registerComponent, getComponentInit, initChild, upsertChild } from './registry'
 export { insert, type BranchConfig } from './insert'
 export { updateClientMarker } from './client-marker'
 
--- a/packages/client/src/runtime/registry.ts
+++ b/packages/client/src/runtime/registry.ts
@@ -8,6 +8,7 @@
 import { BF_SCOPE, BF_CHILD_PREFIX } from '@barefootjs/shared'
 import { hydratedScopes } from './hydration-state'
 import { setCurrentScope } from './context'
+import { createComponent } from './component'
 import type { InitFn } from './types'
 
 /**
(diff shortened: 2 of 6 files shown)


kfly8:
## CI fix: \`upsertChild\` was missing from \`RUNTIME_IMPORT_CANDIDATES\`

The Core CI shard 4 caught:

\`\`\`
Uncaught ReferenceError: upsertChild is not defined
  at feed-demo-81dce552.js:441
\`\`\`

\`upsertChild\` was added to the runtime exports (\`packages/client/src/runtime/registry.ts\`) and the compiler emitted \`upsertChild(...)\` calls, but the compiler's import-tracking allow-list (\`ir-to-client-js/imports.ts::RUNTIME_IMPORT_CANDIDATES\`) was never updated to include the new symbol — so the generated import statement never picked it up.

The unit-test layer never caught this because the existing tests only check the emitted string for substrings, not that the resulting JS actually executes.

Pushed \`cdfc4808\` adding \`'upsertChild'\` to \`RUNTIME_IMPORT_CANDIDATES\`. Re-running locally: all 15 e2e/social-gallery + e2e/controlled-input-focus tests pass.

Issue #1080, opened 2026-04-27
Migrate @barefootjs/chart to JSX-native client components (Phase 9 follow-up)


kfly8:
## Background

The chart package is currently fully imperative: every component (`Bar`, `Line`, `Area`, `Pie`, axes, grids, tooltips) uses `document.createElementNS(SVG_NS, ...)` + `setAttribute(...)` inside `createEffect` to build SVG. This was originally necessary because the JSX compiler couldn't preserve SVG namespace inside `mapArray` loops or bind reactive SVG attributes.

That blocker is gone:

- **#1070** — Phase 9 Block #135 (Graph/DAG Editor) added SVG namespace preservation inside `mapArray` and reactive bindings on `cx`/`cy`/`d`/`viewBox`/`x`/`y`.
- **#1079** — fixed the IR-to-template emitter dropping `export { ... }` named-export blocks from `\"use client\"` files. This unblocks barrel re-exports like `export { Bar } from './bar'` from `chart/index.tsx`.
- **#1077** — narrowed SVG element `ref` callback types with `SVGBaseAttributes`, added per-tag `(element: SVG*Element) => void` overrides for all 22 SVG entries, and broadened `<stop>` to accept camelCase `stopColor` / `stopOpacity`. Type-only regression test wired into `bun run --cwd packages/jsx test`.

The chart package can now be expressed in JSX. Doing so removes hand-rolled DOM diff bookkeeping and aligns with the Solid-style fine-grained reactivity model the rest of the framework uses.

## Earlier PoC and what we learned

PR #1077's first iteration migrated `Bar` to a JSX-native client component. That iteration was reverted because of a structural problem, not a compiler issue:

- The wrapping pattern `<svg style=\"display:contents\"><g ref={mountInto(ctx.svgGroup())}>...</g></svg>` to bridge the imperative parent BarChart used `display:contents` on the outer `<svg>` so it didn't render. Playwright's `toBeVisible()` checks against e2e selectors like `container.locator('svg')` failed because the wrapper `<svg>` was the first match and `display:contents` makes it invisible per Playwright semantics.
- After hydration the inner `<g>` was reparented into BarChart's owned SVG, but the SSR HTML still showed children inside the wrapper, leading to mismatches with e2e DOM expectations.

[...]

PR #1086, opened 2026-04-28
chart: migrate trivial primitives to JSX-native (Step 2 of #1080)

kfly8:
## Summary

Step 2 of the [#1080](https://github.com/piconic-ai/barefootjs/issues/1080) chart migration. Builds on the JSX-native containers from #1082 and the SSR client-shim from #1085.

- Convert the trivial primitive set — `CartesianGrid`, `AreaCartesianGrid`, `XAxis`, `AreaXAxis`, `YAxis`, `AreaYAxis`, `PolarGrid`, `PolarAngleAxis`, `RadialBar` — to JSX-native components. Each one now returns a single `<g>` populated by reactive memos, replacing the imperative `init*` callbacks.
- Drop the imperative primitive files from `@barefootjs/chart` (pre-release, no shim).
- Add `buildRadialBarArcs` so `d3-shape` stays inside `@barefootjs/chart` and never hits the consumer bundle (the browser import map only resolves `@barefootjs/*`).
- Add `CHART_CLASS_*` constants. Importing the class names sidesteps the `cssLayerPrefixer`, keeping `class="chart-x-axis"` etc. un-prefixed so the existing e2e selectors (`.chart-x-axis text`, `.chart-grid line`, ...) keep working.
- Bundle size: `barefoot-chart.js` drops from 119 KB → 101 KB.

## Notes

- **Sibling-map loop-marker collision.** The runtime's `findLoopMarkers` (`packages/client/src/runtime/map-array.ts`) picks the *last* `<!--bf-loop-->` / `<!--bf-/loop-->` comment pair inside a container, so two adjacent `.map()` calls in the same parent silently overwrite each other. `CartesianGrid` / `AreaCartesianGrid` flatten horizontal+vertical lines into one keyed list, and `PolarGrid` flattens concentric shapes + radial spokes the same way, to avoid the collision. Worth a separate runtime fix; not a blocker for step 2.
- **Pre-existing radar Playwright flakes.** `radar-chart.spec.ts:65,83` (\`changing gridType switches to circle grid\` and \`toggling showGrid hides grid lines\`) already fail on `main` — Playwright's `toBeVisible()` returns false for the `<g class=\"chart-polar-grid\">` even when the element has 13 visible children, because Chrome reports `getBoundingClientRect = 0×0` for SVG `<g>` wrappers whose layout is set entirely by descendants. Not introduced by this PR.

## Test plan

- [x] `bun test ui/components/ui/chart/index.test.tsx` — 31 IR tests pass (10 new ones cover the new primitives)
- [x] `bunx playwright test bar-chart.spec.ts area-chart.spec.ts line-chart.spec.ts pie-chart.spec.ts radar-chart.spec.ts radial-chart.spec.ts` — 69 / 71 chart e2e pass; the 2 failing radar tests fail on `main` too (see *Notes*)
- [x] `bun run build` for `packages/chart` and `site/ui` — both clean


kfly8:
Rebased onto fresh `main` after #1091 (sibling-map markers, #1087) and #1095 (SVG ternary wrap, #1088) merged.

Re-running the chart e2e:

- Before: 69 passed / 2 failed (`radar-chart.spec.ts:65,83`)
- After rebase: **70 passed / 1 failed**
- IR tests: 31 / 31 still green
- Bundle: `barefoot-chart.js` unchanged at 101 KB

The remaining failure (`radar-chart.spec.ts:65 — changing gridType switches to circle grid`) is a *third* compiler edge case I just filed as #1098: when a `.map(...)` body is a ternary, the compiler drops the `key={...}` extractor and mapArray falls back to index-based reconciliation. So when `gridType` flips `polygon` → `circle`, the existing polygons are mutated in place instead of being replaced with circles. `RadialBar` hides this today because both branches use `<path>`, but PolarGrid's three-way ternary makes the regression observable.

Will follow up once #1098 lands.

kfly8:
Rebased onto `main` after #1099 (mapArray key extraction across ternary branches) merged.

Re-running the chart e2e:

- Before #1099: 70 passed / 1 failed (`radar-chart.spec.ts:65 — changing gridType switches to circle grid`)
- After rebase: **71 passed / 0 failed** ✓
- IR tests: 31 / 31 still green
- Bundle: `barefoot-chart.js` unchanged at 101 KB

The previous failure was unblocked by #1099 — the keyFn lifted out of the PolarGrid ternary now lets mapArray reconcile by `c-…` / `p-…` keys, so the polygons unmount and circles mount on `gridType` flip.

Ready for review.

PR #1092, opened 2026-04-28
fix(theme-switcher): icon grows after toggle on ui.barefootjs.dev

kfly8:
## Summary
ThemeSwitcher's inner `SunIcon` / `MoonIcon` JSX components were compiled into globally-registered children. site/ui already registers identically-named icons under `ui/components/ui/icon` (lucide-style, no `width`/`height` when used without a `size` prop). At runtime, `renderChild('SunIcon')` returned the wrong component, and the swapped SVG had no size attributes — sizing to fill the button (~36×36).

Symptoms:
- Toggling the theme on `https://ui.barefootjs.dev/` made the icon grow.
- `barefootjs.dev` (site/core) was unaffected because it doesn't register a colliding `SunIcon`/`MoonIcon`.
- Reloading restored the correct size because the SSR template hard-codes the right icon — only the runtime swap was broken.

Fix: inline the two SVGs directly in `ThemeSwitcher`'s JSX. The compiler now emits raw `<svg>` markup in the `insert()` branch templates instead of going through the global component registry, so the collision can't happen.

Added a regression e2e (`icon stays 20x20 after toggle`) that asserts the SVG bounding box is 20×20 both before and after the click.

## Test plan
- [x] `site/ui` clean build
- [x] `bun test --cwd . __tests__` — 1883 pass / 0 fail
- [x] `site/ui` theme-switcher e2e — 7/7 pass (incl. new regression test)
- [x] Manual: navigate to local site/ui, click ThemeSwitcher, confirm icon stays 20×20

### DIFF
--- a/site/shared/components/theme-switcher.tsx
+++ b/site/shared/components/theme-switcher.tsx
@@ -47,23 +47,6 @@ function writeThemeCookie(theme: Theme): void {
   document.cookie = parts.join('; ')
 }
 
-function SunIcon() {
-  return (
-    <svg width="20" height="20" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
-      <circle cx="12" cy="12" r="4" />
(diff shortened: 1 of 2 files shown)


kfly8:
Closing in favor of a compiler-level fix — inlining the SVG was a workaround. The root cause is that the compiler registers components in a flat global registry by name, so identically-named components across files collide. Will reopen with a scoped registry fix.

PR #1132, opened 2026-04-30
fix(jsx): rewrite bare prop refs in localConstants values too (not just templateValue)

kfly8:
Follow-up to #1127.

## Bug

After #1127, the destructured-from-`(props)`-arg pattern works for the destructured names themselves — `const { org } = props` correctly emits `const org = _p.org`, and JSX `data-org={org}` rewrites to `_p.org` in the template. But OTHER local consts that *reference* those destructured names still keep bare names in their init-body source:

```ts
function DeskCanvas(props: Props) {
  const { org, projectNumber, deskNumber } = props
  const viewportCacheKey = `desk-viewport-${org}-${projectNumber}-${deskNumber}`
  //                                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  // bare names — `templateValue` was being rewritten to `_p.X` but `value` wasn't
}
```

When the minifier collapses init-scope consts into a single comma-separated declaration, `viewportCacheKey` ends up evaluated before the expanded `org` / `projectNumber` / `deskNumber` declarations in the same chain (temporal dead zone) — `ReferenceError: org is not defined` at runtime.

## Fix

When `propsObjectName` is set and a const's value contains bare prop refs, mirror the existing `templateValue` rewrite into `value` itself. The late-stage `props.X → _p.X` rename only catches explicit `props.` prefixes; bare names slip through.

SolidJS-style components (no destructured-from-props locals) are unaffected:
- Their consts use `props.X` directly, which the late-stage rename handles.
- `rewriteBarePropRefs` skips identifiers in property-access positions so `props.X` stays untouched.

## Repro

`piconic-ai/desk`'s `DeskCanvas.tsx` — same source pattern as the PR #1127 repro, just deeper into the function body.

## Test

Wider corpus stays at 37/37 conformance + 848 JSX pass (same 4 pre-existing alias-resolver failures). The compiled output for the destructured-from-props case now reaches `${_p.org}` everywhere, not just in JSX attributes.

### DIFF
--- a/packages/jsx/src/__tests__/destructured-from-props-localconst-value.test.ts
+++ b/packages/jsx/src/__tests__/destructured-from-props-localconst-value.test.ts
@@ -1,0 +1,111 @@
+/**
+ * Pins the rewrite of bare prop references inside `localConstants`
+ * values for `(props)`-arg components that destructure inside the body.
+ *
+ * Without the rewrite, a dependant const like
+ *   `const cacheKey = \`desk-\${org}-\${projectNumber}\``
+ * would keep bare `${org}` / `${projectNumber}` in the emitted init body
+ * and throw `ReferenceError` (TDZ) once the minifier collapses the
+ * declarations into a single comma-separated `const` chain.
+ *
+ * Includes a shadow-guard test: a signal / memo / earlier local that
+ * happens to share a name with a prop must NOT be rewritten — bare refs
+ * to it target the local binding, not `_p.X`.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { compileJSXSync } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
+
+const adapter = new TestAdapter()
+
+describe('destructured-from-props-object → localConstants value rewrite', () => {
+  test('bare prop refs inside a dependant local const are rewritten to `_p.X` in init body', () => {
+    const source = `
+      'use client'
+      import { createSignal } from '@barefootjs/client'
+
+      interface Props {
+        org: string
+        projectNumber: number
+      }
+
+      export function Page(props: Props) {
+        const { org, projectNumber } = props
+        // Reference inside a side-effect closure so the const isn't
+        // statically inlined into the template.
+        const cacheKey = \`desk-\${org}-\${projectNumber}\`
+        const [count, setCount] = createSignal(0)
+        return (
+          <div onClick={() => setCount(count() + cacheKey.length)}>
+            {count()}
+          </div>
+        )
+      }
+    `
+
+    const result = compileJSXSync(source, 'Page.tsx', { adapter })
+    expect(result.errors).toHaveLength(0)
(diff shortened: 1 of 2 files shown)


kfly8:
## Self-review

I'm self-reviewing this PR before requesting human review and found a real concern that needs to be addressed.

### 🔴 Bug: missing shadow guard for signal/local names

The merged predecessor (#1127) added a shadow guard in `jsx-to-ir.rewriteBarePropRefs` so that bare references to signals / memos / locals **shadowing** prop names don't get rewritten to `_p.X`. This PR adds the same kind of rewrite at the analyzer level (for `localConstants` values) but **without the shadow guard**.

#### Repro

```ts
'use client'
interface Props { count?: number }
export function Foo(props: Props) {
  const doubled = count() * 2  // bare `count` references the SIGNAL getter
  return <span>{doubled}</span>
}
```

#### Compiled output with this PR

```js
export function initFoo(__scope, _p = {}) {
  if (!__scope) return
  const doubled = _p.count() * 2   // ❌ rewritten — `count` here was the signal getter, not the prop
  ...
}
```

`_p.count()` calls `.count` (presumably `undefined` from props) as a function — `TypeError: _p.count is not a function`.

The CSR conformance suite passed because `signal-prop-same-name` only references the signal in JSX (`{label()}`), not in any other localConstant value. We need a fixture / unit test that pulls the bare ref into a localConstant initializer.

### 🟡 Other concerns

1. **No new test added.** Should pin the new "rewrite in localConstants value" behavior with a focused test similar to `destructured-from-props-object-template.test.ts`.

2. **Shadow set is incomplete at analysis time.** Even if we add the same shadow guard from #1127 here, `collectConstant` runs DURING the body visit — signals / memos / later locals declared after the current const aren't yet in `ctx.signals` / `ctx.memos` / `ctx.localConstants`. A complete shadow set is only available AFTER the visit completes. The cleanest fix may be to defer the bare-ref rewrite to a post-pass (after `visitComponentBody` finishes).

### Plan

I'll push a follow-up commit that:

1. Adds a shadow guard at the analyzer rewrite site (mirroring #1127's `jsx-to-ir` wrapper logic), OR moves the rewrite to a post-visit pass.
2. Adds a unit test for the signal-shadows-prop case in localConstants values.
3. Re-runs the full conformance + JSX suites.

Holding the merge until those are addressed.

kfly8:
Pushed e82ffd2 addressing the self-review concerns:

- ✅ Shadow guard mirrors #1127's logic: signals / memos / earlier locals that share a name with a prop are excluded from the rewrite.
- ✅ Pure-alias exception ( from body destructure) stays included so dependant consts can reach  — the whole point of this PR.
- ✅ 3 unit tests in `destructured-from-props-localconst-value.test.ts` pin: (1) the rewrite happens, (2) signal-shadow doesn't rewrite, (3) local-with-default-shadow doesn't rewrite.

Suite: 37/37 conformance, 851 JSX pass (same 4 pre-existing alias-resolver failures unrelated).

Issue #1158, opened 2026-05-01
xyflow: Flow component never invokes props.onInit(store) — consumers can't capture the store handle


kfly8:
## Bug

`ui/components/ui/xyflow/index.tsx` declares `onInit?: (store: FlowStore<NodeType, EdgeType>) => void` in `FlowComponentProps` (and `FlowProps`), but the `Flow` function never calls it. Consumers that pass `<Flow onInit={...}>` to capture the store handle (so they can call `store.setNodes(...)` later) get nothing — the callback is silently dropped.

This breaks any downstream pattern where the consumer needs imperative access to the store after creation.

## Repro

```tsx
import { Flow, type FlowStore } from '@ui/components/ui/xyflow'

function App() {
  let store: FlowStore | null = null
  return (
    <Flow
      nodes={[]}
      edges={[]}
      onInit={(s) => { store = s }}   // ❌ never called
    />
  )
}
```

After hydration, `store` stays `null`. Any code that later does `store?.setNodes(...)` is a no-op.

## Live repro in `piconic-ai/desk`

`worker/components/canvas/DeskCanvas.tsx:2505` passes `onInit={onFlowInit}` to `<Flow>`. `onFlowInit(store)` is supposed to populate the closure-captured `flowStore` ref so the node-building `createEffect` can call `store.setNodes(buildNodes(...))`. With `onInit` never invoked, `flowStore` stays null, the effect bails on `if (!flowStore) return`, and no issue cards ever render — even though `fetchItems()` returns 65 items and `setIssues(...)` populates the local signal.

## Suggested fix

In `Flow`, after `const store = createFlowStore<NodeType, EdgeType>(props)`, call the callback:

```ts
const store = createFlowStore<NodeType, EdgeType>(props)
props.onInit?.(store as FlowStore<NodeType, EdgeType>)
```

Should run synchronously during the component body so the consumer's reference is set before the JSX subtree (and any descendant `<Background>` / `<MiniMap>` / `<NodeWrapper>` `useContext(FlowContext)` lookups) renders.

## Tests to add

- A `<Flow onInit={spy}>` with `spy` recording the callback. Assert `spy` is called exactly once with a `FlowStore`-shaped object after first render.
- Variant: imperative test that `onInit` callback's captured store can be used to call `setNodes()` and the next reactive read of `store.nodes()` reflects the change.

## Context

[...]

Issue #1156, opened 2026-05-01
compiler/runtime: useContext called from inlined child JSX template doesn't see provideContext from parent's init


kfly8:
## Bug

When a parent `'use client'` component (e.g. xyflow's `<Flow>`) provides a context value imperatively at init time (`provideContext(FlowContext, store)`) and renders **child JSX components** (e.g. `<Background>`, `<MiniMap>`) that call `useContext(FlowContext)` at the **top of their own component body**, the JSX-native compiler inlines those child component bodies into the parent's TEMPLATE literal. The child's `useContext` ends up running during template evaluation (called via `renderChild`), in a scope chain that does NOT include the parent's `provideContext` (which only runs in init scope).

Result: `Error: useContext: no provider found and no default value` at hydration / `render()` time.

This is structurally similar to the family fixed by #1128 / #1136 / #1137 (init-scope identifier capture), but for **runtime context map propagation across template boundaries**, not lexical bindings.

## Repro

```tsx
'use client'
import { createSignal, createContext, useContext, provideContext } from '@barefootjs/client'

const FlowContext = createContext<{ value: () => number } | undefined>(undefined)

function Child() {
  const store = useContext(FlowContext)            // ← called at top of Child's body
  return <div>{store?.value() ?? 'no store'}</div>
}

export function Flow() {
  const [v] = createSignal(42)
  provideContext(FlowContext, { value: v })
  return <div><Child /></div>
}
```

The compiled `Flow.client.js` template inlines `Child`'s JSX directly:

```js
hydrate('Flow', {
  init: initFlow,                                    // calls provideContext(FlowContext, store)
  template: (_p) => `<div>${renderChild('Child', {})}</div>`,
})
hydrate('Child', {
  init: initChild,                                   // const store = useContext(FlowContext) — runs in TEMPLATE scope
  template: (_p) => `<div>${(() => useContext(FlowContext)?.value() ?? 'no store')()}</div>`,
})
```

When `render(target, 'Flow', {})` runs, `initFlow` provides the context. Then `Flow.template(_p)` renders, calling `renderChild('Child', {})` which invokes `Child.template` in a separate scope. `useContext(FlowContext)` inside Child's template scope returns `undefined` (or throws "no provider found"), even though Flow's init scope has the provider.

## Live repro

[...]

PR #1159, opened 2026-05-01
fix(xyflow): invoke props.onInit(store) so consumers can capture the store handle (#1158)

kfly8:
## Summary

The `Flow` JSX-native component declared `onInit` in its prop type but never called it — consumers passing `<Flow onInit={...}>` silently lost the callback, and any imperative `store.setNodes(...)` outside the JSX subtree was a no-op.

Surfaced during P6 verification in `piconic-ai/desk`: `DeskCanvas` passes `onInit={onFlowInit}` to capture the store, and a `createEffect` later does `store.setNodes(buildNodes(...))`. With the callback dropped, the effect bailed on `if (!flowStore) return` and 65 issue cards never rendered.

Fix: invoke `props.onInit?.(store as FlowStore<NodeType, EdgeType>)` synchronously after `createFlowStore(props)`, so the consumer's reference lands before any descendant `<Background>` / `<MiniMap>` / `<NodeWrapper>` renders. IR test pins the call site.

## Test plan

- [x] `bun test ui/components/ui/xyflow` — 23 / 0 (existing 22 + 1 new pin)
- [x] Verified live in piconic-ai/desk that `onFlowInit` is now invoked (next: cards render once #1156 + this lands)

### DIFF
--- a/ui/components/ui/xyflow/index.test.tsx
+++ b/ui/components/ui/xyflow/index.test.tsx
@@ -230,4 +230,13 @@ describe('Flow', () => {
     expect(result.find({ componentName: 'SimpleEdge' })).not.toBeNull()
     expect(result.find({ componentName: 'NodeWrapper' })).not.toBeNull()
   })
+
+  test('invokes props.onInit(store) so consumers can capture the store handle', () => {
+    // Asserts the Flow body contains a `props.onInit(...)` call site
+    // (gated by `if (props.onInit)` to satisfy the JSX-native compiler's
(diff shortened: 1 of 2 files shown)


kfly8:
Closing — investigation showed that `attachFlowSubsystems` already invokes `props.onInit(store)` internally (packages/xyflow/src/flow-subsystems.ts:191), so the Flow component doesn't need to add a separate call. The original symptom in piconic-ai/desk (cards not rendering) has a different root cause currently being debugged. Will reopen with a more targeted fix if needed.

PR #1168, opened 2026-05-02
feat(xyflow): add disableDefaultNodeStyles flag to opt out of node chrome

kfly8:
## Summary

- Add a `disableDefaultNodeStyles` flag on `FlowProps`. When set, `attachFlowSubsystems` injects a stylesheet that skips the per-node chrome rules (`.bf-flow__node` background / border / padding / grab cursor / centered text and the `--custom` / `--selected` companions) and ships only the layout-critical `box-sizing` + `user-select`.
- Edge, handle, resize, selection rectangle, and group/child rules still ship — those are not "chrome" the consumer is overriding.
- New unit test in `packages/xyflow/src/__tests__/disable-default-node-styles.test.ts`.

## Why

`<Flow renderNode={Fn}>` consumers that mount their own per-node visuals (imperative axis / box / svg renderers, a `data-bf-bridge` JSX bridge, etc.) want to suppress xyflow's default node chrome. The previous escape hatch was the `.bf-flow__node--custom` class. That works at first paint, but `<NodeWrapper>`'s reactive className binding rebuilds the class string from `(selected | group | child)` state on every store update — including position-only updates from drag / resize — and **wipes any consumer-added class**.

Concretely: a one-time `classList.add('bf-flow__node--custom')` from a `ref` callback gets stomped the first time a user starts to resize a node. The chrome reappears mid-interaction, and the canvas flashes white. We hit this in desk's `/dev/catalog/axis` (an imperative axis that calls `store.setNodes` on every resize-drag tick), and the only desk-side workaround was a `:has([data-bf-bridge])` CSS rule injected by the page layout — not great as a default.

`disableDefaultNodeStyles` removes the underlying fight entirely: there's no chrome to suppress, so the reactive className rebind is harmless.

## API

```tsx
<Flow
  nodes={...}
  edges={...}
  disableDefaultNodeStyles
  renderNode={NodeBridge}
/>
```

The flag flows through `FlowProps` → `attachFlowSubsystems` → `injectDefaultStyles(disableNodeStyles)`. Consumers who don't set it see no behavior change.

## Test plan

- [x] `bun test src/` in `packages/xyflow` (33 pass — added 2 new tests for the flag's on/off behavior)
- [x] `bun run typecheck:tests` in `packages/xyflow`
- [ ] Real-world consumer (desk `/dev/catalog/{axis,box,svg,issue-card,drawing-overlay}`) drops the `:has([data-bf-bridge])` CSS workaround in favor of `<Flow disableDefaultNodeStyles>`.

## Notes / future work

This addresses the symptom (default chrome leaking through). The deeper issue — `<NodeWrapper>`'s reactive `className={memo()}` overwrites consumer-added classes — is a more general pattern question (e.g. should reactive class bindings expose a `classList`-style merge primitive instead of full-string rebind?). That's worth a separate discussion / RFC; this PR does not block on it.

### DIFF
--- a/packages/xyflow/src/__tests__/disable-default-node-styles.test.ts
+++ b/packages/xyflow/src/__tests__/disable-default-node-styles.test.ts
@@ -1,0 +1,56 @@
+/**
+ * `<Flow disableDefaultNodeStyles>` lets consumers opt out of the
+ * per-node chrome (white background, dark border, padding, grab
+ * cursor, centered text) injected by `attachFlowSubsystems`.
+ *
+ * The escape exists because `<NodeWrapper>`'s reactive className
+ * binding rebuilds the class string from `(selected | group | child)`
+ * state on every store update, wiping any consumer-added class
+ * (`bf-flow__node--custom` or otherwise). With the flag on, no
+ * `.bf-flow__node` chrome is in the injected stylesheet at all, so
+ * the consumer doesn't need to fight the rebind.
+ */
+import { afterEach, beforeAll, describe, expect, test } from 'bun:test'
+import { GlobalRegistrator } from '@happy-dom/global-registrator'
+
+beforeAll(() => {
+  if (!GlobalRegistrator.isRegistered) GlobalRegistrator.register()
+})
+
+afterEach(() => {
+  document.getElementById('bf-flow-styles')?.remove()
+})
+
+describe('attachFlowSubsystems disableDefaultNodeStyles', () => {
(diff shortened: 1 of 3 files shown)


kfly8:
Closing in favor of a root-cause fix in the renderNode hydration path — see follow-up PR. The flag was a workaround; the real issue is that `renderNode={Fn}` is compiled such that `Fn(node)` runs from outside FlowContext and `node` isn't lifted into the bridge's `bf-p` payload, forcing DOM walk-up gymnastics in consumers.

PR #1170, opened 2026-05-02
fix(client/runtime): set currentScope before init in hydrateComponent

kfly8:
## Summary

- `hydrateComponent` now wraps its `init(scopeEl, props)` call with `setCurrentScope(scopeEl)` (DOM-walker path *and* comment-scope path), mirroring what `createComponent` already does.
- New unit test in `packages/client/__tests__/runtime/hydrate-current-scope.test.ts` covering "parent-provided context resolves from a top-level-hydrated child".
- Full package tests stay green (2069 pass).

## Why

`createComponent` already wraps the registered `init` call with `setCurrentScope(element)` so `useContext` / `provideContext` inside `init` resolve via DOM ancestor walk relative to the new component's own element. `hydrateComponent` — the DOM-walker path that picks up top-level scope elements at hydrate time — didn't.

So when a component's `bf-s="Name_…"` marker landed inside an existing parent scope's DOM, the child's `init` ran with a stale `currentScope` (from whatever component last touched it, or `null`). `useContext` then walked from the wrong anchor and missed every ancestor-provided value.

The concrete shape we hit downstream: a `<Flow renderNode={Fn}>` bridge whose JSX is emitted by Flow's SSR template into the children of `NodeWrapper` carries its own top-level `bf-s` marker (no `~` prefix because the SSR template's `props.renderNode(node)` call doesn't pass `__bfChild`). The DOM walker hydrates it as top-level. Inside the bridge, `useFlow()` (= `useContext(FlowContext)`) returned `undefined` on the initial paint — so the bridge couldn't access the store, and consumers had to walk up to a host-element property (`.bf-flow.__bfFlowStore`, added in #1166) as a workaround. With this PR, that workaround becomes unnecessary on the initial-paint path: the walk in `useContext` starts from the bridge's own element, traverses up to `.bf-flow` via `parentElement`, finds the FlowContext value Flow stamped there, and returns it.

## Implementation

Both `hydrateComponent` paths receive the same `setCurrentScope` wrap — DOM-walker (line 195) and comment-scope (line 130). `setCurrentScope` returns the previous scope for `try`/`finally` restore so a synchronous `init` that throws can't poison the scope for the next component.

## Test plan

- [x] `bun test packages/client/__tests__/runtime/hydrate-current-scope.test.ts` — 1 pass (new)
- [x] `bun test packages/` — 2069 pass, 0 fail
- [ ] Real-world consumer (desk's `/dev/catalog/{axis,box,svg,issue-card}`) renders without the `useFlow()`-walk-up fallback; the bridges resolve `useFlow()` directly.

## Relationship to #1166 / #1169

- #1166's host-element store stamp stays — still the right escape for *imperative* code that needs the store from outside any component init (e.g. desk's DrawingOverlay imperative subsystem mounted from a non-component ref).
- #1169's callable shim stays — covers the runtime reactive path where `props.renderNode(node)` is invoked on an `setNodes` update.
- This PR fixes the third leg: the initial paint where the DOM walker hydrates an orphan scope. With all three together, JSX render-prop consumers can use `useFlow()` directly without DOM walk-up gymnastics.

### DIFF
--- a/packages/client/__tests__/runtime/hydrate-current-scope.test.ts
+++ b/packages/client/__tests__/runtime/hydrate-current-scope.test.ts
@@ -1,0 +1,60 @@
+/**
+ * `hydrateComponent` (the DOM walker that picks up top-level scope
+ * elements at hydrate time) sets `currentScope` to the scope element
+ * before calling `init`. Without it, a component nested inside a
+ * parent scope's DOM but hydrated as a top-level scope (e.g. a
+ * `<Flow renderNode={Fn}>` bridge whose JSX is rendered by Flow's
+ * SSR template into Flow's children but emits its own
+ * `bf-s="ChildName_…"` marker) would call `useContext` against a
+ * stale `currentScope` and miss every provided value upstream.
+ *
+ * The fix mirrors `createComponent`'s `setCurrentScope(element)`
+ * around `init`, so `useContext`'s DOM ancestor walk starts from the
+ * right anchor and finds the parent's `provideContext` value via
+ * `parentElement` traversal.
+ */
+import { beforeAll, describe, expect, test } from 'bun:test'
+import { GlobalRegistrator } from '@happy-dom/global-registrator'
+
+beforeAll(() => {
+  if (!GlobalRegistrator.isRegistered) GlobalRegistrator.register()
+})
+
+describe('hydrate sets currentScope before calling init', () => {
+  test('child init can resolve a parent-provided context value', async () => {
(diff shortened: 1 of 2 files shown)


kfly8:
Closing — wrong layer. `hydrateComponent` setting `currentScope` would mask symptoms (orphan children) instead of fixing the root cause (why bridges emitted from `props.renderNode(node)` end up as top-level scopes at all). The bridge should be marked as a child via `__bfChild=true` + `__bfScope=<parent>` at IR/emit time, then DOM walker skips it (BF_CHILD_PREFIX check) and Flow's initChild does its job. Re-doing as a CLI-side render-prop detector.

PR #1183, opened 2026-05-04
refactor(jsx): unify Provider/Async required-prop errors under ErrorCode (BF046)

kfly8:
## Summary

Health-check finding #1: four `throw new Error(...)` sites in `packages/jsx/src/jsx-to-ir.ts` (lines 505, 529, 553, 612) bypassed the project's `BF###` error system. Missing `value` on `<X.Provider>` or `fallback` on `<Async>` aborted the whole compile with an uncategorised crash, while every other validation in the file accumulates structured `CompilerError`s on `ctx.analyzer.errors`.

This PR routes those four sites through `createError(ErrorCodes.COMPONENT_REQUIRED_PROP_MISSING, ...)` and returns `null` so:

- The compiler keeps walking and can report multiple problems per build.
- IDE / CLI surfaces get a stable code (`BF046`) and severity, matching the rest of the diagnostic API.
- Failure mode now matches sibling validations like `MISSING_KEY_IN_LIST` (BF023).

## Changes

- `packages/jsx/src/errors.ts` — add `COMPONENT_REQUIRED_PROP_MISSING: 'BF046'` and its default message; flatten the misleading "Local function errors (BF045-BF049)" sub-grouping (only BF045 lived there).
- `packages/jsx/src/jsx-to-ir.ts` — replace the four `throw`s with `ctx.analyzer.errors.push(createError(...))` + `return null`. Per-site message keeps the original tag/prop info via the `message` override. Return types of `transformProviderElement`, `transformSelfClosingProviderElement`, `transformAsyncElement`, `transformSelfClosingAsyncElement`, `transformJsxElement`, `transformSelfClosingElement` widened to `IRNode | null` (already accepted by callers `transformNode` / `transformJsxExpression` / `transformChildren`).
- `packages/jsx/src/__tests__/ir-async.test.ts` — rewrite the `toThrow('fallback')` case to assert `BF046` on `ctx.errors`.
- `packages/jsx/src/__tests__/ir-provider.test.ts` — add the symmetric BF046 cases for `<X.Provider>` (was previously untested) and self-closing `<X.Provider />`.

## Out of scope

This is item #1 of the jsx-area health-check (see prior agent report). The remaining items — `processAttributes` / `processComponentProps` DRY-up, and finishing the staged-IR `OriginInfo` emit so BF060/BF061 can flip default-on — will land as separate PRs.

## Test plan

- [x] `bun test src/__tests__/ir-async.test.ts src/__tests__/ir-provider.test.ts` — 15 pass
- [x] `bun test` — 955 pass; 4 pre-existing unrelated failures in `primitive-resolver-alias` / resolver-migration suites (verified failing on `main` before edits)
- [x] `bun run typecheck:tests` — clean
- [x] `bun run build` — clean

https://claude.ai/code/session_01B4jyJXynN2jpGvKZ8U7Y36

---
_Generated by [Claude Code](https://claude.ai/code/session_01B4jyJXynN2jpGvKZ8U7Y36)_

### DIFF
--- a/packages/jsx/src/__tests__/ir-async.test.ts
+++ b/packages/jsx/src/__tests__/ir-async.test.ts
@@ -95,7 +95,7 @@ describe('<Async> streaming boundary', () => {
     expect(asyncNode.fallback.type).toBe('component')
   })
 
-  test('reports BF046 when fallback prop is missing and emits a transparent stub', () => {
+  test('reports BF046 when fallback prop is missing and emits a fragment stub', () => {
     const source = `
       export function Page() {
         return (
@@ -135,6 +135,15 @@ describe('<Async> streaming boundary', () => {
     expect(error?.severity).toBe('error')
     expect(error?.message).toContain('fallback')
     expect(ir?.type).toBe('fragment')
+
+    // Empty stub at root must NOT emit `needsScopeComment` — a bare
+    // bf-scope comment would let the runtime fall back to
+    // comment.parentElement, hydrating the broken component against its
+    // parent container instead of an isolated boundary.
+    if (ir?.type === 'fragment') {
+      expect(ir.children.length).toBe(0)
+      expect(ir.needsScopeComment).toBeUndefined()
+    }
   })
 
   test('compileJSX surfaces BF046 in errors without crashing on multi-child stub', () => {
--- a/packages/jsx/src/__tests__/ir-async.test.ts
+++ b/packages/jsx/src/__tests__/ir-async.test.ts
@@ -1,7 +1,11 @@
 import { describe, test, expect } from 'bun:test'
 import { analyzeComponent } from '../analyzer'
 import { jsxToIR } from '../jsx-to-ir'
+import { compileJSX } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
 import type { IRAsync, IRElement } from '../types'
+
+const adapter = new TestAdapter()
 
 describe('<Async> streaming boundary', () => {
   test('transforms <Async> with fallback and children into IRAsync', () => {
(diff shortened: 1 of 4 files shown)


kfly8:
All five Copilot review comments addressed in `c7c200d`.

**Comments 1 & 2 (nullable dispatcher dropped subtrees):** Reverted `transformJsxElement` / `transformSelfClosingElement` to non-null. Verified that `parseFallbackProp` (line 591) and `transformJsxFunctionCall` (line 1003) coerced `null` into a text node and `expr: 'null'` respectively — that path is now dead again.

**Comments 3 & 4 (early return suppressed descendant diagnostics):** Both Provider and Async paths now call `transformChildren(node.children, ctx)` before returning, so BF023/BF025/BF044/etc. inside the broken subtree still accumulate. Matches the multi-error behaviour the PR description claims.

**On-failure shape:** `reportMissingRequiredProp` now just pushes the diagnostic; the call site returns a transparent `IRFragment` stub wrapping the (walked) children. Consistent with how every other validation in this file behaves — diagnostic is the source of truth, consumers fail the build on `severity:'error'`.

**Comment 5 (missing self-closing test):** Added `reports BF046 when self-closing <Async /> is missing fallback`.

Test results: 956 pass, 4 pre-existing unrelated failures in resolver-migration suites (verified failing on `main`).

---

kfly8:
All three Copilot comments addressed in `538c743`.

**Empty stub leaked scope to parentElement (line 529):** For self-closing `<Async />` / `<Ctx.Provider />` at root, the stub still set `needsScopeComment`. Adapters emitted a bare `bf-scope` comment with no element sibling, so the runtime fell back to `comment.parentElement` as the proxy scope — hydrating the broken component against its parent container. Now gated on `children.length > 0` so empty stubs are inert. Pinned by new assertions on both self-closing tests.

**Misleading "consumers fail the build" comment (line 512):** Reworded to make the consumer's responsibility explicit: `compileJSX` keeps emitting files when the IR root is non-null, so consumers must check `result.errors` and fail on `severity:'error'`.

**"transparent stub" test name (line 98):** Renamed to "fragment stub" to match the implementation, which deliberately does *not* set `transparent`.

Test results: 958 pass, 4 pre-existing unrelated failures.

---

PR #1215, opened 2026-05-08
fix(runtime): track multi-root Fragment items as one unit in mapArray (#1212)

### LINKED ISSUE #1212: runtime: mapArray misaligns hit/visible pairs when child component returns a multi-element Fragment
## Symptom

When a list-rendered component returns a Fragment with multiple top-level elements, `mapArray` mis-pairs DOM elements across keys after the first hydration pass — visible elements walk one slot per key, and one item per N drops out entirely.

## Concrete case

`<SimpleEdge>` in `ui/components/ui/xyflow/index.tsx` used to return:

```tsx
return (
  <>
    <path data-hit-id={edgeId} ... />
    <path data-id={edgeId}     ... />
  </>
)
```

`<Flow>` drives one `<SimpleEdge>` per edge through `mapArray(visibleEdges, ...)`. SSR rendered all 8 paths correctly for 4 edges (e1-2, e1-3, e2-4, e3-4). After hydration the SVG had:

```
visible: e1-3, e3-4, e2-4, e3-4   // e1-2 missing, e3-4 duplicated
hit:     e1-2, e2-4, e2-4, e3-4   // e1-3 missing, e2-4 duplicated
```

i.e. mapArray treated each `<path>` as one entry instead of recognising the 2-element Fragment as one logical child per key.

## Reproduction in this repo

`/xyflow/components` Overview demo (`XyflowPreviewDemo`) with the original `<SimpleEdge>` Fragment shape — the Source→Pipeline edge was missing in the rendered view. The DOM evidence is captured in PR #1209.

## Workaround we shipped

Wrap the two paths in a `<g>` so each `<SimpleEdge>` is one top-level SVG element (PR #1209, commit b35c91f5). mapArray now keys 1:1 with edges and stays in sync.

## Why it matters as a runtime issue

The `<g>` wrap is fine for SimpleEdge, but the underlying limitation will bite anyone who writes a list-rendered component that legitimately wants to emit a Fragment of multiple peers (e.g. a `<DataRow>` returning a `<tr>` plus a hidden expandable `<tr>`). Either:

- mapArray should track N-elements-per-key, or
- the compiler should require/wrap a single root for components used inside `mapArray`.

kfly8:
## Summary

`mapArray` previously assumed one DOM element per key, so a list-rendered child returning a JSX Fragment with two or more sibling roots silently misaligned. On SSR hydration `existingChildren[i]` was paired with `items[i]`, walking the visible/hit pairs off by one slot per key (one item per N dropped out entirely). On CSR the template-clone path only cloned `firstElementChild`, dropping every subsequent sibling root. Reactive attribute lookups via `qsa(__el, ...)` likewise only ever saw slots inside the primary element.

This fix introduces per-item boundary markers `<!--bf-loop-i-->` that the compiler emits **only** when the loop body is multi-root (`bodyIsMultiRoot` IR flag). The runtime tracks each item as a `(startMarker, primaryEl, extras[])` triple and operates on the whole range as one unit for hydration, insert, move, and remove. A new `qsaItem` helper walks past `__el` into its marker-bounded siblings so reactive attributes / events on every root bind. Single-root loops (the common case) keep their existing emission verbatim — zero overhead.

Surfaced by `<SimpleEdge>` in `ui/components/ui/xyflow/index.tsx` (Source→Pipeline edge missing in `XyflowPreviewDemo` after hydration). PR #1209 shipped a `<g>` wrap as a workaround for that specific component; the underlying limitation now lifts so any list-rendered component (e.g. `<DataRow>` returning `<tr>` + a hidden expandable `<tr>`) can return a multi-element Fragment.

## Changes

- **Compiler** — new `bodyIsMultiRoot` flag in `IRLoop` / Plan types, threaded through every loop emission site (top-level, composite, branch, inner). Multi-root template clone helper in `template-parse.ts`. `qsa` → `qsaItem` switch in reactive-effects / event-listener emitters when the flag is set.
- **SSR adapters** — Hono and Go-template `renderLoop` prepend a per-item start marker inside the `.map()` / `{{range}}` body when `bodyIsMultiRoot`.
- **Runtime** — `ItemScope` now carries `startMarker` + `primaryEl` + `extras`. `findItemRanges` partitions the loop range by `<!--bf-loop-i-->` markers, falling back to one-element-per-item when no markers are present (legacy SSR fallback). `insertScope` / `removeScope` move whole ranges. New `qsaItem(primaryEl, selector)` and `getLoopNodes(container, markerId)` helpers.
- **Tests** — new compiler unit (3 tests), runtime unit (5 tests), adapter conformance fixture (`fragment-loop-children.ts`). `normalizeHTML` extended to strip `<!--bf-loop-i-->`.
- `<SimpleEdge>` keeps the `<g>` wrap (cleaner SVG hygiene, retains `data-edge-id` selectors). Synthetic regression test in `fragment-loop-children` fixture covers the original Fragment shape.

## Test plan

- [x] `bun test packages/client packages/jsx packages/adapter-hono packages/adapter-tests packages/adapter-go-template` — all green (1668 pass / 0 fail / 2 skip)
- [x] No existing fixture or snapshot needed updating (verified — no current fixture exercises Fragment-as-loop-child)
- [x] `xyflow.spec.ts` (existing e2e) keeps passing via the retained `<g>` wrap
- [x] Conditional emission means single-root loops produce byte-identical output to before

Closes #1212


kfly8:
Thanks for the review — all three issues are addressed in 8184832d:

**1. `qsaItem` boundary stops (`packages/client/src/runtime/qsa-item.ts:42`)**
Added hard stops for both `<!--bf-loop:*-->` and `<!--bf-/loop:*-->` markers in the sibling walk. The last item in a loop can no longer escape its range — verified by a new regression test: a `<div bf="trap">` placed after `<!--bf-/loop:l0-->` is unreachable from the last item's `qsaItem` call.

**2. CSR-time `__bfExtras` lookup (`template-parse.ts:294`)**
When `__el.nextSibling` is `null` (the renderItem body runs before `mapArray` inserts the new item), `qsaItem` now falls back to walking `__el.__bfExtras`. The stash is read-and-deleted by `mapArray.createItemScope` after `renderItem` returns, so reactive-effect/event setup *inside* the body still has access to it. Verified by a regression test that asserts a slot living in an extra root is reachable while the primary and extra are still detached.

**3. `upsertChildItem` for child components in extras (`shared.ts:244`)**
Added `upsertChildItem` (paired with `qsaItem` in the same file). Same iteration order — primary descendants → in-tree siblings within bounds → `__bfExtras` — and uses `qsaItem`-style root-or-descendant matching so it also works when a sibling root *is* the component scope element itself. The compiler now emits `upsertChildItem(__el, ...)` instead of `upsertChild(__el, ...)` when `bodyIsMultiRoot`.

Refactored the root walk into a shared `itemRootElements` generator so the two helpers can't drift apart.

Test coverage: 4 new runtime tests in `map-array-multi-root.test.ts`. Full suite still green (1672 pass / 0 fail).

Issue #1227, opened 2026-05-11
compiler silently strips 'use client' imports from non-client .ts files (runtime ReferenceError)


kfly8:
## Summary

When a `.ts` file (without `'use client'`) imports a sibling `.tsx` file marked `'use client'`, the barefoot compiler removes the import at bundle time **without emitting a warning or error**. The reference survives in the bundled code, so the page crashes at runtime with `ReferenceError: X is not defined`.

This makes incremental refactors away from a large imperative `.ts` module to a JSX child component very hard to land — the symptom shows up only on the live page (not in the build log, not in `bun run typecheck`, not in unit tests).

## Environment

- `barefootjs` at `2072acd5` (via `github:piconic-ai/barefootjs#2072acd5` in piconic-ai/desk)
- Bun 1.3.0
- The repro below comes from piconic-ai/desk; the workaround is in PR https://github.com/piconic-ai/desk/pull/103 and the original revert that drove this report is at commit `884173f` (https://github.com/piconic-ai/desk/commit/884173f).

### Files

```ts
// nodes/DraftTitleEditor.tsx — `'use client'` JSX component
'use client'
export function DraftTitleEditor(props: { initialTitle: string }) {
  return <textarea value={props.initialTitle} />
}
```

```ts
// nodes/IssueCardNode.ts — imperative `.ts`, NO `'use client'`
import { DraftTitleEditor } from './DraftTitleEditor'

export function IssueCardNode(this: HTMLElement, props: { id: string }) {
  // ... 1000 lines of createElement-based DOM ...
  this.appendChild(DraftTitleEditor({ initialTitle: '' }))
}
```

```ts
// DeskCanvas.tsx — `'use client'` JSX, registers IssueCardNode as a nodeType
'use client'
import { IssueCardNode } from './nodes/IssueCardNode'
export function DeskCanvas() {
  return <Flow nodeTypes={{ issueCard: IssueCardNode }}>...</Flow>
}
```

### Observed `barefoot build` output

```
Inlined: ./nodes/IssueCardNode into components/DeskCanvas.client.js
Stripped server component import: ./nodes/DraftTitleEditor from components/DeskCanvas.client.js
Build complete: 18 compiled, 0 cached, 6 skipped, 0 errors
```

The `Stripped server component import:` line is the only signal — it's emitted at info level, not as a warning/error, and the build exit code is `0`.

### Runtime failure

```
ReferenceError: DraftTitleEditor is not defined
[...]

PR #1241, opened 2026-05-12
fix(cli): rewrite stripped 'use client' .tsx imports as runtime stubs (#1240)

### LINKED ISSUE #1240: bundler strips JSX-component imports passed through Record/non-JSX values, even though FlowNodeTypeBridge now accepts them at runtime (#1238 follow-up)
## Summary

#1238 fixed the runtime side of #1236 — `FlowNodeTypeBridge` now appends the JSX-shim's returned DOM element when an entry in the `nodeTypes` map yields one. That makes JSX-component dispatch via `nodeTypes` work *at runtime*.

But the bundler still strips the import before the runtime ever sees it: passing a JSX-component reference into a `Record<string, fn>` literal counts as "non-JSX usage" and triggers BF053. The two layers disagree about what's valid, so the runtime fix can't actually be exercised by app code that registers JSX components in a `nodeTypes` map.

This blocks the migration path the BF053 hint itself recommends ("Move the call site into a 'use client' .tsx parent and render … as JSX") for projects that want to keep their existing `nodeTypes` shape during a piecewise rewrite.

## Environment

- `barefootjs` at `baa504af` (latest `main`, includes #1227 / #1228 / #1238 fixes)
- piconic-ai/desk PR https://github.com/piconic-ai/desk/pull/103 (the workaround that survives the strip)

### Files

```ts
// nodes/IssueCardNode.tsx — `'use client'` JSX-shim form
'use client'
import { DraftTitleEditor } from './DraftTitleEditor'

export function IssueCardNode(this: HTMLElement, props: { id: string }) {
  // (imperative body internally; uses DraftTitleEditor as a JSX child via direct import)
  this.appendChild(DraftTitleEditor({ initialTitle: '' }) as unknown as HTMLElement)
}
```

```tsx
// nodes/index.tsx — `'use client'` re-export
'use client'
import { IssueCardNode } from './IssueCardNode'
import { BoxNode } from './BoxNode'
// biome-ignore lint/suspicious/noExplicitAny: see nodeTypes typing
export const nodeTypes: Record<string, (this: HTMLElement, props: any) => void> = {
  issueCard: IssueCardNode as unknown as (this: HTMLElement, props: any) => void,
  box: BoxNode,
}
```

```tsx
// DeskCanvas.tsx — `'use client'` parent that mounts <Flow nodeTypes={…}>
'use client'
import { Flow } from '../ui/xyflow'
import { nodeTypes } from './nodes'

export function DeskCanvas() {
  return <Flow nodeTypes={nodeTypes}>…</Flow>
}
```

### Observed `barefoot build`

```
Stripped client component import: ./nodes from components/canvas/DeskCanvas.client.js
Stripped client component import: ../nodes/IssueCardNode from components/canvas/catalog/IssueCardCatalog.client.js
ERROR[BF053]: Import `nodeTypes` from './nodes' was stripped from the client bundle `components/canvas/DeskCanvas.client.js`, but `nodeTypes` is still referenced in the assembled JS. Client components ('use client' .tsx) are not callable as plain functions from imperative .ts modules — render them as JSX from a 'use client' parent instead.
ERROR[BF053]: Import `IssueCardNode` from '../nodes/IssueCardNode' was stripped from the client bundle `components/canvas/catalog/IssueCardCatalog.client.js`, but `IssueCardNode` is still referenced in the assembled JS.
```

The strip fires across two boundaries:
1. `DeskCanvas` (`'use client'`) → `nodes/index.tsx` (`'use client'` re-export) → BF053 on `nodeTypes`.
2. `IssueCardCatalog` (`'use client'`) → directly imports `IssueCardNode.tsx` (`'use client'`) — also stripped, even though both endpoints are client.

Both consumers ARE `'use client'` and both pass the entry through `<Flow nodeTypes={…}>`. The runtime bridge would handle the JSX shim correctly per #1238; the bundler simply never lets it get there.

### Workaround currently used in piconic-ai/desk

`data._draftTitleEditor` plumbing + a hidden `<DraftTitleEditor />` JSX touchpoint in DeskCanvas. The hidden JSX use saves the import; the data plumbing routes the JSX shim into the imperative card. See https://github.com/piconic-ai/desk/pull/103.

## Expected behaviour

The bundler's strip rule for `'use client'` imports should treat references that flow into `<Flow nodeTypes={…}>` (and similar JSX-prop bindings) as live. After #1238 the runtime accepts the shim, so the binding is no longer dead in the assembled output — strip is a false positive.

A few possible shapes the fix could take:

1. **Look for the import being passed to a JSX prop value**, not just used inside a JSX child position. Today the strip seems to require the import name to appear as a `<X />` JSX child; assignments like `nodeTypes={{ key: X }}` (or `array.map(() => X)`) don't count even though the value reaches a JSX prop.
2. **Trust runtime bridges**: stop stripping when the `'use client'` import is referenced from another `'use client'` module at all. The runtime contract is now well-defined for the JSX-component shim.
3. **Module-level "client export" annotation** so the bundler knows a particular `'use client'` export is intentionally invoked programmatically (`export const __barefootClientExport = true` or similar). More verbose but explicit.

Option 2 has the smallest blast radius for existing apps — it just removes a too-aggressive heuristic now that #1238 makes the strip's premise ("non-JSX usage cannot work") wrong.

## Why this matters for piconic-ai/desk

After #1227 + #1228 + #1238 we tried the simplification on PR #103: drop the data plumbing and the hidden JSX, convert `IssueCardNode.ts` → `.tsx` `'use client'`, register it in the `nodeTypes` map directly. Build fails with the BF053 above on every consumer (DeskCanvas + every catalog page) — there's no path to land the simplification short of also migrating every `<Flow nodeTypes={…}>` site to `<Flow renderNode={(n) => <NodeDispatcher node={n} />}>`. That's a much bigger refactor than the JSX-component conversion itself.

For desk that means the `data._draftTitleEditor` workaround stays in place until either this strip-rule loosens or we commit to the `renderNode` API migration.

## Related

- piconic-ai/barefootjs#1227 (silent strip → loud BF053; landed)
- piconic-ai/barefootjs#1228 (nested-arrow `let X = Y` analyzer; landed)
- piconic-ai/barefootjs#1236 (this one's runtime sibling; #1238 landed the runtime fix only)
- piconic-ai/desk#86 (the larger TSX rewrite this blocks)
- piconic-ai/desk#103 (the workaround PR — see commit `b38dfc6` for the simplification attempt that surfaced this bug)

kfly8:
Closes #1240. Pairs with #1227 (silent strip → BF053) and #1238 (`FlowNodeTypeBridge` accepts JSX-shim entries) to make `'use client'` `.tsx` components actually usable through `<Flow nodeTypes={…}>` end to end.

## Problem

After #1238 the runtime bridge does the right thing with a JSX-component shim entry in `nodeTypes`, but the bundler half stayed in the original strip-and-leave-dangling shape: passing a JSX-component reference into a Record literal counts as "non-JSX usage" and triggers BF053. The two layers disagreed about what's valid, so the runtime fix wasn't actually reachable from app code that registers JSX components in a `nodeTypes` map — the only path forward was rewriting every `<Flow nodeTypes={…}>` site to `<Flow renderNode={…}>`, a much larger refactor than the JSX migration itself.

## Fix

Replace the strip behaviour for **named** `.tsx` imports with a local runtime stub that delegates to `createComponent`:

```js
import { X } from './X'           // before
const X = (props, key) => createComponent('X', props, key)  // after
```

The stub works in every reference shape — JSX (`<X />`, where the JSX compiler also rewrites to `createComponent` and never touches the binding), Record literal (`<Flow nodeTypes={{ kind: X }}>`), plain function call from an inlined `.ts` helper. The runtime resolves the binding via the existing hydrate registry, the same path JSX usage already takes.

`createComponent` is in scope on every JSX-emitted client bundle (it lands as part of the umbrella `barefoot.js` import that every `<X />` needs), so the stub doesn't add a new module dependency. We deliberately do NOT hoist a separate `import { createComponent } from '@barefootjs/client/runtime'` — the page's importmap only exposes the umbrella `barefoot.js` module, so the bare runtime path would 404 in the browser. The realistic case of "consumer doesn't have `createComponent` in scope" — a non-JSX bundle accidentally importing a `'use client'` component — already wouldn't have worked under the old strip path either, and now produces a clearer "createComponent is not defined" error pointing at the stub line.

Default and namespace imports of `.tsx` fall back to the pre-#1240 strip + BF053 path. The registered runtime name for those isn't trivially recoverable from the import statement alone — default exports compile without a stable JSX-registration name, namespace imports would need every export of the target file enumerated against the registry. The named-import path covers the practical case (`nodeTypes={{ kind: Component }}`) that drove this issue.

## Test plan

- [x] `bun test packages/cli/src/__tests__/resolve-imports.test.ts` — 16 pass / 0 fail. New cases:
  - "replaces a named .tsx client-component import with a runtime-resolving stub" — happy path
  - "replaces .tsx binding called from an inlined .ts helper with a runtime stub instead of stripping" — closes the #1227 dangling-reference case for named imports (the binding now has a definition, so no BF053)
  - "still fires BF053 for namespace .tsx imports"
  - "still fires BF053 for default .tsx imports"
- [x] Existing cli tests (12) updated where the previous behaviour expected a strip-only outcome
- [x] `bun test packages/cli/` — 396 pass / 1 pre-existing fail unrelated (mojo adapter `runtimes.generated` missing)

### DIFF
--- a/packages/cli/src/__tests__/resolve-imports.test.ts
+++ b/packages/cli/src/__tests__/resolve-imports.test.ts
@@ -49,7 +49,7 @@ console.log(highlight('hello'))
     expect(result).toContain("from '@barefootjs/client'")
   })
 
-  test('strips .tsx client component import when binding is unused', async () => {
+  test('replaces a named .tsx client-component import with a runtime-resolving stub (#1240)', async () => {
     writeFileSync(resolve(COMPONENTS_DIR, 'ClientComp.tsx'), `'use client'
 export function ClientComp() {
   return <div>client only</div>
@@ -68,21 +68,33 @@ console.log('client code')
     const { errors } = await resolveRelativeImports({ distDir: DIST_DIR, manifest })
 
     const result = await Bun.file(resolve(COMPONENTS_DIR, 'Parent-abc123.js')).text()
-    expect(result).not.toContain('ClientComp')
+    // The original `import` statement is gone — replaced with a stub that
+    // delegates to `createComponent` so the local binding works in any
+    // call shape (JSX, Record literal, .call(el, props), …).
+    expect(result).not.toContain("from './ClientComp'")
+    // The stub must keep the original local binding name.
+    expect(result).toContain('const ClientComp =')
+    expect(result).toContain("createComponent(\"ClientComp\", props, key)")
+    // The stub relies on the host bundle already importing `createComponent`
+    // from the umbrella `barefoot.js` runtime (every JSX-emitted client
+    // bundle does, since `<X />` compiles to `createComponent(...)`). We
+    // deliberately do NOT hoist a `from '@barefootjs/client/runtime'` line
+    // — that bare specifier isn't on the page's importmap and would 404 in
+    // the browser. See the comment on the stub branch in resolve-imports.ts.
+    // Pre-existing imports + body lines stay intact.
     expect(result).toContain("from '@barefootjs/client'")
     expect(result).toContain("console.log('client code')")
-    // No dangling reference → no diagnostic.
+    // The binding has a working definition → no dangling reference, no diagnostic.
     expect(errors).toHaveLength(0)
   })
 
-  // Regression: bf#1227 — a `.ts` helper (no `'use client'`) inlined into a
-  // client bundle imports a sibling `'use client'` `.tsx` AND calls it. The
-  // strip used to be silent (`console.log` only, build exit 0), so the
-  // bundle shipped with a dangling identifier and crashed at runtime with
-  // `ReferenceError: DraftTitleEditor is not defined`. After the fix the
-  // strip is detected via a post-assembly identifier scan and surfaced as
-  // BF053, so the build exits non-zero.
-  test('errors when stripped .tsx binding is still referenced (bf#1227)', async () => {
+  // Regression: bf#1227 used to surface BF053 when a stripped `'use client'`
+  // import left a dangling reference. bf#1240 turns the strip into a
+  // runtime-resolving stub instead, so the same call site that previously
+  // crashed at runtime (the original `884173f` ReferenceError) now works.
+  // BF053 still fires for non-stubbable shapes — see the default + namespace
+  // tests below.
+  test('replaces .tsx binding called from an inlined .ts helper with a runtime stub instead of stripping (#1240, supersedes #1227 dangling case)', async () => {
     writeFileSync(resolve(COMPONENTS_DIR, 'DraftTitleEditor.tsx'), `'use client'
 export function DraftTitleEditor() { return <textarea /> }
 `)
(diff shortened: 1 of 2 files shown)


kfly8:
Filed #1242 to track the regex-splice → TS-AST splice cleanup across all strip / replace branches in this file. Out of scope for this PR (matches the existing convention of the file), but worth a follow-up since #1241 makes substitution more sensitive to splice fidelity than deletion ever was.

PR #1246, opened 2026-05-12
fix(jsx): static-loop dispatch picks empty SSR template, leaving DOM children unrendered when array is init-body-only (#1245)

### LINKED ISSUE #1245: bug(jsx): static-loop dispatch picks empty SSR template, leaving DOM children unrendered when array is init-body-only
## Repro

A `'use client'` component renders `.map()` over a component-local `const` whose initializer cannot be relocated to template scope (e.g. `Object.entries(props.x ?? {}).filter(...)`). At runtime the rendered DOM contains the outer container but **no children** — the loop appears to silently produce nothing.

```tsx
'use client'
export function Bar(props) {
  const entries = Object.entries(props.reactions ?? {}).filter(([, users]) => users.length > 0)
  return (
    <div>
      {entries.map(([emoji, users]) => (
        <button key={emoji} type="button" onClick={() => props.onClick(emoji)}>
          <span>{emoji}</span>
          <span>{String(users.length)}</span>
        </button>
      ))}
    </div>
  )
}
```

After `createComponent('Bar', { reactions: { '👍': ['alice', 'bob'] } })`:

```html
<div bf="s3"><!--bf-loop:l0--><!--bf-/loop:l0--></div>
```

…three chips expected, zero rendered.

## Root cause

Two cooperating compiler stages disagree about whether the array is reachable from template scope:

1. **`html-template.ts:929-944` (template emitter)** — `transformExpr(node.array, …)` returns `UNSAFE_TEMPLATE_EXPR` for `entries` because it's in `unsafeLocalNames`. The substitution path then emits `${[].map(…)}`. The comment ("init's reconcile pass populates the loop") is correct **only for the dynamic `mapArray` path** — see (3).
2. **`jsx-to-ir.ts:2324`** — `isStaticArray = !isSignalOrMemoArray(array, ctx) && !hasCalls`. `entries` is a bare identifier (not a signal call, no function calls), so `isStaticArray = true`.
3. **`control-flow.ts:60`** dispatch:
   ```ts
   if (elem.isStaticArray) emitStaticArrayUpdates(...)   // ← chosen
   else emitDynamicLoopUpdates(...)                      // would have used mapArray
   ```
   `stringifyStaticLoop` (in `control-flow/stringify/loop.ts:88-133`) emits a `forEach` that binds reactive effects on `containerVar.children[childIndexExpr]` — **assuming the SSR template already rendered one DOM child per array entry**. But (1) emitted `${[].map(…)}` so SSR rendered zero children, and the `if (__iterEl)` guard silently skips every iteration.

The static path has no codegen branch that creates children when the SSR shape is empty — only `mapArray`'s reconciliation does.

## Test guarding the broken behavior

`packages/jsx/src/__tests__/template-closure.test.ts:179-198` (`'init-scope loop array becomes [] — avoids \`undefined.map(...)\` TypeError'`) asserts the template substitution but does NOT verify the runtime DOM. The CSR conformance suite skips this case explicitly:

```ts
// packages/adapter-tests/src/__tests__/csr-conformance.test.ts:25-28
// Local array variable (items) is not available at CSR template module scope.
// CSR templates only have access to props and signals, not file-scope constants.
'static-array-children',
```

## Proposed fix

Re-route the dispatch: when the array references an unsafe local name, force the dynamic emitter (`mapArray`). The dynamic path:
- Reads the live array in init scope (`entries` IS defined there)
- Materialises children from the per-iteration template clone source (which `TopLevelLoop` already carries on `elem.template` and which `buildPlainLoopPlan` already consumes)
- Binds reactive effects per item

The performance cost (vs. static loop) only matters when SSR rendered children — for unsafe arrays SSR rendered nothing, so the comparison is moot.

PR follows.

kfly8:
Closes #1245.

## Summary

When a `'use client'` component renders `.map()` over a component-local `const` whose initializer cannot be relocated to template scope (e.g. `Object.entries(props.x ?? {}).filter(...)`), the rendered DOM contains the outer container but **no children** — the loop silently produces nothing.

Two compiler stages disagreed about whether the array was reachable from template scope:

1. `html-template.ts` substituted `${[].map(...)}` for the unsafe identifier — SSR template with zero children
2. `jsx-to-ir.ts:2324` classified the loop as `isStaticArray = true` (the array expression itself is just a bare identifier — no signal call, no function calls)
3. `control-flow.ts` dispatched to the static path, which `forEach`s over the array and tries to attach reactive effects to `containerVar.children[idx]` — but the SSR template emitted nothing, so every iteration silently `if (__iterEl) { ... }` skipped

The static path has no codegen branch that creates children when the SSR shape is empty. Only `mapArray` does.

## Fix

Re-route in `emitLoopUpdates`: when the array references a name in `unsafeLocalNames`, dispatch to `emitDynamicLoopUpdates` instead. The dynamic path uses `mapArray`, which materialises children at init time from the per-iteration template clone source the `TopLevelLoop` element already carries on `elem.template` (consumed by `buildPlainLoopPlan`).

The static path's performance edge only matters when SSR rendered children — for unsafe arrays SSR rendered nothing, so the comparison is moot.

## Wiring

- Add `unsafeLocalNames: Set<string>` to `PhaseCtx` and thread the pre-computed value (`buildInlinableConstants`) through `generate-init.ts` once. The `loop-updates` phase and the later `emitRegistrationAndHydration` template stage now consume the same set, so they can't disagree about whether a given identifier is unsafe.
- `emitLoopUpdates` accepts the set and consults a small `arrayReferencesUnsafeName` whole-word scan. Loop arrays in practice are bare identifiers or short member chains (`items`, `state.items`), so a tiny regex is cheaper than an AST parse, and identifier names need no escaping.

## Test plan

- [x] New test `packages/jsx/src/__tests__/init-scope-loop-routing.test.ts` — verifies `clientJs` contains `mapArray(` and does NOT contain the broken `entries.forEach(` static-loop signature for the repro case. Confirmed red without the fix (the `forEach` + `children[__idx]` pattern produces empty DOM at runtime), green with it.
- [x] `bun test packages/jsx` — 1046 pass / 4 pre-existing fail (alias resolver tests, unrelated to this PR)
- [x] `bun test packages/adapter-tests` — 39 / 39 CSR conformance pass (no regression)
- [x] **Real-world end-to-end**: piconic-ai/desk's `IssueCardCatalog` now renders three `<ReactionBar>` chip buttons (`👍2`, `🎉1`, `❤️2`) for a card with `data.deskReactions = { '👍': ['alice','bob'], '🎉': ['alice'], '❤️': ['carol','alice'] }` and `data.currentUser = 'alice'`. Before this fix the same component rendered an empty `<div data-reaction-bar="true">`. Verified via Chrome DevTools / playwright MCP.

### DIFF
--- a/packages/jsx/src/__tests__/init-scope-loop-routing.test.ts
+++ b/packages/jsx/src/__tests__/init-scope-loop-routing.test.ts
@@ -1,0 +1,46 @@
+'use client'
+import { describe, test, expect } from 'bun:test'
+import { compileJSX } from '../compiler'
+import { TestAdapter } from '../adapters/test-adapter'
+
+const adapter = new TestAdapter()
+
+function getClientJs(source: string, filename: string): string {
+  const result = compileJSX(source, filename, { adapter })
+  expect(result.errors.filter(e => e.severity === 'error')).toHaveLength(0)
+  const clientJs = result.files.find(f => f.type === 'clientJs')
+  expect(clientJs).toBeDefined()
+  return clientJs!.content
+}
+
+describe('init-scope loop array routes through mapArray (#1245)', () => {
+  test('component-local const array derived from props uses mapArray', () => {
+    const source = `
+      'use client'
+      export function Bar(props) {
+        const entries = Object.entries(props.reactions ?? {}).filter(([, users]) => users.length > 0)
+        return (
+          <div>
+            {entries.map(([emoji, users]) => (
+              <button key={emoji} type="button" onClick={() => props.onClick(emoji)}>
+                <span>{emoji}</span>
+                <span>{String(users.length)}</span>
+              </button>
+            ))}
+          </div>
+        )
+      }
+    `
+    const clientJs = getClientJs(source, 'Bar.tsx')
+    // Before #1245 this loop hit the static-array path: `forEach` over the
+    // array in init scope, attaching reactive effects to pre-existing
+    // `containerVar.children[idx]` — but the template emitted `${[].map(...)}`
+    // for the unsafe array, so SSR rendered zero children and init's loop
+    // silently skipped every iteration. After #1245 the dispatch detects
+    // the unsafe array and routes through `mapArray`, which materialises
+    // children at init time.
+    expect(clientJs).toContain('mapArray(')
+    // Negative assertion: the broken static-loop signature should NOT appear.
+    expect(clientJs).not.toMatch(/entries\.forEach\(/)
+  })
+})
--- a/packages/jsx/src/analyzer.ts
+++ b/packages/jsx/src/analyzer.ts
@@ -1701,8 +1701,12 @@ function extractValueBranches(node: ts.Expression, ctx: AnalyzerContext): string
 /**
  * Extract free identifier references from an AST node by walking the tree.
  * Skips property keys, member access properties, and bound parameter names.
+ *
+ * Exported so other compiler stages (e.g. `jsx-to-ir.ts` for loop arrays)
+ * can record the same canonical free-identifier set on their IR nodes
+ * instead of re-deriving it from the printed expression string.
  */
-function extractFreeIdentifiersFromNode(node: ts.Node): Set<string> {
+export function extractFreeIdentifiersFromNode(node: ts.Node): Set<string> {
   const ids = new Set<string>()
   const boundNames = new Set<string>()
 
--- a/packages/jsx/src/ir-to-client-js/collect-elements.ts
+++ b/packages/jsx/src/ir-to-client-js/collect-elements.ts
@@ -526,6 +526,7 @@ export function collectElements(
         childComponent: l.childComponent,
         nestedComponents: l.nestedComponents,
         isStaticArray: l.isStaticArray,
+        arrayFreeIdentifiers: l.arrayFreeIdentifiers,
         useElementReconciliation,
         innerLoops: (useElementReconciliation || (l.isStaticArray && innerLoops?.length)) ? innerLoops : undefined,
         siblingOffset: siblingOffsets.get(l) || undefined,
--- a/packages/jsx/src/ir-to-client-js/control-flow.ts
+++ b/packages/jsx/src/ir-to-client-js/control-flow.ts
@@ -61,7 +61,7 @@ export function emitLoopUpdates(
   unsafeLocalNames: Set<string>,
 ): void {
   for (const elem of ctx.loopElements) {
-    if (elem.isStaticArray && !arrayReferencesUnsafeName(elem.array, unsafeLocalNames)) {
+    if (elem.isStaticArray && !arrayReferencesUnsafeName(elem, unsafeLocalNames)) {
       emitStaticArrayUpdates(lines, elem)
     } else {
       emitDynamicLoopUpdates(lines, elem)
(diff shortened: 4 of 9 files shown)


kfly8:
## Refactor: regex → IR-driven (6f812251)

Per CLAUDE.md ("Compiler internals — analysis, error codes, codegen") this codebase derives free identifiers from the source AST, not from printed expression strings. The dispatch helper added in the previous commit was scanning `elem.array` with a regex per `unsafeLocalNames` entry — fragile and out of step with the rest of the pipeline.

### Before
```ts
function arrayReferencesUnsafeName(arrayExpr: string, unsafeLocalNames: Set<string>): boolean {
  for (const name of unsafeLocalNames) {
    const re = new RegExp(`(^|[^\\w$])${name}([^\\w$]|$)`)
    if (re.test(arrayExpr)) return true
  }
  return false
}
```

### After
```ts
function arrayReferencesUnsafeName(elem: TopLevelLoop, unsafeLocalNames: Set<string>): boolean {
  const freeIds = elem.arrayFreeIdentifiers
  if (!freeIds || freeIds.size === 0) return false
  for (const id of freeIds) {
    if (unsafeLocalNames.has(id)) return true
  }
  return false
}
```

`arrayFreeIdentifiers` is populated in `jsx-to-ir.ts` via the same `extractFreeIdentifiersFromNode` AST walker that already fills `localConstants.freeIdentifiers`. The walker correctly skips property access RHS / property keys / arrow-bound parameter names, so chained access like `state.items` registers as `{state}` — letting us flag the loop unsafe when the root identifier is unsafe, without parsing the printed expression.

### Verified
- `bun test packages/jsx/src/__tests__/init-scope-loop-routing.test.ts` — green
- `bun test packages/jsx` — 1046 pass / 4 pre-existing fail (alias resolver, unrelated)
- piconic-ai/desk catalog still renders 3 chips (`👍2 🎉1 ❤️2`) end-to-end

kfly8:
Closing as superseded by **#1247** (which carries the redesign discussion).

## Why this approach didn't work

The "re-route static loops with unsafe arrays through `mapArray`" strategy in this PR fixes the original bug (#1245) — CSR mounts of `<ReactionBar reactions={…}>` correctly render their chip buttons — but it broke the **SSR-then-hydrate** path for components that previously worked. CI on `fix/1245-static-loop-init-only-array` showed ~50 e2e regressions across `site/ui`:

- All Props Playgrounds where `Select`/`Input` change a signal: alert, badge, button, kbd, empty
- Chart Playgrounds: area, bar, line, pie, radar, radial
- Interaction tests: carousel navigation, dialog overlay click, dropdown-menu checkbox/radio/destructive, hover-card data-state transitions, item playground, popover preview, scroll-area type-always

Per-iteration param wrapping (`tag` → `tag()` signal accessor) and `mapArray`'s hydrate-vs-render decision behave differently than the static path's direct `_s3.children[idx]` binding when SSR has already rendered the children — the regressions surface as missing reactive attribute / text updates and broken state-driven previews.

Local `bun test packages/jsx` and `bun test packages/adapter-tests` stayed green (1046 / 39 conformance), so the regression isn't visible at the unit-test layer. It only shows up under real e2e against the SSR+hydrate flow that `site/ui` exercises.

## What's salvageable

Both commits on this branch (regex version `3557feb0`, IR-free-ids refactor `6f812251`) build cleanly and keep the new compile-level test green. The `arrayFreeIdentifiers: Set<string>` field added to `IRLoop` / `TopLevelLoop` is independently useful for any future analysis that needs to detect "this loop's array is reachable from <some name set>" without re-parsing the printed expression — a redesign per #1247 may reuse it.

For now, the desk repository (where the original bug surfaced) is falling back to imperative children inside a JSX wrapper to keep `data._draftTitleEditor`-style plumbing out while waiting for the upstream fix.

Issue #1255, opened 2026-05-12
Adapter: trust IRProp.isLiteral as authoritative; remove regex-based JS-expression classification


kfly8:
## Problem

The Hono adapter (and any sibling adapter following the same pattern) re-decides "is this prop value a JS expression or a string literal?" via regex (`isJsExpression`) at emit time, even though the IR already carries `IRProp.isLiteral` as the authoritative answer from Phase 1. The regex is independently incomplete from the IR flag, so the two can disagree silently — the direct cause of `fill="var(--x)"` being emitted as `fill={var(--x)}` previously.

This is a sub-issue split out of #1248 (reactivity classification unification). #1248 unifies reactivity-related classification across phases; this issue addresses the same structural anti-pattern — IR fields exist but downstream emit re-derives them — for the non-reactivity case of JS-expression vs string literal.

## Current shape

- ``packages/adapter-hono/src/adapter/hono-adapter.ts:911–926`` — ``isJsExpression(value: string): boolean`` regex with three patterns (arrow function shape, function call shape, ``set*`` / ``handle*`` / ``on*`` naming).
- The IR already carries ``IRProp.isLiteral`` (``packages/jsx/src/types.ts``) — set by Phase 1 when the prop's source was a literal attribute (``foo=\"bar\"``) vs a JS expression (``foo={bar}``).
- The adapter currently consults the regex instead of (or before) the IR flag.

## Why this produces defects

1. **IR field exists but isn't trusted.** ``IRProp.isLiteral`` is the authoritative classification — set when the JSX parser saw a string-literal attribute vs an expression container. The regex masquerades as a second source of truth that can disagree with the first.
2. **Regex misclassifies CSS-shaped values.** ``var(--x)``, ``url(#g)``, and similar look like function calls to the naming-pattern regex; the IR knows they originated as string literals.
3. **The anti-pattern recurs.** This is structurally the same shape as #1248: downstream emit re-derives a classification the IR already has. #1244 §F lists it as a recurring design pattern that produces defects.

## Refactor direction

**Short-term.** Hono adapter's emit path consults ``IRProp.isLiteral`` first; the regex remains only as a fallback for IRs that lack the flag (legacy / external producers).

**Long-term.** Lift attribute-value representation from ``string + flags`` to a discriminated union:

\`\`\`ts
type AttrValue =
  | { kind: 'literal'; value: string }
  | { kind: 'expression'; expr: string; ast?: ts.Node }
[...]

PR #1256, opened 2026-05-12
refactor(jsx): unify reactivity classification across phases (#1248 + #1251)

### LINKED ISSUE #1248: Compiler: unify reactivity classification across phases
## Problem

BarefootJS uses three parallel, independently-running reactivity classification channels that can disagree on whether an expression is reactive: (1) TypeChecker-based brand detection in `reactivity-checker.ts` (Phase 1 JSX→IR), (2) regex-based signal/memo matching in `jsx-to-ir.ts` (Phase 1, fallback), and (3) string-level pattern matching in `ir-to-client-js/reactivity.ts` (Phase 2, emit). Each channel has incomplete overlap, so an expression can be correctly classified by one path but missed or misclassified by another, producing silent reactivity bugs.

## Current shape

- `packages/jsx/src/reactivity-checker.ts:28–33` — TypeChecker-based `Reactive<T>` brand detection; requires AST + type resolution; fails if types are unresolved (library adapters, virtual file paths, imports without declarations).
- `packages/jsx/src/jsx-to-ir.ts:210–226` — regex patterns built from `analyzer.signals` / `analyzer.memos` / `analyzer.propsParams`, using word-boundary + call-site shape (`\\b${getter}\\s*\\(`).
- `packages/jsx/src/jsx-to-ir.ts:3234–3255` — Phase 1 entry point `isReactiveExpression` tries TypeChecker first then falls back to regex; props always return true without signal tainting.
- `packages/jsx/src/ir-to-client-js/reactivity.ts:79–102` — Phase 2 string-level check `needsEffectWrapper`; operates on already-expanded prop values, regex-only, no AST or type context.
- `packages/adapter-hono/src/adapter/hono-adapter.ts:911–926` — `isJsExpression` regex (arrow-function shape, `set*`/`handle*` patterns) — a 4th classification path that re-decides string-vs-expression without consulting IR flags.

## Why this produces defects

1. **Type-resolution gaps.** A library getter such as `form.isSubmitting()` (from `@barefootjs/form`) carries `Reactive<T>` via TypeChecker. If types fail to resolve, Phase 1 regex misses it (`isSubmitting` is not in `analyzer.signals` / `analyzer.memos`) and Phase 2 has no chance to recover.
2. **Memo identifier without call site.** `const x = memo; return x` passes the memo as a value, not as `memo()`. Phase 1 regex requires `\\b${name}\\s*\\(`, so the bare reference falls through. TypeChecker still recognises the brand on the identifier, so Phase 1 and Phase 2 disagree on the same identifier within one component.
3. **Destructured / renamed props.** `const Comp = ({ value: renamed }) => …` — Phase 1 marks `renamed` reactive (prop name match heuristic), Phase 2's regex looks for `props.renamed` and never finds the renamed local, so the two channels diverge.
4. **Adapter regex masquerading as classifier.** The Hono adapter's `isJsExpression` was the direct cause of `fill="var(--x)"` being emitted as `fill={var(--x)}`: the IR's `isLiteral` flag was authoritative but the adapter consulted the regex.

## Repro

```tsx
// Case 1 — library getter without resolved type
const Form = (props: { form: any }) => {
  // Type-resolved compile: Reactive<T> brand recognised, wrapped.
  // Type-unresolved compile (virtual path): all three channels miss it.
  return <div>{props.form.isSubmitting()}</div>
}

// Case 2 — memo as value, not as call
const memo = createMemo(() => count() + 1)
const Component = () => {
  const x = memo  // brand on `memo`, Phase 1 regex requires `memo(`
  return <div>{x}</div>
}

// Case 3 — destructured / renamed prop
const Comp = ({ value: renamed }: { value: string }) => {
  const x = renamed
  return <div>{x}</div>  // Phase 1 wraps via prop-name match; Phase 2 doesn't see `props.renamed` in expanded string
}
```

Expected: all three channels arrive at the same classification.
Actual: channels silently disagree; whichever fires last wins.

## Refactor direction

**Short-term.** Annotate every `IRExpression` / `IRAttribute` / `IRConditional` with a `reactivityReason` field naming the channel that classified it. Pass `ctx.signals` / `ctx.memos` / expanded prop references (not just param names) to Phase 2 so its string match has the same identifier set as Phase 1.

**Long-term.** Collapse the three channels into one `ReactivityResolution` working on normalised IR (not raw AST text), returning a structured `ReactivityProof` (`{ kind: 'type-brand' | 'signal-call' | 'prop-reference', via: string }`). Emit phases consume the proof, never re-classify. Library-getter cases require a resolved `ts.Program` at compile time — once that lands, the regex fallback is deletable.

## Effort / risk

**Medium / large.** Refactor touches the Phase 1 IR schema (new `reactivityReason` field), harmonises pattern construction between `jsx-to-ir` and `ir-to-client-js`, and threads checker context to Phase 2. Risk: library-adapter pattern coverage gaps; potential regression in wrap-by-default heuristics if regex patterns are tightened.

## Acceptance criteria

- [ ] `IRExpression`, `IRAttribute`, and `IRConditional` carry `reactivityReason` tracing classification source.
- [ ] Phase 2's `needsEffectWrapper` receives full signal / memo / props metadata from Phase 1 context; no in-emit regex re-scan of `ctx.signals` / `ctx.memos`.
- [ ] Library getter without resolved type is handled in Phase 1 (with explicit `unresolved-type` reason) — Phase 2 stops being a safety net.
- [ ] Test suite covers (a) library getter without resolved type, (b) memo passed as value, (c) destructured / renamed prop, (d) cross-phase consistency over the same fixture.
- [ ] Hono adapter's `isJsExpression` consults `IRProp.isLiteral` first; regex remains only as a fallback for IRs that lack the flag.

## References

Sub-issue of #1244 (compiler stress test catalog).

### LINKED ISSUE #1251: Compiler: complete staged-IR migration (origin / freeRefs)
## Problem

The staged-IR migration that introduced `OriginInfo` (phase / scope / effect / freeRefs) on expression-bearing IR nodes is incomplete. All fields are declared optional; the analyzer populates them at some sites but not uniformly; and emit passes (`relocate`, `compute-inlinability`) keep two parallel code paths — one that consults `origin`, one that falls back to legacy `templateValue` / `templateExpr` inference. The dual paths hide silent regressions: adding a new expression kind without populating `origin` does not error, it just routes through the legacy heuristic, which can disagree with the modern path in edge cases.

## Current shape

- `packages/jsx/src/types.ts:144–157` — `OriginInfo` interface; `phase` / `scope` / `effect` declared optional; `freeRefs` optional.
- `packages/jsx/src/analyzer.ts:1234, 1807, 1980–1984` — three sites populate `origin` (init statements, some literals, local constants). Other expression sources omit it.
- `packages/jsx/src/ir-to-client-js/relocate.ts:246–248` — `relocate()` prefers AST-based `collectFreeRefs()` when an AST node is available; falls back to regex-based `scanRefsByName()` when `freeRefs` is absent.
- `packages/jsx/src/ir-to-client-js/relocate.ts:558–568` — `scanRefsByName()` performs crude word-boundary matching against `bindings` keys; cannot resolve identifiers, only detect mentions.
- `packages/jsx/src/ir-to-client-js/compute-inlinability.ts:18, 136, 201, 439, 453, 461` — fallback branches labeled "legacy" that consult `templateValue` / `templateExpr` when `origin` is missing.

## Why this produces defects

1. **Silent degradation when `origin` is absent.** A new expression kind added to the IR but missed by the analyzer doesn't error — the emit path quietly takes the legacy branch. The legacy branch's classification can differ (more conservative wrapping, missed free refs), so behaviour diverges from the canonical `origin`-driven path.
2. **`scanRefsByName` heuristic loses shadow / closure info.** When `freeRefs` is missing, `scanRefsByName` walks the expression string and matches identifiers against `bindings`. It cannot distinguish a property name from a value, a shorthand `{ x }` from a reference to `x`, or a shadowed local from the outer binding. The "free refs" it produces can differ from what `collectFreeRefs` (AST-aware) would have produced.
3. **Migration debt compounds.** Each new pass either consults `origin` (and breaks when absent) or falls back (and accumulates legacy code). Over time, the codebase has two coexisting models of the same information, and no compile-time enforcement of which is authoritative.

## Repro

```tsx
// Hypothetical: a new IR node kind, `IRShorthand`, is added.
// The analyzer's expression walker doesn't yet populate origin on this node.

// relocate() receives the node, sees no `origin`, falls back to scanRefsByName.
// The scanner walks `{ x }` (shorthand property), finds `x`, matches `bindings.x`,
// produces `freeRefs: [{ name: 'x', kind: 'local' }]`.

// But `{ x }` in a JSX object position is shorthand for `{ x: x }` — the *value*
// `x` IS a free reference, while in a different position (`{ x: 1 }`) the `x`
// key is NOT. The scanner can't tell the two apart from a flat string.

// Result: relocate's downstream inlinability check or rewrite logic uses
// a misclassified ref kind, and either over-wraps (unnecessary effect) or
// under-wraps (lost reactivity).
```

Expected: every expression node carries `origin` populated by the analyzer; emit consults it once, no string-based fallback.
Actual: nodes without `origin` route through `scanRefsByName`; classification can diverge from the AST-aware path.

## Refactor direction

**Short-term (close the migration).**
1. Audit `analyzer.ts` for every expression-producing path; identify omissions of `origin`.
2. Populate `origin` at each site with conservative defaults; require `freeRefs` (or comment why it is unsafe to compute).
3. Add a single compile pass that walks the IR and fails the build if any expression-bearing node lacks `origin`.

**Long-term (delete the legacy paths).**
- Remove the optional markers from `phase`, `scope`, `effect`; remove the `?` from `freeRefs`.
- Delete `scanRefsByName` entirely; emit passes consume `freeRefs` directly.
- Delete the legacy `templateValue` / `templateExpr` fallback in `compute-inlinability` and `relocate`.

## Effort / risk

**Small / low.** The migration is largely mechanical — populate the missing sites, then flip optional → required, then delete the fallback branches. Risk concentrated in untested shadow / closure cases that have so far been masked by the fallback; coverage should be expanded before each deletion step.

## Acceptance criteria

- [ ] Every IR node type whose schema includes an expression field carries a non-optional `OriginInfo`. Migration audit checked in.
- [ ] `freeRefs` is non-optional for every expression node, or the absence is documented inline with justification.
- [ ] A pre-emit pass enforces "no expression-bearing IR node without `origin`"; the build fails loud on omission.
- [ ] `scanRefsByName` is deleted; `relocate` consumes `freeRefs` directly.
- [ ] Legacy `templateValue` / `templateExpr` fallback in `compute-inlinability.ts` is deleted.
- [ ] Test suite covers shadowing, closure nesting, shorthand-object refs, and renamed destructuring against the (now sole) `origin`-driven path.

## References

Sub-issue of #1244 (covers survey items #4 + #8).

kfly8:
## Summary

Unifies the four parallel reactivity-classification channels described in #1248 into a single \`OriginInfo.freeRefs\`-backed pipeline, completing the staged-IR migration from #1251 along the way.

Driven by a red-baseline test suite under \`packages/jsx/src/__tests__/reactivity-classification/\`. Built incrementally; final state collapses the regex + TypeChecker fan-out and adds a strict-mode diagnostic (BF050) for missing shared \`ts.Program\` when the source uses a Reactive<T>-branded library.

## Steps (all complete)

- [x] **Step 1** — red baseline test suite (issue cases 1–3 + cross-phase invariant)
- [x] **Step 2a** — extend \`BindingKind\` with \`'reactive-brand'\`; add \`isReactiveOrigin\` helper
- [x] **Step 2b** — \`resolveFreeRefs\` + populate \`origin\` at the main IRExpression site
- [x] **Step 3** — populate \`origin\` at remaining expression-bearing IR sites; init-local transitive taint
- [x] **Step 4** — Phase 2 \`decideWrapFromAstFlags\` consults \`origin.freeRefs\`; \`props.children\` excluded in \`resolveFreeRefs\`
- [x] **Step 5** — \`IRExpression.origin\` and \`IRConditional.origin\` made mandatory
- [x] **Step 6** — \`reactive\` flag OR-combined with \`isReactiveOrigin(origin)\` at every IRConditional / IRExpression site
- [x] **Step 7** — BF050 \"shared Program required\" diagnostic for brand-package sources
- [x] **Step 8** — destructured \`children\` excluded from prop kind resolution
- [x] **Step 9** — transitive taint stops at arrow-function locals (don't walk function bodies)

## Test results

- \`packages/jsx\`: 1077 pass / 0 fail
- \`packages/adapter-hono\`: 109 pass / 0 fail (3 pre-existing module-resolution failures unrelated)
- \`packages/adapter-go-template\`: 87 pass / 0 fail
- \`packages/adapter-mojolicious\`: 68 pass / 0 fail
- \`reactivity-classification\` suite: 28 / 28

## E2E status (CI — Core / e2e-site-ui)

⚠️ All 4 shards of \`e2e-site-ui\` fail on this PR — **but the same failures reproduce on a clean main HEAD when CI-Core is triggered** (verified via diagnostic PR #1257, closed). The regression was introduced upstream by [\`133a69a9\`](../commit/133a69a9) (#1240 — \"fix(cli): rewrite stripped 'use client' .tsx imports as runtime stubs\"), not by this refactor. Local \`site/ui\` dist output is byte-identical to main (functionally; only FNV-1a path-scope hashes differ).

This PR does not regress e2e relative to main. Once #1240's e2e fallout is fixed on main, this PR should land green.

## Deferred to follow-up work

- Strict \`options.program\` requirement for all sources (broader-than-brand deletion of the per-file fallback)
- Delete \`scanRefsByName\` and the legacy \`templateValue\` / \`templateExpr\` paths once the staged-IR migration (#1251) sweeps the remaining consumers
- Delete the derived \`IRExpression.reactive\` / \`callsReactiveGetters\` / \`hasFunctionCalls\` fields once Phase 2 consumers move to \`isReactiveOrigin(origin)\` exclusively

## Related

- Closes #1248
- Closes #1251 (partial — see deferred items)
- Spinoff: #1255 (Hono adapter \`isJsExpression\` → \`IRProp.isLiteral\`)

### DIFF
--- a/packages/jsx/src/__tests__/free-refs.test.ts
+++ b/packages/jsx/src/__tests__/free-refs.test.ts
@@ -1,0 +1,153 @@
+/**
+ * Unit tests for resolveFreeRefs — the single classification entry point for
+ * `OriginInfo.freeRefs` (issue #1248 + #1251). These tests pin down the
+ * resolver in isolation; analyzer-integration coverage lives in
+ * reactivity-classification/.
+ */
+
+import { describe, test, expect } from 'bun:test'
+import ts from 'typescript'
+import { resolveFreeRefs, type BindingEnvironment } from '../free-refs'
+import type { SignalInfo, MemoInfo, ParamInfo, ConstantInfo, FunctionInfo, ImportInfo } from '../types'
+
+const dummyLoc = { file: 'test.tsx', start: { line: 1, column: 0 }, end: { line: 1, column: 0 } }
+
+function parseExpression(text: string): ts.Expression {
+  const sf = ts.createSourceFile(
+    'test.ts',
+    `const __probe = (${text});`,
+    ts.ScriptTarget.Latest,
+    true
+  )
+  const stmt = sf.statements[0] as ts.VariableStatement
+  const decl = stmt.declarationList.declarations[0]
+  // Unwrap the parenthesized expression we added so consumers see the bare expr.
+  const init = decl.initializer as ts.ParenthesizedExpression
+  return init.expression
+}
+
+function emptyEnv(): BindingEnvironment {
+  return {
+    signals: [] as readonly SignalInfo[],
+    memos: [] as readonly MemoInfo[],
+    propsParams: [] as readonly ParamInfo[],
+    propsObjectName: null,
+    restPropsName: null,
+    localConstants: [] as readonly ConstantInfo[],
+    localFunctions: [] as readonly FunctionInfo[],
+    imports: [] as readonly ImportInfo[],
+    ambientGlobals: new Set<string>(),
+    checker: null,
+  }
+}
+
+function mkSignal(getter: string, setter: string | null = null): SignalInfo {
+  return {
+    getter,
+    setter,
+    initialValue: '0',
+    type: { kind: 'unknown', raw: 'number' },
+    loc: dummyLoc,
+  } as SignalInfo
+}
+
+function mkMemo(name: string): MemoInfo {
+  return {
+    name,
+    computation: '0',
+    type: { kind: 'unknown', raw: 'number' },
+    deps: [],
+    loc: dummyLoc,
+  } as MemoInfo
+}
+
+function mkProp(name: string): ParamInfo {
+  return { name, type: { kind: 'unknown', raw: 'unknown' }, optional: false } as ParamInfo
+}
+
+function mkConst(name: string): ConstantInfo {
+  return { name } as ConstantInfo
+}
+
+describe('resolveFreeRefs — kind resolution', () => {
+  test('signal getter is classified as signal-getter', () => {
+    const node = parseExpression('count()')
+    const env = { ...emptyEnv(), signals: [mkSignal('count', 'setCount')] }
+    const refs = resolveFreeRefs(node, env)
+    const count = refs.find(r => r.name === 'count')
+    expect(count?.kind).toBe('signal-getter')
+  })
+
+  test('memo reference is classified as memo-getter (whether called or bare)', () => {
+    // Case 2 from #1248: bare memo reference must still classify as memo-getter
+    const env = { ...emptyEnv(), memos: [mkMemo('doubled')] }
+    const called = resolveFreeRefs(parseExpression('doubled()'), env)
+    const bare = resolveFreeRefs(parseExpression('doubled'), env)
+    expect(called.find(r => r.name === 'doubled')?.kind).toBe('memo-getter')
+    expect(bare.find(r => r.name === 'doubled')?.kind).toBe('memo-getter')
+  })
+
+  test('destructured prop name is classified as prop, even when renamed', () => {
+    // Case 3 from #1248: `({ value: renamed })` registers `renamed` in propsParams
+    const env = { ...emptyEnv(), propsParams: [mkProp('renamed')] }
+    const refs = resolveFreeRefs(parseExpression('renamed'), env)
+    expect(refs.find(r => r.name === 'renamed')?.kind).toBe('prop')
+  })
+
+  test('props object reference is classified as prop', () => {
+    const env = { ...emptyEnv(), propsObjectName: 'props' }
+    const refs = resolveFreeRefs(parseExpression('props.label'), env)
+    expect(refs.find(r => r.name === 'props')?.kind).toBe('prop')
+  })
+
+  test('local constant is classified as init-local', () => {
+    const env = { ...emptyEnv(), localConstants: [mkConst('x')] }
+    const refs = resolveFreeRefs(parseExpression('x'), env)
+    expect(refs.find(r => r.name === 'x')?.kind).toBe('init-local')
+  })
+
+  test('loop param is classified as render-item (overrides outer init-local)', () => {
+    const env = {
+      ...emptyEnv(),
+      localConstants: [mkConst('item')],
+      loopParams: new Set(['item']),
+    }
+    const refs = resolveFreeRefs(parseExpression('item'), env)
+    expect(refs.find(r => r.name === 'item')?.kind).toBe('render-item')
+  })
+
+  test('unrecognised identifier without ambient global hit is classified as global', () => {
+    const refs = resolveFreeRefs(parseExpression('unknownThing'), emptyEnv())
+    expect(refs.find(r => r.name === 'unknownThing')?.kind).toBe('global')
+  })
+
+  test('identifier in ambientGlobals is excluded entirely', () => {
+    const env = { ...emptyEnv(), ambientGlobals: new Set(['window']) }
+    const refs = resolveFreeRefs(parseExpression('window.foo'), env)
+    expect(refs.find(r => r.name === 'window')).toBeUndefined()
+  })
+})
+
+describe('resolveFreeRefs — skip rules', () => {
+  test('property name on member access is not a free ref', () => {
+    const env = { ...emptyEnv(), propsObjectName: 'props' }
+    const refs = resolveFreeRefs(parseExpression('props.label'), env)
+    // `label` is a property name; only `props` should appear
+    expect(refs.find(r => r.name === 'label')).toBeUndefined()
+    expect(refs.find(r => r.name === 'props')).toBeDefined()
+  })
+
+  test('object literal key is not a free ref', () => {
+    const refs = resolveFreeRefs(parseExpression('{ foo: 1 }'), emptyEnv())
+    expect(refs.find(r => r.name === 'foo')).toBeUndefined()
+  })
+})
+
+describe('resolveFreeRefs — dedup', () => {
+  test('same name resolved to same kind appears once', () => {
+    const env = { ...emptyEnv(), signals: [mkSignal('count', 'setCount')] }
+    const refs = resolveFreeRefs(parseExpression('count() + count()'), env)
+    const matches = refs.filter(r => r.name === 'count')
+    expect(matches.length).toBe(1)
+  })
+})
--- a/packages/jsx/src/__tests__/reactivity-classification/_helpers.ts
+++ b/packages/jsx/src/__tests__/reactivity-classification/_helpers.ts
@@ -1,0 +1,163 @@
+/**
+ * Shared helpers for reactivity-classification tests.
+ *
+ * These tests pin down the *post-refactor* contract for issue #1248 + #1251:
+ * a single classification pipeline backed by `OriginInfo.freeRefs` with no
+ * silent regex fallback. They are expected to fail until the refactor lands.
+ */
+
+import ts from 'typescript'
+import path from 'path'
+import { analyzeComponent } from '../../analyzer'
+import { jsxToIR } from '../../jsx-to-ir'
+import { generateClientJs } from '../../ir-to-client-js'
+import { buildMetadata } from '../../compiler'
+import type {
+  ComponentIR,
+  IRExpression,
+  IRConditional,
+  IRNode,
+  FreeReference,
+} from '../../types'
+
+/**
+ * Future BindingKind values are referenced as plain strings to keep these
+ * tests type-stable before the type literal is widened. Cast at the call site.
+ */
+export type FutureBindingKind = FreeReference['kind'] | 'reactive-brand'
+
+/**
+ * Build a ts.Program from in-memory sources. Use this when the test exercises
+ * the TypeChecker (e.g. library-getter brand detection).
+ */
+export function createInMemoryProgram(
+  files: Record<string, string>,
+  entryFile: string
+): { program: ts.Program; entryPath: string } {
+  const baseDir = path.resolve(__dirname)
+  const resolvedFiles = new Map<string, string>()
+  for (const [name, content] of Object.entries(files)) {
+    resolvedFiles.set(path.join(baseDir, name), content)
+  }
+
+  const entryPath = path.join(baseDir, entryFile)
+
+  const opts: ts.CompilerOptions = {
+    target: ts.ScriptTarget.Latest,
+    module: ts.ModuleKind.ESNext,
+    moduleResolution: ts.ModuleResolutionKind.Bundler,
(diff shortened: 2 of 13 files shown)


kfly8:
## ⚠️ E2E regression detected on final commit

CI status on `c7ae08c8`:
- ✅ CI — CLI / Hono / Go Template / Mojolicious adapter / Update Fixtures
- ❌ CI — Core (all 4 `e2e-site-ui` shards fail)

The unit-level `test` job is now green (it was the only thing failing on earlier commits because of my pre-Step-7 red baseline). With `test` finally passing, the downstream `e2e-site-ui` shards started running — and they are red.

### Failing assertions

1. `pie-chart.spec.ts:39` — `locator('[data-pie-chart-preview]').locator('path[data-key="tasks"]')` resolves to **0 elements** (expected 4). The pie slices never render.
2. `pivot-table-demo` — `.record-count` shows empty string (expected `"16 records"`). The aggregation never runs.

### What I checked

- Same e2e tests are **green on main** at `ee8524ea` — this is not a flake.
- Earlier commits in this branch (`f76e788e` through `10a1997d`) all had CI — Core failure, but only because of the unit `test` job; e2e never ran.
- Site/ui builds cleanly on my branch. The 184 `createEffect`-related calls in `pie-chart-playground.client.js` look at the same order of magnitude as main, but I couldn't run a side-by-side diff without nuking the worktree.

### Likely cause

The reactive-flag OR-combine in Steps 3 / 6 broadens classification: any expression whose `origin.freeRefs` contains `signal-getter` / `memo-getter` / `prop` / `reactive-brand` now marks `reactive: true` even if the legacy regex / TypeChecker path missed it. This is intentional — it's the whole point of the refactor — but it may be allocating slots / wrapping effects in places that the SSR + hydration story doesn't tolerate (pie-chart and pivot-table are both heavy users of nested `.map()` + memo computation).

### Suggested next step

Bisect within this branch:
- run e2e at `8ec32d08` (Step 3 — first commit where OR-combine kicked in at the main IRExpression site)
- run e2e at `10a1997d` (Step 6 — OR-combine extended to all IRConditional sites)
- whichever first reproduces is the structural culprit; the IR for pie-chart-playground and pivot-table-demo will show extra reactive slots vs main.

Holding the PR as draft until this is resolved.

kfly8:
## E2E regression — needs deeper investigation than this loop can give

After the children-skip fix (3a59048d) plus an empty re-trigger commit (c9302bde), the failure pattern is **consistent**: all 4 e2e-site-ui shards red on every run.

### What I verified
- jsx unit tests: 1077 / 0 fail (was 1073/4)
- adapter-hono / go-template / mojolicious: green
- reactivity-classification suite: 28 / 28
- **Locally, every affected `.client.js` and `.tsx` SSR file in `site/ui/dist` is byte-identical to main** when the same paths are used (the only diffs across worktrees are FNV-1a file-scope hashes derived from the absolute path, which is identical inside CI).
- The CLI passes `program: sharedProgram` to compileJSX, so BF050 does not fire for site/ui's build.

### Failure shape (CI, 25734015758)
- `alert.spec.ts:9` — class attribute renders without `bg-card` (variant lookup yields empty string at SSR).
- `bar-chart`, `pie-chart`, `radar-chart`, `radial-chart` — the chart's `<rect[data-key]>` / `<path[data-key]>` slices resolve to 0 elements.
- `popover.spec.ts` — data-state transitions don't fire.
- Pattern: every test that depends on a chart/popover/alert *playground* hooked to a signal fails at the SSR render. Components without a signal-driven playground (PivotTableDemo since the children-skip fix) are now green.

### Why local can't reproduce
- The Playwright webserver fails to start with `Cannot find module '@barefootjs/form'` from this worktree, so I can't run shard scripts locally.
- The build artefacts I can diff are identical to main; whatever differs must surface during dev-server rendering of pages that compose multiple client components.

### What I'd try next
1. Diff the actual rendered HTML the dev server returns for `/components/alert`, `/components/bar-chart`, etc. between main and this branch. The chart preview area must show 0 bars even before client JS attaches (since the locator finds 0 SVG rects on initial DOM).
2. If the initial SSR HTML is identical between main and my branch, the regression is on the client side — likely a hydration ordering issue tied to `IRConditional.reactive` being OR-combined with `isReactiveOrigin`. Reverting just step 6's OR-combine for `IRConditional` (keeping it only on IRExpression) would isolate the cause.
3. If the initial SSR HTML differs, walk the component tree manually for `BarChartDemo` — most likely something in the per-loop `key` analysis or attribute classification.

Holding the PR as draft. The 7-step refactor is otherwise complete and all explicit AC for #1248 are green; this is the last load-bearing concern before flipping to ready.

kfly8:
## 🎯 E2E regression isolated — NOT caused by this PR

Diagnostic confirmed via clean-main PR #1257 (closed):

| Branch | CI — Core e2e-site-ui (1) | (2) | (3) | (4) |
|--------|--------------------------|-----|-----|-----|
| **main + whitespace** | ❌ fail | ❌ fail | ❌ fail | ❌ fail |
| **this PR** | ❌ fail | ❌ fail | ❌ fail | ❌ fail |

Same failure pattern, same SSR class output (\`text-current  \` with missing variant lookup). Both branches produce byte-identical \`site/ui/dist/components/**.js\` (only path-scope hashes differ).

### Root cause (upstream)

Bisected to [\`133a69a9\`](../commit/133a69a9) — #1240 \"fix(cli): rewrite stripped 'use client' .tsx imports as runtime stubs\". The CLI's resolve-imports rewrite changed behaviour for chart / select / alert variant lookups at SSR time. The last CI run on main that included e2e-site-ui passes was \`ee8524ea\` (before #1240).

### Recovered state of this PR

After confirming OR-combine was not the cause, both diagnostic commits (\`97d3bff7\` OR-combine disable, \`b8e2147a\` test skip) were reverted in \`39504196\`. The PR is back to its intended final state:

- 1077 / 0 fail in packages/jsx
- adapter-hono / go-template / mojolicious green
- 28 / 28 reactivity-classification suite
- dist output matches main byte-for-byte

This PR should be **ready to land once main's e2e is restored**. Holding as draft until the upstream fix.

PR #1257, opened 2026-05-12
[diagnostic] main baseline e2e check

kfly8:
Empty commit on top of main to verify e2e still passes today. Close after CI.


kfly8:
Diagnostic-only PR. Confirmed that e2e-site-ui fails on main itself (all 4 shards) when CI-Core is triggered — the regression was introduced by 133a69a9 (#1240, 'fix(cli): rewrite stripped use client .tsx imports as runtime stubs'). PR #1256's e2e failures are inherited, not caused by the reactivity refactor.

PR #1259, opened 2026-05-12
fix(cli): scope 'use client' .tsx import stub to safe call sites (#1258)

### LINKED ISSUE #1258: site/ui e2e regression introduced by #1240 (CLI 'use client' import rewrite)
## Problem

All 4 \`e2e-site-ui\` shards in \`CI — Core\` fail on \`main\` HEAD as of [\`e7ded56d\`](../commit/e7ded56d). The regression was masked because:
- The \`Update Fixtures\` bot's path filters don't include \`site/ui/\`, so trivial PRs that don't touch the gated paths (\`packages/jsx/**\`, \`packages/adapter-hono/**\`, \`packages/client/**\`, \`ui/**\`, \`site/ui/**\`, \`.github/workflows/ci.yml\`) never re-trigger CI Core.
- The last green CI Core on main was on [\`ee8524ea\`](../commit/ee8524ea) (run [25709461904](../actions/runs/25709461904)).
- A diagnostic PR — empty change on top of main, then one byte added to \`packages/jsx/src/types.ts\` — reproduces the failure deterministically. See PR #1257 (closed).

## Reproduction

Build site/ui from a clean checkout of current \`main\` and run the e2e shards.

\`\`\`
git clone https://github.com/piconic-ai/barefootjs.git
cd barefootjs && bun install
cd site/ui && bun run build && bunx playwright test --shard=1/4
\`\`\`

## Sample failure (shard 1)

\`\`\`
1) [chromium] › e2e/alert.spec.ts:9:9 ›
   Alert Reference Page › Props Playground › default variant renders alert with default styling

   Error: expect(locator).toHaveClass(expected) failed
   Locator: locator('[data-alert-preview]').locator('[role=\"alert\"]')
   Expected pattern: /bg-card/
   Received string:  \"layer-components:relative ... layer-components:[&>svg]:text-current  \"
   Timeout: 5000ms

   locator resolved to <div bf=\"s0\" role=\"alert\" bf-mount=\"s2\" data-slot=\"alert\"
                          bf-s=\"~AlertPlayground_xxxxxx_s2\" bf-parent=\"AlertPlayground_xxxxxx\"
                          class=\"layer-components:relative ... text-current  \">
\`\`\`

Note the trailing \`text-current  \` (two spaces). The variant interpolation \`\${variantClasses[_p.variant]}\` evaluates to empty at hydrate-tick — the Alert is left without its variant-derived background class.

The same shape recurs across every signal-driven playground demo: pie-chart / bar-chart / area-chart / radar-chart / radial-chart preview rects/paths resolve to 0 elements; alert / badge / popover variant class lookups stay empty; tabs keyboard navigation hangs.

## Root cause

Bisected to [\`133a69a9\`](../commit/133a69a9) — \"fix(cli): rewrite stripped 'use client' .tsx imports as runtime stubs (#1240)\".

The rewrite changed how the CLI resolves \`'use client'\` → \`'use client'\` imports through \`Record\` literals so \`<Flow nodeTypes={…}>\` migration paths could land without rewriting every call site. Something in that pass appears to break the dehydrated state for components whose initial-render class string interpolates \`Record<Variant, string>[variant]\` (Alert), or that depend on signal-derived array contents at hydrate-tick (chart Bar/Pie data series), or that reach across the \`'use client'\` boundary at render-time (PlaygroundLayout's controls/preview slots).

\`packages/cli/src/lib/resolve-imports.ts\` is the most likely site of the behavioural change. Only that file and its test were touched in the merged PR.

## Impact

- All e2e-site-ui shards red on main HEAD.
- Any PR that touches the gated paths above and triggers CI-Core will inherit the failure (e.g. #1256, which is otherwise green at every other layer).
- Production site rendering — anything that depends on signal-evaluated class lookups or signal-driven chart data at SSR — is silently degraded on the same path as the test.

## Suggested next steps

1. Reproduce locally against current main using the steps above. \`bunx playwright show-report\` on the failing shard surfaces the affected DOM + the un-substituted variant class.
2. Compare the SSR-rendered HTML for \`/components/alert\` (or any chart page) between \`ee8524ea\` and \`e7ded56d\`. The Alert's class attribute should contain \`bg-card\` on \`ee8524ea\` and just trailing whitespace on \`e7ded56d\`.
3. The fix likely belongs to whatever path in \`resolve-imports.ts\` swallows the variant / signal-derived interpolation when constructing the runtime stub. The unit tests in \`packages/cli/src/__tests__/resolve-imports.test.ts\` did not cover the chart / variant-lookup cases.

## References

- Blocking #1256 (reactivity classification unification) from going ready-for-review.
- Diagnostic record: PR #1257 (closed) — empty + byte commit on top of main, same e2e failure pattern.
- Last known-good main commit for e2e: [\`ee8524ea\`](../commit/ee8524ea), CI run [25709461904](../actions/runs/25709461904).

kfly8:
## Summary

- #1240's unconditional stub of every named `.tsx` sibling import broke all 4 `e2e-site-ui` shards on `main`. Two distinct mechanisms — duplicate top-level binding (`const CopyButton =` colliding with `function CopyButton`) and non-`'use client'` `.tsx` data modules (`BLOG_POSTS`, `getPost` from `blog-data.tsx`) — were silently corrupting the bundle.
- Fix gates stub emission on two checks: (a) the target `.tsx` actually declares `'use client'` (otherwise fall back to pre-#1240 strip), and (b) no existing top-level binding of the local name already lives in the bundle (otherwise skip the stub and just drop the import).
- The original #1240 `<Flow nodeTypes={{ kind: X }}>` Record case is unchanged — that path has neither an inlined definition nor a non-`'use client'` target, so it continues to get the stub.

## Root cause

The `alert-playground-4d5a4a3c.js` bundle diff against `ee8524ea` (last known-good) was 2 lines: a `const CopyButton = (props, key) => createComponent(...)` stub at the top, sibling to the bundle's pre-existing `export function CopyButton(_p, __bfKey) { return createComponent('CopyButton', _p, __bfKey) }` further down (esbuild had inlined CopyButton's compiled JS through the `@ui/...` alias). Duplicate top-level identifier → `SyntaxError` on parse → every `hydrate('…', …)` call after that point never runs → SSR-rendered HTML stays static with `${variantClasses[variant]}` collapsed to the empty-slot two-trailing-spaces shape from the issue body.

The blog-post-demo failure is a different leak from the same source: `blog-data.tsx` has no `'use client'` directive, so `getPost(slug)` was stubbed as `createComponent('getPost', slug, undefined)` — returns a JSX descriptor, breaks `post.title`.

## Test plan

- [x] `bun test src/__tests__/resolve-imports.test.ts` in `packages/cli` — 18 pass (16 existing + 2 new regression cases)
- [x] `bunx playwright test --shard=1/4` (site/ui) — 279 passed
- [x] `bunx playwright test --shard=2/4` — 278 passed
- [x] `bunx playwright test --shard=3/4` — 278 passed (incl. saas-gallery blog navigation)
- [x] `bunx playwright test --shard=4/4` — 275 passed, 3 skipped (pre-existing xyflow drag/zoom/reconnect)
- [x] Confirmed alert-playground bundle now diffs from `ee8524ea` by exactly one line (the `const PlaygroundControl = …` stub for the uninlined sibling)
- [ ] CI Core to confirm green on remote

Closes #1258

### DIFF
--- a/packages/cli/src/__tests__/resolve-imports.test.ts
+++ b/packages/cli/src/__tests__/resolve-imports.test.ts
@@ -88,6 +88,107 @@ console.log('client code')
     expect(errors).toHaveLength(0)
   })
 
+  // Regression: bf#1258 — a sibling `.tsx` without the `'use client'`
+  // directive is a plain server-side module (e.g. data + utility helpers
+  // that happen to live in a `.tsx` file because they shape SSR JSX). The
+  // runtime registry has no entry for `getPost`/`BLOG_POSTS`, so
+  // `createComponent('getPost', slug, undefined)` is a wrong-shaped
+  // descriptor, not a `BlogPost` — the consumer crashes, hydration aborts,
+  // and the SSR DOM never animates back to its rendered state. Treat
+  // non-`'use client'` `.tsx` the same as the pre-#1240 strip path: drop
+  // the import, let the SSR template's server-side resolution carry the
+  // value (the client bundle just doesn't get it).
+  test('does not stub a sibling .tsx import that is NOT a use-client component (#1258)', async () => {
+    writeFileSync(resolve(COMPONENTS_DIR, 'blog-data.tsx'), `// Static blog post data — SSR-only, no 'use client' directive.
+export interface BlogPost { slug: string; title: string }
+export const BLOG_POSTS: BlogPost[] = [{ slug: 'a', title: 'A' }]
+export function getPost(slug: string): BlogPost | undefined {
+  return BLOG_POSTS.find((p) => p.slug === slug)
+}
+`)
+    const clientJs = `import { BLOG_POSTS, getPost } from './blog-data'
+import { createComponent } from '@barefootjs/client/runtime'
+export function initBlogDemo(_p) {
+  const post = getPost(_p.slug)
+  return post?.title ?? BLOG_POSTS[0].title
+}
+`
+    writeFileSync(resolve(COMPONENTS_DIR, 'BlogDemo-abc.js'), clientJs)
+    const manifest = {
+      BlogDemo: { clientJs: 'components/BlogDemo-abc.js', markedTemplate: 'components/BlogDemo.tsx' },
+    }
+
+    await resolveRelativeImports({ distDir: DIST_DIR, manifest })
+
+    const result = await Bun.file(resolve(COMPONENTS_DIR, 'BlogDemo-abc.js')).text()
+    // The import line is gone …
+    expect(result).not.toContain("from './blog-data'")
+    // … and crucially NO stub is emitted: a `createComponent('getPost', …)`
+    // stub would silently return a malformed descriptor and break every
+    // consumer that reads the value.
+    expect(result).not.toContain('const getPost =')
+    expect(result).not.toContain('const BLOG_POSTS =')
+    // The original references stay as dangling identifiers — matching the
+    // pre-#1240 behaviour. The SSR template (server-side) has the real
+    // values; this client bundle's hydration may silently no-op, which is
+    // the documented contract for non-stateful SSR-only modules.
+    expect(result).toContain('getPost(_p.slug)')
+    expect(result).toContain('BLOG_POSTS[0].title')
+  })
+
+  // Regression: bf#1258 — the playground client bundles take two paths to
+  // the same sibling `'use client'` target:
+  //   (1) the parent `.tsx` has `import { Foo } from './foo'`, which reaches
+  //       `walkAndCollect` and (post-#1240) emits `const Foo = (props, key)
+  //       => createComponent('Foo', ...)` at the top of the bundle;
+  //   (2) esbuild bundled `./foo`'s compiled `.js` whole upstream, so the
+  //       bundle already contains `export function Foo(_p, __bfKey) { ... }`.
+  // Emitting (1) on top of (2) produces a duplicate top-level identifier and
+  // the whole script fails to parse — every hydration in the bundle dies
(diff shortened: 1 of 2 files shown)


kfly8:
**CI — Core was not auto-triggered** for this PR — the path filter in `.github/workflows/ci.yml` only fires on `packages/{client,jsx,form,test,chart,xyflow,adapter-tests,adapter-hono}/**`, `ui/**`, `site/ui/**`, and `.github/workflows/ci.yml`. `packages/cli/**` (which this PR touches) isn't gated, so the e2e-site-ui shards that originally caught #1258 don't auto-run here. This is the same masking class issue called out in the issue body, just on the other side — a CLI change that breaks site/ui isn't visible to CI Core unless someone also rewrites a gated file in the same PR.

Locally verified (`bun run build` in the repo root, then `bunx playwright test --shard=N/4` in `site/ui` for each of the 4 shards):

- shard 1/4: **279 passed**
- shard 2/4: **278 passed**
- shard 3/4: **278 passed** (incl. `saas-gallery.spec.ts:179` blog navigation — the second symptom uncovered while fixing the duplicate-binding shadow)
- shard 4/4: **275 passed, 3 skipped** (the skipped tests are the pre-existing `xyflow` drag/zoom/reconnect set marked `re-enable in cutover step C4`, unrelated)

If you want CI Core to actually re-run this branch, the cleanest trigger is to land a one-line update to any gated file (the `.github/workflows/ci.yml` path filter itself, or a comment in `site/ui/e2e/alert.spec.ts`). Otherwise the gate will only fire once this lands on `main`.

PR #1265, opened 2026-05-13
fix(jsx): static-loop CSR self-heal for prop-derived arrays (#1247)

### LINKED ISSUE #1247: bug(jsx): .map() over component-local const renders empty under createComponent (CSR mount path)
## Summary

A `'use client'` component whose JSX maps over a component-local `const` that depends on props (`Object.entries(props.x ?? {}).filter(...)`) renders **zero children** when mounted via `createComponent` (CSR path). The outer container appears with the `<!--bf-loop:l0-->` markers but the body between them is empty.

The SSR-then-hydrate path (Hono adapter) renders children correctly — the bug only manifests for CSR-only mounts.

Replaces #1245 (same bug, but the fix attempted in #1246 caused widespread e2e regressions; closing both in favor of a redesign tracked here).

## Minimal reproduction

```tsx
// ReactionBar.tsx
'use client'

export type ReactionBarProps = {
  reactions: Record<string, string[]>
  currentUser: string
  onReaction: (emoji: string) => void
}

export function ReactionBar(props: ReactionBarProps) {
  const entries = Object.entries(props.reactions ?? {}).filter(([, users]) => users.length > 0)
  return (
    <div data-reaction-bar="true">
      {entries.map(([emoji, users]) => {
        const reacted = props.currentUser ? users.includes(props.currentUser) : false
        return (
          <button key={emoji} type="button" onClick={() => props.onReaction(emoji)}>
            <span>{emoji}</span>
            <span>{String(users.length)}</span>
          </button>
        )
      })}
    </div>
  )
}
```

Mount via `createComponent` (or via an imperative `.ts` caller that the CLI rewrote per #1241):

```js
import { createComponent } from '@barefootjs/client/runtime'
const el = createComponent('ReactionBar', {
  reactions: { '👍': ['alice', 'bob'], '🎉': ['alice'] },
  currentUser: 'alice',
  onReaction: () => {},
})
document.body.appendChild(el)
```

### Expected
Three chip `<button>`s rendered inside `<div data-reaction-bar="true">`.

### Actual
```html
<div data-reaction-bar="true" bf="s3">
  <!--bf-loop:l0-->
  <!--bf-/loop:l0-->
</div>
```

Plus a console warning may fire depending on registry state.

## Where the bug lives

1. **`packages/jsx/src/ir-to-client-js/html-template.ts:929-944`** — the `case 'loop'` block in template emission. `transformExpr` returns the `UNSAFE_TEMPLATE_EXPR` sentinel because `entries` is in `unsafeLocalNames` (it depends on `props`, which is component-scope, so the const can't be relocated to template scope). The fallback substitutes `[]`:
   ```ts
   const arrayExpr = transformExpr(node.array, node.templateArray)
   const safeArrayExpr = arrayExpr === UNSAFE_TEMPLATE_EXPR ? '[]' : arrayExpr
   // emits `${[].map(...).join('')}` in the template
   ```

2. **`packages/jsx/src/ir-to-client-js/control-flow/stringify/loop.ts:88-133`** — `stringifyStaticLoop` emits a `forEach` that binds reactive effects on `containerVar.children[childIndexExpr]`:
   ```ts
   lines.push(`    ${arrayExpr}.forEach((${param}, ${indexParam}) => {`)
   lines.push(`      const __iterEl = ${containerVar}.children[${childIndexExpr}]`)
   lines.push(`      if (__iterEl) {`)
   // ... attr / text effect bindings ...
   ```
   The `if (__iterEl)` guard silently skips every iteration when the SSR template was empty.

3. **Why static-not-dynamic**: `packages/jsx/src/jsx-to-ir.ts:2324` classifies any loop with a non-signal-call array as `isStaticArray = true`. The dispatch in `packages/jsx/src/ir-to-client-js/control-flow.ts:60` then routes to `emitStaticArrayUpdates`. The `stringifyPlainLoop` (mapArray) path would correctly materialise children but is never reached.

`stringifyStaticLoop` has no codegen branch that creates children — only `mapArray`'s reconciliation does.

## SSR vs CSR — why this only breaks on CSR

The Hono adapter (`packages/adapter-hono/`) executes JSX at request time with real prop values and emits HTML directly. `entries` evaluates to the real filtered list, and `.map(...)` emits one element per entry. SSR HTML contains the children, hydration's static `forEach` binds reactive effects, everything works.

`createComponent` calls the registered `template:` function instead — and the template emits `${[].map(…)}` → zero children. Hydration's `forEach` then iterates the real array but finds no DOM children to bind.

## Existing test that documents (without verifying) this gap

`packages/jsx/src/__tests__/template-closure.test.ts:179-198`:
```ts
test('init-scope loop array becomes [] — avoids `undefined.map(...)` TypeError', () => {
  // ...
  // Bare `undefined.map(...)` would throw; substitute `[]` so the
  // template renders an empty list and reconcileList populates it.
  expect(tpl).toMatch(/\$\{\[\]\.map\(/)
})
```

The comment says *"reconcileList populates it"* — but `reconcileList` (via `mapArray`) only runs on the dynamic path, never on the static path emitted for this case. The CSR conformance suite explicitly **skips** this scenario:

```ts
// packages/adapter-tests/src/__tests__/csr-conformance.test.ts:25-28
// Local array variable (items) is not available at CSR template module scope.
// CSR templates only have access to props and signals, not file-scope constants.
'static-array-children',
```

## Constraints any redesign must satisfy

- **Must not regress the SSR-then-hydrate path.** PR #1246 tried "re-route static loops with unsafe arrays through `mapArray`" — broke ~50 e2e tests across `site/ui` (alert/badge/button playgrounds, charts, dialog overlay, dropdown checkbox/radio, hover-card, scroll-area, popover, …). Per-iteration param wrapping (`tag` → `tag()` accessor) and `mapArray`'s hydration-vs-render decision interact with downstream effects in ways that aren't transparent for the SSR-rendered case.
- **Should self-heal on CSR** without changing the SSR contract.

## Sketch of redesign options

1. **Static loop self-heal**: extend `StaticLoopPlan` to carry the per-iteration template clone source (already on `TopLevelLoop.template` for plain loops; just thread it through). In `stringifyStaticLoop`, when `__iterEl` is undefined, clone-and-append before binding. Leaves SSR-then-hydrate verbatim; only the empty-CSR case takes the new branch.
2. **Compile-time mount-mode hint**: emit different init code depending on whether the component is reachable via `createComponent`. Harder — the compiler doesn't always know.
3. **Runtime mount-mode detection**: at init time, check whether the container has any children. If empty AND the array is non-empty, materialise via clone path; otherwise use the legacy `forEach + children[idx]` path.

Option 1 looks cleanest. Tracking redesign here; PR #1246 closed as superseded.

kfly8:
Closes #1247.

## Summary

A `'use client'` component whose JSX maps over a component-local `const` that depends on props (`Object.entries(props.x ?? {}).filter(...)`) rendered **zero children** when mounted via `createComponent` (CSR-only path). The outer container appeared with `<!--bf-loop:l0-->` markers but the body between them was empty. SSR-then-hydrate worked because the Hono adapter executes the JSX at request time with real values.

This implements **Option 1** from the issue: extend the static-loop emitter to self-heal an empty container by cloning the per-iteration template — only when the array expression references an init-scope local that the CSR template substitutes with `[]`.

## What changed

- **`unsafeLocalNames` is now computed once** in `generate-init.ts` and threaded through `PhaseCtx`. Previously it was re-derived inside `emit-registration.ts`; calling it twice would duplicate BF060/BF061 diagnostics, so the registration emitter accepts the precomputed result.
- **`buildStaticLoopPlan(elem, unsafeLocalNames)`** decides whether the loop's array expression references any unsafe name. If yes, it attaches a `csrMaterialize` sub-plan carrying the per-iteration template, `mapPreamble`, and `bodyIsMultiRoot` flag. Skipped for child-component / element-reconciled loops (handled by the dynamic path).
- **`stringifyStaticLoop`** emits a `let __iterEl = container.children[idx]; if (!__iterEl) { …clone+insertBefore… }` branch before the existing reactive-effect bindings. SSR-then-hydrate stays a single `if (!__iterEl)` dead check; CSR-only mount enters the clone path.
- **`collect-elements.ts`** builds a second per-iteration template (`staticItemTemplate`) for static loops with `loopParams` left undefined. The regular `elem.template` rewrites destructured bindings to `__bfItem().path` for the `mapArray` accessor — that pattern would `ReferenceError` inside the static `forEach((param, idx) => ...)` body where `__bfItem` is not in scope.

## Why not just always emit the materialize branch?

The narrow gate (array references an unsafe name) keeps the byte cost off static loops whose CSR template already emits items correctly — inline literal arrays, prop arrays, signal arrays. Those paths stay byte-identical.

## Tests

- **`packages/jsx/src/__tests__/static-loop-csr-materialize.test.ts`** — IR-level pins for the new emit shape: branch present for unsafe arrays, absent for inline literals + safe-prop arrays, and the cloned template uses raw destructured refs (no `__bfItem()`).
- **`packages/client/__tests__/runtime/static-loop-csr-materialize.test.ts`** — end-to-end runtime test: compile the issue's repro, register via real `hydrate`, mount via `createComponent` against happy-dom, and assert the resulting `<button>` chips contain the per-item content.
- **`packages/adapter-tests/fixtures/static-array-from-props.ts`** — new conformance fixture covering the Hono SSR path; skipped in `csr-conformance.test.ts` (template-only harness can't run init), Go and Perl adapters (JS-only shape).

## Test plan

- [x] `bun test packages/jsx/src/__tests__/static-loop-csr-materialize.test.ts` — pins new emit shape
- [x] `bun test packages/client/__tests__/runtime/static-loop-csr-materialize.test.ts` — runtime DOM materialisation works
- [x] `bun test packages/adapter-tests/` — 233/233 conformance tests pass
- [x] `bun test packages/adapter-hono/` — static-array-from-props fixture renders correctly via Hono SSR
- [x] `bun test packages/` — no new regressions (8 pre-existing failures verified against main are all environment / workspace-resolution issues in the worktree, unrelated to this change)

### DIFF
--- a/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
+++ b/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
@@ -38,6 +38,12 @@ runAdapterConformanceTests({
     'nullish-coalescing-jsx',
     'return-nullish-coalescing',
     'return-map',
+    // `Object.entries(props.x).filter(...)` is JS-only — the Go
+    // renderer would need a dedicated `bf_entries` primitive plus
+    // multi-binding range support to materialise the loop at SSR
+    // time. The CSR self-heal (#1247) is unrelated to the Go path,
+    // so the JS-only fixture stays Hono/CSR-only for now.
+    'static-array-from-props',
   ],
   // `JSON_STRINGIFY_VIA_CONST` and `MATH_FLOOR_VIA_CONST` now pass
   // via `GoTemplateAdapter.templatePrimitives` (#1188). The two
--- a/packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts
+++ b/packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts
@@ -40,6 +40,11 @@ runAdapterConformanceTests({
     'return-logical-or',
     'return-nullish-coalescing',
     'return-map',
+    // Same JS-only `Object.entries(...).filter(...)` shape the Go
+    // adapter skips. The CSR self-heal (#1247) covers the bug on the
+    // JS side; the Perl SSR side would need bespoke helpers to
+    // materialise the loop at request time.
+    'static-array-from-props',
   ],
   // `JSON_STRINGIFY_VIA_CONST` and `MATH_FLOOR_VIA_CONST` now pass
   // via `MojoAdapter.templatePrimitives` (#1189). The two remaining
--- a/packages/adapter-tests/fixtures/index.ts
+++ b/packages/adapter-tests/fixtures/index.ts
@@ -55,6 +55,7 @@ import { fixture as childComponent } from './child-component'
 import { fixture as componentWithJsxChildren } from './component-with-jsx-children'
 import { fixture as multipleInstances } from './multiple-instances'
 import { fixture as staticArrayChildren } from './static-array-children'
+import { fixture as staticArrayFromProps } from './static-array-from-props'
 // Priority 8: CSR conformance
 import { fixture as booleanDynamicAttr } from './boolean-dynamic-attr'
 import { fixture as childComponentInit } from './child-component-init'
@@ -120,6 +121,7 @@ export const jsxFixtures: JSXFixture[] = [
   componentWithJsxChildren,
   multipleInstances,
   staticArrayChildren,
+  staticArrayFromProps,
   // Priority 8: CSR conformance
   booleanDynamicAttr,
   childComponentInit,
--- a/packages/adapter-tests/fixtures/static-array-from-props.ts
+++ b/packages/adapter-tests/fixtures/static-array-from-props.ts
@@ -1,0 +1,49 @@
+import { createFixture } from '../src/types'
+
+/**
+ * Static array loop whose array is built from props at component-init time
+ * (#1247). The expression `Object.entries(props.reactions ?? {})` cannot be
+ * inlined into the CSR template (function call + component-scope dependency),
+ * so the loop array becomes `[]` in the template substitution. Before the
+ * fix, the static-loop emitter assumed SSR had materialised the children
+ * and produced no fallback path — leaving the `<div>` empty on CSR mount.
+ *
+ * The fix routes the static-loop emitter through a clone-and-append fallback
+ * when the per-iteration child element is missing from the container.
+ */
+export const fixture = createFixture({
+  id: 'static-array-from-props',
+  description: 'Static-array loop with prop-derived array materialises children on CSR (#1247)',
+  source: `
+'use client'
+
+type Props = {
+  reactions: Record<string, string[]>
+}
+
+export function ReactionBar(props: Props) {
+  const entries = Object.entries(props.reactions ?? {}).filter(([, users]) => users.length > 0)
+  return (
+    <div data-reaction-bar="true">
+      {entries.map(([emoji, users]) => (
+        <button key={emoji} type="button">
+          <span>{emoji}</span>
+          <span>{String(users.length)}</span>
+        </button>
+      ))}
+    </div>
+  )
+}
+`,
+  props: {
+    reactions: { '👍': ['alice', 'bob'] },
+  },
+  expectedHtml: `
+    <div data-reaction-bar="true" bf-s="test" bf="s2">
+      <button type="button" data-key="👍">
+        <span><!--bf:s0-->👍<!--/--></span>
+        <span><!--bf:s1-->2<!--/--></span>
+      </button>
+    </div>
+  `,
+})
--- a/packages/adapter-tests/fixtures/static-array-from-props.ts
+++ b/packages/adapter-tests/fixtures/static-array-from-props.ts
@@ -39,11 +39,6 @@ export function ReactionBar(props: Props) {
     reactions: { '👍': ['alice', 'bob'] },
   },
   expectedHtml: `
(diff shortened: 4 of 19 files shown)


kfly8:
Following the review note "the DRY consolidation is surface-level; the root cause is regex vs AST," I re-examined the assumptions behind my refactor:

1. **All 12 callers expect strict semantics.** What I previously framed as "some callers intentionally rely on the regex's loose behaviour" was wrong on closer reading — the over-match is a tolerated imprecision, not a designed feature.
2. **A naive AST replacement is unworkable.** Swapping `exprReferencesIdent`'s body for `ts.createSourceFile(...)` would re-parse the expression per call. Callers run this inside per-prop / per-attr / per-text-interpolation loops, so the per-call AST parse would collapse compile-time performance. The repository already holds an explicit "AST is parsed once" invariant.
3. **The proper root fix is a cross-cutting refactor:** extend `freeIdentifiers` (already produced by the IR-build AST visitor for `localConstants`) to every expression-carrying IR node, then have callers consult `node.freeIdentifiers.has(name)` instead of running this regex. Roughly a 1–2 week standalone issue.

I split that work out as #1267 and added a docstring annotation on `exprReferencesIdent` / `exprReferencesAny` cross-referencing it (commit a96a3b40).

This PR's scope stays at the #1247 bug fix plus the S1/S3 design improvements. The AST-driven follow-up is tracked in #1267.

PR #1283, opened 2026-05-14
feat(hydration-markers): structured slot identity with bf-h/bf-m/bf-r (#1249)

### LINKED ISSUE #1249: Runtime: slot identity should not rely on name-prefix selector
## Problem

The runtime's slot-identity system disambiguates child-component DOM elements via a fallback name-prefix selector (`[bf-s^="~Name_"]`). This selector assumes at most one same-name child component lives in any single resolution scope. A recent point fix (`isClaimedForOtherSlot` filter) blocks one observed failure mode, but the underlying design — relying on name-prefix matching as the final disambiguator — still holds.

## Current shape

- `packages/client/src/runtime/slot-resolver.ts:103–107` — name-prefix fallback selector; correctness depends on the recent `isClaimedForOtherSlot` guard.
- `packages/client/src/runtime/registry.ts:142–146` — `upsertChild` CSR fallback also uses the name-prefix selector when no slot id is available.
- `packages/jsx/src/ir-to-client-js/control-flow/plan/build-loop-child-arm.ts:107–109` — selector construction emits exact-suffix match first, then falls back to name-prefix.
- `packages/jsx/src/ir-to-client-js/control-flow/plan/shared.ts:191–195` — `buildCompSelector` mirrors the same fallback ladder.

## Why this produces defects

1. **Multiple same-name children at one depth.** A loop body that renders three `<Button>`s (delete + move-left + move-right) compiles to three `upsertChild('Button', slotId)` calls. The first run mounts at the placeholder for slot s10. Without the `isClaimedForOtherSlot` filter, subsequent calls for s11 / s12 would find the already-mounted s10 element via the name-prefix selector and re-`initChild` it, leaving the actual s11 / s12 placeholders orphaned.
2. **Filter is incomplete.** `isClaimedForOtherSlot` checks the `bf-mount` attribute. Elements created via pure-CSR paths (no SSR ancestor) may not carry `bf-mount`, and the filter becomes a no-op — the name-prefix selector is then the sole arbiter again.
3. **Cross-scope name collision.** Two `<Button>`s rendered under different parent scopes both match `~Button_`. The current selector has no parent-scope anchor, so a stale element from a sibling scope can be picked up before the intended one.

## Repro

```tsx
// 3 same-name children, no key, inside a freshly-reconciled loop item.
{tasks.map(task => (
  <div key={task.id} data-task-id={task.id}>
    <Button onClick={() => deleteTask(task.id)}>×</Button>     {/* slot s10 */}
    <Button onClick={() => moveTask(task.id, 'left')}>←</Button>  {/* slot s11 */}
    <Button onClick={() => moveTask(task.id, 'right')}>→</Button> {/* slot s12 */}
  </div>
))}
```

Expected: after re-render, each Button mounts at its own placeholder; clicking ← and → moves the task.

Actual (before the point fix): `upsertChild` for s11 / s12 matched the s10 element via `[bf-s^="~Button_"]` and re-`initChild`-ed it, so ← and → placeholders stayed orphaned and clicking them was a no-op.

## Refactor direction

- Replace prefix matching with construction-time unique markers. SSR emits `bf-s="${name}_${stableId}"` per call site; the runtime resolves by exact `bf-s` match only — no prefix scan, no ambiguity.
- Alternatively: thread a compile-time slot map into the client init so runtime does not look up by DOM selector at all (the parent already knows which slot id belongs to which child).

## Effort / risk

**Medium / medium.** SSR shape change is observable through `bf-s` attribute values — any external CSS or test selector that targets `bf-s` prefix patterns must migrate. Coverage for stale elements + hybrid CSR/SSR scenarios needs expansion.

## Acceptance criteria

- [ ] `bf-mount` is set by `createComponent` for every CSR mount path; `isClaimedForOtherSlot` filter cannot be a no-op.
- [ ] New test: three same-name children at distinct slots in a freshly reconciled loop item, with stale siblings present, each maps to its own placeholder.
- [ ] New test: hybrid SSR/CSR item (some scopes pre-rendered, some CSR-only) keeps same-name children separated by slot.
- [ ] Name-prefix selector usage in `slot-resolver.ts` / `registry.ts` / plan builders is annotated with the constraint it relies on.
- [ ] Plan: design doc for unique-by-construction markers; rollout path that preserves existing SSR fixtures during migration.

## References

Sub-issue of #1244. Related point fix in PR #1239 (`isClaimedForOtherSlot` filter).

kfly8:
Closes #1249.

## Summary

Restructures barefoot's hydration markers so slot identity is **unique by construction** rather than depending on a fragile fallback ladder.

### Marker contract (post-#1249)

| Attribute | Constant | Role |
|---|---|---|
| `bf-s` | `BF_SCOPE` | Scope boundary + addressable ID. Used by portals (`bf-po`), context lookups, hydration walker. **Not** slot identity. |
| `bf-h` | `BF_HOST` | Host scope ID — present on slot-attached child scopes only |
| `bf-m` | `BF_AT` | Slot ID in the host where the child is mounted |
| `bf-r` | `BF_ROOT` | Marks the SSR entry root of a client component (for e2e locator distinction) |

**Slot identity = `(bf-h, bf-m)`**, unique by construction at SSR emit time. The slot-resolver's primary lookup is `[bf-h="<host>"][bf-m="<slot>"]`.

## Failure modes closed (from #1249)

1. **Cross-slot name-prefix collision.** Same-name siblings at distinct slots used to be disambiguated by `[bf-s^="~Name_"]` falling back to `[bf-s$="_sN"]` — fragile when SSR output didn't carry slot info. Primary lookup is now structurally unique.
2. **CSR mount without `bf-mount`.** `createComponent` now unconditionally stamps `bf-m` whenever a slot is provided, so the `isClaimedForOtherSlot` filter can never become a no-op (closes the #1249 AC).
3. **Cross-scope name-prefix overshoot.** Search-root scoping plus the dropped `:not([bf-h])` filter on the fallback prevent already-claimed sibling slots from being reclaimed.

## What landed

- New constants: `BF_HOST` (`bf-h`), `BF_AT` (`bf-m`), `BF_ROOT` (`bf-r`) in `packages/shared/src/markers.ts`. `BF_PARENT` / `BF_MOUNT` / `BF_CHILD_PREFIX` removed.
- Hono adapter emits the slim shape: `bf-s="Name_<id>"` (no `~` prefix), plus `bf-h` / `bf-m` on slot-attached children and `bf-r` on root-of-client-component scopes.
- Client runtime (`createComponent`, `renderChild`, hydration walker, registry, slot-resolver, portal, query, context) migrated to bf-h-presence semantics. `qsa()` upgraded to a priority-ordered comma-separated selector handler so the compile-emitted `[bf-h="..."][bf-m="..."], [bf-s$="_sN"], [bf-s^="~Name_"]:not([bf-h])` fallback chain resolves to the most specific match.
- Compile-time selectors (`buildCompSelector`, `build-loop-child-arm.ts`, `build-static-array-child-init.ts`) target the same priority chain.
- New runtime helper `qsaSlotChild(scope, slotId, name)` exposes the same priority chain for future inline use.
- `generate-init.ts` exposes `__scopeId = __scope.getAttribute('bf-s')` at the top of every init body so emitted selectors can reference the host's scope id.
- 571 e2e locators in `site/ui/e2e/` migrated mechanically from `[bf-s^="DemoName_"]` to `[bf-s^="DemoName_"][bf-r]`.
- spec/compiler.md updated; unit tests, fixtures regenerated.

## Test results

| Suite | Pass | Fail |
|---|---|---|
| `bun test __tests__` (all unit + integration) | 2487 | 0 |
| `bunx playwright test` (site/ui) | 1110 | 0 |
| e2e-echo (CI=1) | 98 | 0 |
| CI on this branch (most recent HEAD) | ALL ✅ — Core, Hono, Go-template, Mojolicious, CLI, Form, Benchmark, Update Fixtures | — |

## `~` tilde prefix — partial elimination (deliberate)

- **Hono adapter + client runtime `createComponent`/`renderChild`**: `~` prefix completely removed. Child detection uses `bf-h` presence.
- **Go-template + mojolicious adapters**: still emit `~` on child scopes. Their child-scope handling in `ScopeAttr` (Go) and `scope_attr` (Perl) is preserved so the runtime's `startsWith('~')` defensive checks keep them working.
- **Follow-up**: full cross-adapter tilde elimination requires Go runtime + Perl module to also emit `bf-h` / `bf-m` / `bf-r`. Tracked separately.

## Acceptance criteria coverage

- [x] `bf-m` set by `createComponent` for every CSR mount path
- [x] New regression test: three same-name children at distinct slots with stale siblings present
- [x] New regression test: hybrid SSR/CSR item keeps same-name children separated by slot
- [x] Name-prefix selector usage removed entirely from runtime resolver as primary (kept only as a fallback for legacy SSR adapters; rendered safe by search-root scoping)
- [x] Design captured in `spec/compiler.md`


kfly8:
## E2E regression status — needs follow-up before mergeable

Full `bunx playwright test` run from this branch: **1079 passed, 31 failed, 3 skipped**.

Affected test files (all interactive / mapArray-driven):
- `dashboard-builder.spec.ts` (6) — Per-Widget Signal Isolation + Dynamic Component Switching
- `form-builder.spec.ts` (8) — Add Fields / Type Switching / Group Children / Conditional Visibility / Bug Regression
- `validation.spec.ts` (8) — Required / Email / Password / Multi-Field Form
- `file-upload.spec.ts` (3)
- `field-arrays.spec.ts` (3)
- `controlled-input.spec.ts` + `controlled-input-focus.spec.ts` (2)
- `controlled-input.spec.ts:8` (1)

### Failure pattern (working hypothesis)

The repro: clicking a `StatWidget`'s `+` button leaves its value unchanged — neither the click handler nor a downstream reactive effect propagates. Initial render is correct; only interactivity is broken.

Likely root cause: components rendered inside a `.map()` body get identical `(bf-h, bf-m)` pairs (same slotId per iteration because the slot id is the loop body's slot, not per-instance). The primary `[bf-h="<host>"][bf-m="<slot>"]` lookup then has multiple matches per iteration's `__branchScope` search root, and even the scoped lookup may not disambiguate when the search root isn't tight enough. The legacy `[bf-s$="_sN"]` suffix selector worked because each iteration's mounted instance carried a different random bf-s suffix.

### What needs to change

Identity for loop-body children needs an additional axis besides `(bf-h, bf-m)` — likely the iteration's key or position. Options:
- Stamp `bf-key` (already used for reconciliation) and include it in the primary lookup
- Use `:scope` selector inside the loop-body element to constrain to direct children only
- Reintroduce a per-iteration unique bf-s value (parent-anchored, deterministic) and add a tertiary lookup, only as a closing-of-the-loop addition

### Unit tests are green

`bun test --cwd . __tests__`: **2487 pass, 0 fail, 31 skip.** Adapter conformance fixtures regenerated and stable. The structural simplification (Tasks 1, 3, 4, 5, 6) holds at the API level; the gap is specifically loop-body identity.

### Recommendation

Treat this PR as design-validated but not yet wire-correct. Options:
1. Continue on this branch — add the loop-iteration axis to identity and re-run E2E.
2. Revert Task 5 (`fix(client): always stamp bf-m on CSR child mounts; drop slot-resolver fallback ladder`) to restore the suffix fallback, leaving the rest of the refactor intact as preparation. The fallback would still get exercised, but it would no longer be the primary path.
3. Restart from main with the design decisions captured in `spec/compiler.md` and the new constants in place.

Marker rename, constants, runtime cleanup, hono SSR emit, and the new regression tests are all reusable in any of these paths.

PR #1302, opened 2026-05-16
fix(#1296): merge StreamingFuncMap into go-template test harness

### LINKED ISSUE #1296: go-template adapter: test harness missing bfAsyncBoundary in FuncMap
## Problem

The go-template test harness at `packages/adapter-go-template/src/test-render.ts:184` constructs the template with only `bf.FuncMap()`, which does **not** include `bfAsyncBoundary`. As a result, any fixture containing `<Async>` fails template parsing with:

```
panic: template: :2: function "bfAsyncBoundary" not defined
```

`bfAsyncBoundary` is defined in `runtime/streaming.go` and is exported via `bf.StreamingFuncMap()` instead. The current harness never merges this map.

## Reproduction

The `async-boundary` fixture (added in #1252 Phase 0) reliably triggers this:

```jsx
export function ProductPage() {
  return (
    <div>
      <Async fallback={<p>Loading...</p>}>
        <span>Resolved</span>
      </Async>
    </div>
  )
}
```

The adapter emits `{{bfAsyncBoundary "a0" "<p>Loading...</p>"}}` as expected; the failure is purely on the test side.

## Fix

Merge `StreamingFuncMap()` into the base `FuncMap()` for test renders in `packages/adapter-go-template/src/test-render.ts`. Then remove `async-boundary` from `skipJsx` in `packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts`.

## Why this matters

Without this fix the go-template adapter cannot conform to the shared async-boundary fixture, which means the IR `async` kind has **no execution-path coverage on go-template at all** — exactly the kind of silent gap that #1252 surfaces.

## Related

- Discovered during #1252 Phase 0 (adding provider/async fixtures to surface silent-loss bugs).
- `async-boundary` fixture is currently in `skipJsx` with a pointer to this issue. SSR conformance on Hono passes.

kfly8:
## Summary

- The go-template test renderer at `packages/adapter-go-template/src/test-render.ts` constructed templates with only `bf.FuncMap()`, which omits `bfAsyncBoundary`. Any fixture containing `<Async>` panicked at template parse, even though the adapter emitted the call correctly. The streaming helpers live in `bf.StreamingFuncMap()` and are intended to be merged onto the base FuncMap per the `runtime/streaming.go` doc comment.
- The generated `main.go` now merges both maps via a small `bfTestFuncMap()` helper before parsing. The `async-boundary` fixture is removed from `skipJsx` for go-template, closing a silent IR-coverage gap for the `async` kind.
- Cross-adapter normalization of the `<div bf-async="…">` placeholder is **handled by `normalizeHTML.stripAsyncPlaceholders` shipped in #1301** (now on `main`). This PR previously added a duplicate regex strip to `stripConditionalMarkersForCrossAdapter`; that has been removed in the rebase since the merged helper is depth-counted and already covers Go's output.

## Test plan

- [x] `bun test packages/adapter-go-template` — 154 pass, 0 fail (previously 153 pass + 1 skip)
- [x] `bun test packages/adapter-go-template packages/adapter-tests` — 419 pass, 0 fail
- [x] Pre-existing `@barefootjs/client` resolution failures from this worktree are unchanged from `main` baseline

Closes #1296

### DIFF
--- a/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
+++ b/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
@@ -46,11 +46,6 @@ runAdapterConformanceTests({
     'nullish-coalescing-jsx',
     'return-nullish-coalescing',
     'return-map',
-    // #1296: test harness only loads `bf.FuncMap()`, which omits
-    // `bfAsyncBoundary` (lives in `StreamingFuncMap()`). The adapter
-    // correctly emits the call; the harness needs to merge the
-    // streaming map before the fixture can pass.
-    'async-boundary',
     // #1297: `compileJSX` with `outputIR: true` does not emit an `ir`
     // file for `'use client'` sources when the go-template adapter
     // is selected, even though the Hono path does. The harness then
--- a/packages/adapter-go-template/src/test-render.ts
+++ b/packages/adapter-go-template/src/test-render.ts
@@ -167,6 +167,18 @@ import (
 // Silence unused import for bf if only FuncMap is used
 var _ = bf.FuncMap
 
+// Merge StreamingFuncMap into the base FuncMap so fixtures using
+// <Async> (which compiles to a bfAsyncBoundary call) can be parsed
+// by the test harness. See packages/adapter-go-template/runtime/streaming.go
+// for the recommended merge recipe.
+func bfTestFuncMap() template.FuncMap {
+	funcMap := bf.FuncMap()
+	for k, v := range bf.StreamingFuncMap() {
+		funcMap[k] = v
+	}
+	return funcMap
+}
+
 const tmplContent = \`${escapedTemplate}\`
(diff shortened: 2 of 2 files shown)


kfly8:
Rebased onto `main` and dropped the duplicate `<div bf-async>` strip from `stripConditionalMarkersForCrossAdapter`. The depth-counted `stripAsyncPlaceholders` shipped in #1301 already handles the same placeholder (and is robust to nested `<div>` fallbacks, attribute reordering, `<divider>`-style tag boundaries, and self-closing `<div/>` — all of which the simpler regex this PR used to add was not). `jsx-runner.ts` is back to byte-equal with `main`.

Scope of this PR is now exactly the `bfTestFuncMap()` helper in `test-render.ts` plus the `skipJsx` removal — see the diff.

PR #1305, opened 2026-05-16
fix(#1297): emit IR per component for multi-component sources

### LINKED ISSUE #1297: go-template test harness: 'use client' source emits no IR file, breaking conformance for client-only fixtures
## Problem

The go-template test harness in `packages/adapter-go-template/src/test-render.ts:105-106` does:

```ts
const irFile = result.files.find(f => f.type === 'ir')
if (!irFile) throw new Error('No IR output (set outputIR: true)')
```

But `compileJSX(..., { adapter: new GoTemplateAdapter(), outputIR: true })` on a `'use client'` source returns only `markedTemplate`, `types`, and `clientJs` files — no `ir` file. The harness then throws even though `outputIR: true` was passed.

## Reproduction

The `context-provider` fixture added in #1252 Phase 0:

```jsx
'use client'
import { createContext, useContext } from '@barefootjs/client'
const ThemeContext = createContext('light')
function ThemeLabel() {
  const theme = useContext(ThemeContext)
  return <span class="theme">{theme}</span>
}
export function ThemeRoot() {
  return (
    <div class="root">
      <ThemeContext.Provider value="dark"><ThemeLabel /></ThemeContext.Provider>
    </div>
  )
}
```

Compiling this with the go-template adapter + `outputIR: true` produces three files but no IR. The Hono adapter produces an IR file under the same conditions.

## Root cause (suspected)

The IR emission path inside `compileJSX` likely short-circuits for client components on the go-template adapter — either intentionally (different code path) or as an unintended interaction. Worth investigating whether IR should always be emitted when `outputIR: true` is requested.

## Fix candidates

1. Make `outputIR: true` always emit IR regardless of `isClientComponent` and adapter.
2. Detect "no IR for client components" in `renderGoTemplateComponent` and skip / handle gracefully (less desirable — this is a contract bug).

## Why this matters

The IR `provider` kind has **no go-template conformance coverage** until this is fixed. Combined with #1295 (CSR harness lacks `createContext` mock), Provider currently only has Hono SSR test coverage — same silent-gap class #1252 documents.

## Related

- Discovered during #1252 Phase 0.
- `context-provider` fixture is currently in `skipJsx` with a pointer to this issue.

kfly8:
## Summary

- `compileMultipleComponents` now emits one `*.<ComponentName>.ir.json` file per component when `outputIR: true` is passed. The previous code path skipped IR emission entirely for any source with two or more components — including every `'use client'` file that defines a helper alongside the exported component — silently returning `markedTemplate` + `clientJs` + `types` and no `ir`.
- Updated the go-template and Mojo test renderers to handle the multi-IR result: pick the entry-point IR by `hasDefaultExport` → `isExported` → first, and concatenate type definitions for sibling components.
- For the Mojo harness specifically, replaced the basename-derivation regex with a `templateBaseName` helper that strips the full multi-segment `adapter.extension` (`.html.ep`). The previous `/\.[^.]+$/` regex left `Counter.html`, silently pairing every sibling template to the entry-point IR via the `?? irs[0]` fallback. No existing fixture exercised this gap (the only multi-component Mojo case is `skipJsx`-ed for a separate SSR-provider issue), so the bug was invisible.
- Side-fix in `GoTemplateAdapter.collectStaticChildInstancesRecursive`: walk `provider` and `async` node children so a static `<Child/>` nested under a context provider still gets a slot field generated on the parent's props struct.

The contract this restores: "if the user asks for IR, they get IR" — regardless of `isClientComponent` and regardless of adapter.

## Scope clarification — `context-provider` JSX fixture

The issue asked for the previously-skipped `context-provider` fixture to pass on both adapters once IR emission is fixed. The IR-emission gate is now correct, but the fixture **still cannot pass** on go-template / Mojo because those adapters have no SSR-side context-propagation mechanism:

- The Hono adapter rewrites `<Ctx.Provider value="dark">…</Ctx.Provider>` into a `provideContextSSR(Ctx, "dark", …)` call that the Hono SSR JS runtime resolves; `useContext(Ctx)` then returns `"dark"`.
- Go template and Mojo emit `{{template "Child" .ChildSlot0}}` / `<%== bf->render_child('child') %>` against templates that reference `.Theme` / `<%= $theme %>` — but neither adapter threads the provider value into the child template's data, so the field is unresolved at template-eval time.

Implementing SSR context resolution on the two template-based adapters is an adapter-feature task, not an IR-emission gap. The fixture stays in `skipJsx` with a rewritten comment that points at the actual remaining gap on each adapter (please open a follow-up issue if SSR provider coverage is wanted on go-template / Mojo).

## Test plan

- [x] `bun test --cwd packages/jsx` — includes the new `ir-multi-component-emission.test.ts` regression
- [x] `bun test --cwd packages/adapter-mojolicious` — includes the new `template-base-name.test.ts` unit test
- [x] `bun test --cwd packages/adapter-go-template` / `packages/adapter-tests` / `packages/adapter-hono` — no new regressions
- [x] Rebased onto current `main` (includes #1301, #1302, #1304)

Closes #1297

### DIFF
--- a/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
+++ b/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
@@ -46,11 +46,15 @@ runAdapterConformanceTests({
     'nullish-coalescing-jsx',
     'return-nullish-coalescing',
     'return-map',
-    // #1297: `compileJSX` with `outputIR: true` does not emit an `ir`
-    // file for `'use client'` sources when the go-template adapter
-    // is selected, even though the Hono path does. The harness then
-    // throws "No IR output (set outputIR: true)" before any rendering
-    // happens. Provider IR coverage on go-template waits on the fix.
+    // #1297 fixed the harness-side IR emission gate (multi-component
+    // sources now emit one `ir` file per component, and the harness
+    // picks the entry-point IR). The remaining gap is adapter-side:
+    // the go-template adapter has no SSR context-propagation
+    // mechanism, so `<Ctx.Provider value="dark">` doesn't make
+    // `useContext(Ctx)` resolve to `"dark"` at template-eval time —
+    // the template emits `.Theme` against a struct that has no
+    // `Theme` field. Provider SSR coverage on go-template waits on
+    // that adapter feature; see #1297 follow-up.
     'context-provider',
   ],
   // Per-fixture build-time contracts for shapes the Go template
--- a/packages/adapter-go-template/src/adapter/go-template-adapter.ts
+++ b/packages/adapter-go-template/src/adapter/go-template-adapter.ts
@@ -1130,6 +1130,24 @@ export class GoTemplateAdapter extends BaseAdapter implements ParsedExprEmitter,
       if (cond.whenFalse) {
         this.collectStaticChildInstancesRecursive(cond.whenFalse, result, inLoop)
       }
+    } else if (node.type === 'provider') {
+      // Provider is a transparent wrapper at the SSR layer — context
+      // propagation is purely a client-runtime concern. Recurse into
+      // its children so any static <Child/> nested under <Ctx.Provider>
+      // still gets a slot field generated on the parent's props type.
+      const p = node as IRProvider
+      for (const child of p.children) {
+        this.collectStaticChildInstancesRecursive(child, result, inLoop)
+      }
+    } else if (node.type === 'async') {
+      // Async fallback + children render server-side via the OOS
+      // protocol; static child components inside them still need slot
+      // fields on the parent struct.
+      const a = node as IRAsync
+      this.collectStaticChildInstancesRecursive(a.fallback, result, inLoop)
+      for (const child of a.children) {
+        this.collectStaticChildInstancesRecursive(child, result, inLoop)
+      }
     }
   }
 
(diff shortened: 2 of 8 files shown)


kfly8:
Rebased onto current `main` (#1301/#1302/#1304 are now in main) and addressed two reviewer findings in commit cd32a489:

**Mojo basename-pairing bug** — `templatesPerComponent` writes `<ComponentName>.html.ep` (Mojo's `adapter.extension` is multi-segment), but the previous `/\.[^.]+$/` regex only stripped `.ep`, leaving `Counter.html` — which never matched the bare component name in `irsByName`. Every sibling template silently paired to the entry-point IR via the `?? irs[0]` fallback. No existing fixture exercised this path so the bug was invisible. Replaced with a `templateBaseName(path, extension)` helper that strips the full `adapter.extension`, and pinned with a 4-case unit test (`template-base-name.test.ts`).

**Rebase conflicts** — `skipJsx` arrays in both go-template and Mojo test files had non-overlapping reasons (#1296/#1298 cited `async-boundary`; this PR cites `context-provider`). Resolved manually by keeping main's post-merge state and applying this PR's `context-provider` comment.

Open follow-ups noted but not implemented in this PR:
- Path-naming asymmetry between single (`<base>.ir.json`) and multi (`<base>.<Name>.ir.json`) cases
- Picking-logic duplication across go-template and Mojo harnesses (4-line copy-paste)
- Side-fix commit (the `provider`/`async` walk in `collectStaticChildInstancesRecursive`) not separated from the core fix

These can become follow-up issues if desired.

Issue #1244, opened 2026-05-12
Compiler stress test: corner cases that escape the happy path


kfly8:
## Goal

Stress the compiler so we remove defects and **faithfully render whatever TSX a user writes**.

This issue is an ongoing catalog of compiler / runtime patterns that aren&#39;t yet well exercised by the existing test suite or block gallery. Each item should produce one of: a fix, a documented limitation, or — best case — a green test that locks the path in.

## The coverage grid we&#39;re sampling

Patterns multiply along several independent axes:

- **Reactive primitive** — signal, memo, effect, store, context, ref
- **Binding site** — attribute, `style={{}}` member, `className` interpolation, event handler, child text, ref, spread, child component prop
- **Value shape** — literal, identifier, template literal, ternary, function call, object literal, spread, `&amp;&amp;` / `||` short-circuit
- **Control-flow context** — top-level, inside `.map()`, inside conditional branch, inside ``, inside ``, inside portal
- **Lifecycle moment** — SSR template, initial hydration, signal update, mount, unmount, error
- **Identifier resolution** — loop param, destructured loop param, scope context, component name, key, slot, ref forwarding

A sparse sample of that product still has far more cells than the suite covers today. The TODO below picks combinations where the wiring is non-obvious, where prior fixes have lived, or where user-reported surprises seem plausible. New items welcome.

## Recurring design patterns that produce defects

Some defects recur because of structural patterns in the compiler and runtime rather than missing tests. Listing them so the catalog above can also surface refactor candidates — addressing the pattern at the source removes a whole family of future defects rather than one instance.

### A. Two emit paths for reactive attributes

`emit-reactive.ts` (top-level), `stringify/inner-loop.ts` (nested-loop renderItem), and the conditional-branch arm each implement their own per-case dispatch over attribute kinds — `className`, `value`, boolean attribute, `presenceOrUndefined`, generic. Each new kind must be added to every site. Omissions are silent: the emitted code looks plausible but uses the wrong shape (a generic `__v != null` branch instead of a truthy check, etc.).

Refactor direction: a single `emitReactiveAttr({ target, attrName, expression, meta }) → lines[]` helper consumed from every emit site.

### B. Parallel `TopLevelLoop` / `NestedLoop` IR types

[...]

Issue #1071, opened 2026-04-27
Compiler: re-attach reactive attribute bindings when a conditional swaps in a fresh element (clean up graph-editor preview workaround)


kfly8:
## Background

Discovered while implementing the Graph/DAG Editor block (#135). The block has a dashed cyan **drag-to-connect preview path** mounted via a conditional:

```tsx
{connectPreview() !== null ? (
  <path d={connectPreview() ?? ''} ... />
) : null}
```

The compiler emits the reactive \`d\` updater (\`createEffect\` over \`connectPreview()\`) **outside** either branch's \`bindEvents\`. When \`insert.ts\` swaps in a fresh \`<path>\` element on the rising edge of the condition, the original effect still holds a reference to the now-detached placeholder and never writes to the live element. As a result the preview \`d\` stays at its initial value and the line never tracks the cursor.

The element-namespace half of the issue (\`<path>\` came back as \`HTMLUnknownElement\`) was fixed in PR #1070 by making \`parseHTML\` SVG-context-aware. That PR also confirmed via \`barefoot inspect\` that the reactive update on \`_s14\` (the path slot) is wired to a single, statically-known DOM node — not refreshed when \`insert\` replaces that node.

## Repro (without the workaround)

\`site/ui/components/graph-editor-demo.tsx\` — replace the always-mounted preview block with the natural conditional shape:

\`\`\`tsx
{connectPreview() !== null ? (
  <path
    className=\"graph-connect-preview\"
    data-connect-preview
    d={connectPreview() ?? ''}
    stroke=\"#2563eb\"
    strokeWidth=\"1.5\"
    strokeDasharray=\"4 4\"
    fill=\"none\"
    pointerEvents=\"none\"
  />
) : null}
\`\`\`

Then in Chrome: start dragging from a node handle and observe that the preview \`<path>\` is mounted (\`tagName: \"path\"\`, SVG namespace ✓ after PR #1070) but its \`d\` attribute stays at the source-source value (\`M 80 100 C 120 100, 40 100, 80 100\`) regardless of cursor position. \`bbox\` is essentially zero because the path collapses to a single point.

## Expected

When the conditional flips from null → element, the reactive bindings whose target slot lives inside the active branch should be re-attached to the freshly-mounted element. Subsequent signal updates should reach the live node.

## Fix sketch

[...]

PR #1351, opened 2026-05-18
feat(jsx/#1244): per-item ref callbacks fire on every renderItem (mount + remount)

kfly8:
## Summary

Catalog item from [#1244](https://github.com/piconic-ai/barefootjs/issues/1244):

> [ ] ref callback re-invocation when the element re-mounts under the same key

Before this PR, `<el ref={fn}>` inside a `.map()` was silently dropped at compile time — the user's `const fn = ...` went unreferenced and the emitted `mapArray` per-item factory never called it. The callback fired neither on initial mount of a keyed item nor when that same key was unmounted and remounted later.

The fix collects per-item refs (`TopLevelLoop.childRefs`) and emits `(fn)(__rf)` inside the multi-line factory body — once per `renderItem` invocation, which is every actual mount: SSR hydration, initial CSR creation, and same-key remount after unmount. `mapArray` does not call `renderItem` for same-key reactive updates (those flow through per-item signal `setItem(...)`), so the callback does not over-fire on plain prop changes.

For comparison:

- Top-level refs compile to `if (_sN) (fn)(_sN)` at the end of `init`.
- Refs inside a conditional branch compile inside the branch's `bindEvents` so they re-attach on swap (#1071 path).
- Loop items previously had no equivalent collection point — `collectFromElement` writes to `ctx.refElements`, but the loop visitor in `collectElements` intentionally stops at the loop scope. The per-item factory built by the loop planner did not consume any refs.

## Commits

1. **`test(jsx/#1244): pin ref re-invocation on same-key remount as todo`** — original `.todo` lock-in (was the whole previous draft scope).
2. **`feat(jsx/#1244): collect per-item ref callbacks into TopLevelLoop IR`** — `LoopChildRef` + `collectLoopChildRefs` + threading through `collectElements` + reference-graph edge so the callback identifier survives dead-code elim.
3. **`feat(jsx/#1244): emit per-item ref calls inside the plain-loop factory`** — `PlainLoopVariant.childRefs` + plan + stringifier emits `qsa(__el, '[bf="<slot>"]')` + `emitRefCall`. Non-empty refs force the multi-line factory layout the same way reactive effects do. Multi-root bodies use `qsaItem` (#1212 pattern).
4. **`test(jsx/#1244): drop .todo on the ref re-invocation stress entry`** — assertion now runs against the corrected compiler.

## Scope

Plain top-level loops only — covers the most common shape (`<el key={…} ref={…}>` inside a `.map()`). Composite / component / branch / nested loops still drop refs; those paths each need the same wiring and will land in follow-ups so the diff stays bisectable. The Layer 6 E2E (`ref-remount` `test.fixme` in `site/ui/e2e/stress-1244.spec.ts`) is still pending a demo page.

## Test plan

- [x] `bun test packages/jsx/src/__tests__/compiler-stress-1244.test.ts -t 'ref callback re-invocation'` — passes
- [x] `bun test packages/jsx` — 1244 pass, 3 todo, 0 fail (no regression)
- [x] `bun test packages/adapter-tests` — 316 pass, 0 fail
- [x] `bun test packages/client` — 267 pass, 0 fail
- [x] `bunx tsc --noEmit` in packages/jsx — clean
- [ ] Follow-up PRs cover composite / component / branch / nested loop variants
- [ ] Layer 6 E2E (`/stress/1244/ref-remount` demo page) confirms remount runtime semantics

### DIFF
--- a/packages/jsx/src/__tests__/compiler-stress-1244.test.ts
+++ b/packages/jsx/src/__tests__/compiler-stress-1244.test.ts
@@ -878,37 +878,28 @@ describe('two refs on the same element via a composeRefs helper', () => {
 })
 
 describe('ref callback re-invocation on remount under the same key', () => {
-  // SURFACED LIMITATION (#1244 catalog: "ref callback re-invocation when
-  // the element re-mounts under the same key"): a `ref={fn}` on a JSX
-  // element inside `.map()` is silently dropped at compile time. The
-  // emitted `mapArray` per-item factory creates the DOM node but never
-  // calls `fn(__el)` — so the callback fires neither on the initial
-  // mount of a keyed item nor when that same key is unmounted and
-  // remounted later (the runtime case the E2E `ref-remount` fixme is
-  // chasing).
+  // #1244 catalog item "ref callback re-invocation when the element
+  // re-mounts under the same key". Pre-fix, a `ref={fn}` on a JSX
+  // element inside `.map()` was silently dropped at compile time:
+  // the user's const went unreferenced and the emitted `mapArray`
+  // per-item factory never called it, so the callback fired neither
+  // on the initial mount of a keyed item nor when that same key was
+  // unmounted and remounted later.
   //
-  // For comparison: refs on a top-level element compile to a single
-  // `if (_sN) (fn)(_sN)` line at the end of `init`, and refs inside a
-  // conditional branch compile inside the branch's `bindEvents` so they
-  // re-attach on swap (#1071 path). The loop case has no equivalent
-  // collection point — `collectFromElement` writes `ctx.refElements`,
-  // but the per-item factory built by the loop planner does not consume
-  // it. The fix needs the loop planner to surface the per-item ref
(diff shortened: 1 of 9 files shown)


kfly8:
Closing in favor of a deeper refactor.

The original scope (plain-loop only) reproduces exactly the divergence pattern that issue #1244 §B warns about — per-item concepts replicated across TopLevelLoop / BranchLoop / NestedLoop and across each plan variant. Adding `childRefs` to one variant without a unified collection point would make the next per-item concept addition equally painful.

A new PR will land:
1. **Phase A**: unify per-item bindings (events / reactiveAttrs / reactiveTexts / refs / conditionals) into a single `LoopChildBindings` struct on `LoopCore`; byte-equal output gated by existing suites.
2. **Phase B**: emit ref calls inside every dynamic-loop variant's per-item factory (plain / composite / static / branch / nested).
3. **Phase C**: expand catalog tests so the contract is locked in per variant.

Closing this branch will not affect history; the work re-lands cleanly on top of main.

PR #1354, opened 2026-05-18
fix(jsx/#1244): CSR emit collapses element spread + explicit attrs into one spreadAttrs({...}) merge

kfly8:
## Summary

- Reproduces and fixes a JSX-semantics violation in the CSR emit path
  surfaced while writing the catalog test for #1244's
  "Destructured loop param — `tasks.map(({ id, title, ...rest }) => …)`
  with `rest` spread back onto the root" item.
- When a JSX element has `{...spread}` alongside non-`key` explicit attrs,
  source-order inline emission silently inverts JSX rightmost-wins on
  collision (HTML's duplicate-attribute parse rule keeps the FIRST
  occurrence). Hono SSR already got this right via its JSX runtime;
  CSR landed `<element explicit ${spreadAttrs(rest)}>` and the explicit
  value survived browser parsing.
- The fix collapses both into a single
  `spreadAttrs({...source-order-members...})` call so JS object-literal
  evaluation resolves the collision before serialization. Applied
  identically across all three CSR-side emit paths (`irToHtmlTemplate`,
  `irToComponentTemplate`, `generateCsrTemplate`). `key` (→ `data-key`)
  stays outside the merge to preserve the unconditional
  `data-key="${value}"` debug contract.

## Verified failure mode (before this PR)

Source:

```jsx
<li key={id} data-priority="medium" {...rest}>{title}</li>
// rest.data-priority === 'high'
```

Rendered DOM after browser parsing:

| Adapter | data-priority value | JSX semantics |
|---------|--------------------|--------------|
| Hono SSR | `'high'` | ✅ |
| CSR (before fix) | `'medium'` | ❌ |
| CSR (after fix) | `'high'` | ✅ |

## Contract pinning

- New fixture `rest-destructure-object-spread-in-map` (Hono / Go / Mojo
  green; CSR skipped — cosmetic-only attribute-order divergence for the
  no-collision lone-spread case).
- New `compiler-stress-1244.test.ts` test asserts both emit sites carry
  the merged `spreadAttrs({"data-priority": "medium", ...rest})` form.
- Three pre-existing tests updated to the new merged shape
  (`client-js-generation`, `template-closure`,
  `staged-ir/01-template-scope`). Behavioural contracts (init-scope
  leak detection, no spread dropped) are preserved.

## Test plan

- [x] `bun test packages/jsx packages/client packages/adapter-hono packages/adapter-go-template packages/adapter-mojolicious packages/adapter-tests`
  → 2385 pass / 0 fail / 33 skip / 3 todo
- [x] Manual SSR + CSR DOM check confirms `rest.data-priority` wins on
  both adapters after fix.

### DIFF
--- a/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
+++ b/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
@@ -122,6 +122,10 @@ runAdapterConformanceTests({
     // limitation declarative: when the Go adapter grows a native
     // rest-lowering, dropping these entries flips the contract on.
     'rest-destructure-object-in-map': [{ code: 'BF104', severity: 'error' }],
+    // #1244 catalog: rest spread back onto the root element. Same
+    // refusal shape as the read-only variant above — `paramBindings`
+    // is non-empty so BF104 fires regardless of how `rest` is used.
+    'rest-destructure-object-spread-in-map': [{ code: 'BF104', severity: 'error' }],
     'rest-destructure-array-in-map': [{ code: 'BF104', severity: 'error' }],
     'rest-destructure-nested-in-map': [{ code: 'BF104', severity: 'error' }],
     // #1244 stress catalog #13 (#1324): JSX spread of a reactive object
--- a/packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts
+++ b/packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts
@@ -85,6 +85,10 @@ runAdapterConformanceTests({
     // refusal regardless of whether the binding is rest or plain.
     // Pinning the contract here makes the limitation declarative.
     'rest-destructure-object-in-map': [{ code: 'BF104', severity: 'error' }],
+    // #1244 catalog: rest spread back onto the root element. Same
+    // refusal shape as the read-only variant above — `paramBindings`
+    // is non-empty so BF104 fires regardless of how `rest` is used.
+    'rest-destructure-object-spread-in-map': [{ code: 'BF104', severity: 'error' }],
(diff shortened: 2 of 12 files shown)


kfly8:
Thanks for the careful review — both comments addressed.

**① `clipPath` divergence** (`ef2b4af3`'s parent `81031561`):
Confirmed `clipPath` is already in the compile-time `SVG_CAMEL_TO_KEBAB` table (`utils.ts:217`) and that mixing the two paths would split the same JSX prop into `clip-path` (explicit) vs `clipPath` (spread). Dropped `clipPath` from the runtime allowlist and pinned the corrected behaviour in the test: `viewBox` / `clipPathUnits` stay camelCase, `clipPath` / `strokeWidth` / `strokeLinecap` lower to kebab-case. The runtime allowlist now mirrors only the XML attribute names that have no kebab-case entry in `SVG_CAMEL_TO_KEBAB`.

**② Triple-implemented merge logic** (`ef2b4af3`):
Consolidated into three shared helpers: `isMergeableAttr`, `shouldUseSpreadAttrsMerge`, `buildSpreadAttrsMergeCall` (with a `MergeContext` for per-path knobs — `clientOnly`, rest-prop filter — and callbacks for the genuinely-variant expression lowering). All three emit paths now go through the same predicates and the same builder.

You also noted the silent drift the duplication enabled — `boolean-shorthand` reaching the merge object on `irToHtmlTemplate`'s helper but being filtered on the two inline copies. The consolidated `buildSpreadAttrsMergeCall` now resolves `boolean-shorthand` to `true` for all three paths, matching JSX boolean-attribute semantics; the inline copies were silently dropping those before.

All tests still green; CI should reflect the same.

PR #1358, opened 2026-05-18
fix(jsx/#1244): accept literal key={null} / key={undefined} / ternary-with-falsy

kfly8:
## Summary

Working on issue #1244's catalog item "`key={0}`, `key={false}`, `key={null}` — what happens?".

**Before:** `classifyKeyProblem` raised BF023 ("Missing key in list rendering") for any explicit `key={null}` or `key={undefined}` literal, and for any ternary chain that could reach one of those literals (the type-based nullable check saw the branch union `T | null` and false-positived). Both the code and the category were misleading: the key IS present, and React's own behaviour for these values is a runtime warning, not a compile error.

**After:** Explicit literal `null` / `undefined` keys are accepted — `mapArray`'s `String()` coercion produces the per-item key (`"null"`, `"undefined"`). `ConditionalExpression` keys skip the type-based nullable check so per-item differentiated ternaries (`key={i === 0 ? 0 : i === 1 ? false : null}`) lower cleanly. Numeric / boolean keys (`0`, `false`) were already accepted; this extends the same policy.

**Preserved:** Key truly missing (no attribute / bare `<el key>` / `key={}`) still raises BF023. INFERRED nullability via the type checker (`item.id` where `id?: string`) still raises BF023 — the user didn't deliberately opt into a null key, the type system caught a missing field at the source.

## Behaviour matrix

| Input | Before | After |
|---|---|---|
| `key={0}` | accept | accept |
| `key={false}` | accept | accept |
| `key={null}` | BF023 ❌ | accept |
| `key={undefined}` | BF023 ❌ | accept |
| `key={""}` | accept | accept |
| `key={i === 0 ? 0 : i === 1 ? false : null}` | BF023 ❌ | accept |
| `key={item.id}` where `id?: string` | BF023 | BF023 (unchanged) |
| `key` attribute missing | BF023 | BF023 (unchanged) |

## Test plan

- [x] `bun test packages/jsx packages/client packages/adapter-*` → 2395 pass / 0 fail / 2 todo (down from 3 — this PR replaces one `test.todo`)
- [x] Five new Layer 1 cases in `compiler-stress-1244.test.ts` (`0` / `false` / `null` / `undefined` / ternary)
- [x] Two existing BF023 suite tests (`a-2: key={null}`, `a-2: key={undefined}`) flipped to "no error" with docstring noting the new policy

### DIFF
--- a/packages/jsx/src/__tests__/compiler-stress-1244.test.ts
+++ b/packages/jsx/src/__tests__/compiler-stress-1244.test.ts
@@ -723,14 +723,71 @@ describe('computed key', () => {
 })
 
 describe('key edge values', () => {
-  // SURFACED LIMITATION (#1244 sub-issue): BF023 (missing key) currently
-  // fires when a literal-falsy key (`key={0}`, `key={false}`,
-  // `key={null}`) is reachable through a ternary. The intended
-  // behaviour is either to accept the value verbatim (ints + booleans
-  // are valid React-style keys) or to diagnose with a more specific
-  // "key cannot be falsy" code, not the generic missing-key error. The
-  // body asserts a clean compile. Drop `.todo` once fixed.
-  test.todo('key={0}, key={false}, key={null} via ternary — accepted as a key', () => {
+  // BF023 (missing-key) used to fire on explicit literal-falsy keys
+  // (`key={null}`, `key={undefined}`) and on any ternary chain whose
+  // branches reached one of those literals — silently routing what
+  // React treats as a runtime warning into a hard compile error with a
+  // misleading "missing key" message. The user explicitly wrote the
+  // value; treat it as a runtime concern and let `mapArray`'s
+  // `String()` coercion produce the per-item key (`"null"`,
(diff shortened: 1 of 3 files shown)


kfly8:
Thanks for catching that — addressed in `04055f59`.

The blanket `isConditionalExpression(expr) → null` bypass was too broad: it disabled the type-based nullable check for every ternary, silently losing the BF023 diagnostic on `key={cond ? item.id : item.fallback}` shapes (where the user did NOT explicitly opt into a `null` / `undefined` literal — the nullability is inferred from the field type).

Narrowed via a new `conditionalHasExplicitNullishBranch` walker that recurses into the ternary (and any nested ternaries inside its branches, stripping parens) and only returns true when at least one leaf is a `NullKeyword` or the `undefined` identifier. The bypass now fires for the catalog's deliberate-opt-in case (`key={i === 0 ? 0 : i === 1 ? false : null}`) but stays out of the way for the inferred-nullable case.

Regression test `a-3: key with inferred-nullable type inside a ternary still raises BF023` in `missing-key-in-list.test.ts` locks the contract.

PR #1399, opened 2026-05-19
fix(jsx): honor /* @client */ on conditional / ifStatement / loop nodes in stage diagnostics

kfly8:
## Summary

`/* @client */` is documented as the escape hatch for stage-violation diagnostics (BF060/BF061) — wrapping a JSX expression in `{/* @client */ EXPR}` should mark the expression `clientOnly`, which routes through `clientOnlyElements` and suppresses the "referenced from template scope" warning.

The `collectTemplateRiskyNames` visitor in `compute-inlinability.ts` honored this for `expression` nodes (line 467-468 pre-fix) but not for `conditional`, `ifStatement`, or `loop` nodes — those walked unconditionally and added every referenced identifier to `templateRiskyNames`, firing BF061 for any chained init-local in the condition / array even when the wrapper was already there.

The IR types already carry `clientOnly?: boolean` on all four node kinds (`IRExpression`, `IRConditional`, `IRIfStatement`, `IRLoop`); the gap was purely in this visitor's branches.

## Repro

Encountered in piconic-ai/desk #86 Phase 6 (`IssueCardNode.ts → .tsx` migration):

```tsx
'use client'
import { useSettings } from './nodes'

export function Foo() {
  const setting = useSettings()
  const enabled = setting === 'on'           // chained init-local
  return <div>{/* @client */ enabled ? 'YES' : 'NO'}</div>
}
```

Pre-fix:
```
[BF061] Init-scope local 'setting' referenced from template scope
        (via const 'enabled'). … Wrap the JSX expression in
        /* @client */ to defer evaluation, or lift the value …
```

The user is told to wrap with `/* @client */` — but the wrapper is already there. The diagnostic is a false positive whenever the only template-side reference is a `clientOnly` conditional / loop.

## Fix

Add the same early-return the `expression` visitor uses:

```ts
conditional: ({ node: c, descend }) => {
  if (c.clientOnly) return    // ← new
  addExprIdents(c.templateCondition ?? c.condition)
  descend()
},
ifStatement: ({ node: i, descend }) => {
  if (i.clientOnly) return    // ← new
  addExprIdents(i.templateCondition ?? i.condition)
  descend()
},
loop: ({ node: l, descend }) => {
  if (l.clientOnly) return    // ← new
  addExprIdents(l.templateArray ?? l.array)
  descend()
},
```

## Regression coverage

Two new tests in `packages/jsx/src/__tests__/staged-ir/10-stage-diagnostics.test.ts` alongside the existing `expression` carve-out test:

- `chained const referenced only inside a /* @client */ ternary does NOT fire BF061`
- `chained const referenced only inside a /* @client */ .map() does NOT fire BF061`

## Verification

- `bun test packages/jsx/src/__tests__/staged-ir/10-stage-diagnostics.test.ts` — **28 / 28 pass** (was 26 + the 2 new ones from this PR)
- Red-bar confirmed: stashing only the `compute-inlinability.ts` change leaves the 2 new tests failing
- `bun test packages/jsx` (full) — 1266 pass / 4 pre-existing fail (alias / locale tests, unrelated), 0 new failures vs main's 1264 pass / 4 fail baseline. Net: +2 pass.

## Out of scope

- `provider` and `component` visitors don't yet honor `clientOnly` either, but they don't surface BF061 today (component props are stripped on UNSAFE per the visitor's own comment, providers have their own carve-out path). If a similar repro surfaces for those, the same pattern would apply.
- Downstream in desk: with this fix, Phase 6's `IssueCardNodeImpl.tsx` can use `{/* @client */ isDraft ? <Draft/> : <Issue/>}` to defer the conditional to client without lifting `isDraft` to a memo.

### DIFF
--- a/packages/jsx/src/__tests__/staged-ir/10-stage-diagnostics.test.ts
+++ b/packages/jsx/src/__tests__/staged-ir/10-stage-diagnostics.test.ts
@@ -577,4 +577,48 @@ describe('compileJSX surfaces stage-violation diagnostics by default', () => {
     // No diagnostic noise either way.
     expect(errors.find(e => e.startsWith('[BF06'))).toBeUndefined()
   })
+
+  // Regression: prior to this fix, only the `expression` IR visitor in
+  // `collectTemplateRiskyNames` consulted `clientOnly`. The
+  // `conditional`, `ifStatement`, and `loop` visitors walked
+  // unconditionally — so wrapping the condition / array with
+  // `/* @client */` (which DOES mark the IR node `clientOnly: true`)
+  // still added every referenced identifier to `templateRiskyNames`,
+  // and `recordStageDiagnostics` then fired BF061 for any chained
+  // init-local. The `@client` directive is documented as the escape
+  // hatch for these cases; the carve-out belongs in all three node
+  // types, not just `expression`.
+  test('chained const referenced only inside a /* @client */ ternary does NOT fire BF061', () => {
+    const { errors } = compile(`
+      'use client'
+      import { useSettings } from './nodes'
+
+      interface Props {}
+
+      export function Foo(_props: Props) {
+        const setting = useSettings()
+        const enabled = setting === 'on'
+        return <div>{/* @client */ enabled ? 'YES' : 'NO'}</div>
+      }
+    `)
+
+    expect(errors.find(e => e.startsWith('[BF061]'))).toBeUndefined()
+  })
+
+  test('chained const referenced only inside a /* @client */ .map() does NOT fire BF061', () => {
+    const { errors } = compile(`
+      'use client'
+      import { useSettings } from './nodes'
+
+      interface Props {}
+
+      export function Foo(_props: Props) {
+        const setting = useSettings()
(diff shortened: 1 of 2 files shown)


kfly8:
## CI fix (120669de)

CI's `build:types` step (`tsgo --emitDeclarationOnly`) caught a wrong assumption from my initial commit: I added `if (i.clientOnly) return` to the `ifStatement` visitor on the basis that `IRConditional`, `IRLoop`, and `IRIfStatement` all carry `clientOnly` — but only the first two do. `IRIfStatement` has no such field because `jsx-to-ir.ts` doesn't currently route `/* @client */` through `if`-statement nodes.

Narrowed the fix to just `conditional` + `loop` (the two cases the new regression tests exercise), with a comment in the `ifStatement` branch documenting the gap so a future addition lands at the right place.

`bun run --filter '@barefootjs/jsx' build` now passes the typecheck step locally.

The PR description's mention of all three node kinds is now slightly aspirational for `ifStatement` — the substantive fix and its tests are unchanged.

Issue #1443, opened 2026-05-20
Go + Mojo adapters: higher-order array methods can't lower (BF101) — blocks registry <Button>/<Slot>


kfly8:
Closing as completed. The originally-blocked driver — `npm create barefootjs@latest --adapter {mojo,echo}` shipping the registry `<Button>` / `<Slot>` instead of a native-`<button>` fallback — works end-to-end on both adapters. Summary of what landed:

| Pattern (from the issue table) | PR |
|---|---|
| `arr.filter(Boolean)` | #1444 (Mojo) / #1445 (Go) — `array-method` IR + array-literal IR + Boolean-callable synthesise |
| `arr.filter(({done}) => done)` (destructured param) | #1446 — parser-side normalisation |
| `arr.filter(function (x) { return x.done })` (`function` keyword) | #1446 — parser-side normalisation |
| `arr.filter(x => x.tags.filter(t => t.active).length > 0)` (nested) | #1447 — `.length` on a higher-order result lowers to `scalar(@{...})` (Mojo) / `len (bf_filter ...)` (Go) |
| `arr.reduce(...)` / `arr.forEach(...)` / `arr.flatMap(...)` | **Deferred to #1448** (catalog issue, `known-limitation`). registry usage audit confirmed no SSR-template positions reach these methods today, so deferring doesn't block the original onboarding goal. |

Registry scaffold wired up in #1449 — Mojo + Echo now ship `SHARED_COUNTER_TSX` (`<Button>`-based) instead of `NATIVE_BUTTON_COUNTER_TSX`, and `bundledRegistryComponents` defaults to `['button']` for both. End-to-end `bf build` passes 0 errors against `pkg.pr.new` previews on both adapters.

The unaddressed `reduce` / `forEach` / `flatMap` cell is tracked alongside the broader unsupported-method surface in #1448 so the next person picking that up gets the design context up front.

Issue #1421, opened 2026-05-19
Mojo adapter: infinite recursion on branch-local higher-order chain (Slot scaffold breaks `bf build`)


kfly8:
## Summary

`bf build` under `--adapter mojo` crashes with an infinite recursion (Node stack-overflow) when a `"use client"` component declares a branch-local variable whose initializer contains a higher-order array method chain (`.filter(...).join(...)` etc.) and references that local at a JSX attribute. The default scaffold's `Slot` component hits this pattern, so the **out-of-the-box `npm create barefootjs@latest -- --adapter mojo --yes` flow is broken** — the very first `npm run dev` fails before the server can boot.

## Repro

Minimal failing component (paste into any `--adapter mojo` scaffold's `components/ui/slot/index.tsx`):

```tsx
"use client"

function Slot({ children, className }: { children?: unknown; className?: string }) {
  if (children) {
    const merged = [className].filter(Boolean).join(' ')
    return <div className={merged}>x</div>
  }
  return <div>fallback</div>
}

export { Slot }
```

```sh
./node_modules/.bin/bf build
# RangeError: Maximum call stack size exceeded
# at parseModifiers (.../typescript.js:38288:26)
# ...
# at MojoAdapter.convertHigherOrderExpr
# at MojoAdapter.convertExpressionToPerl
# at MojoAdapter._convertExpressionToPerlPublic
# at MojoTopLevelEmitter.unsupported   <-- infinite loop here
# at emitParsedExpr
# at MojoTopLevelEmitter.member
# at MojoTopLevelEmitter.call
# at MojoAdapter.renderParsedExprToPerl
# at MojoAdapter.convertHigherOrderExpr   <-- back to start
```

**Does NOT crash if the branch wrapping is removed:**

```tsx
function Slot({ children, className }: { children?: unknown; className?: string }) {
  // same chain, but at function scope rather than inside an `if`
  const merged = [className].filter(Boolean).join(' ')
  return <div className={merged}>{children}</div>
}
```

Other adapters (`hono`) build the same scaffold cleanly — this is Mojo-specific.

## Root cause

Two interacting changes:

[...]

Issue #1389, opened 2026-05-18
Go + Mojo adapters: templatePrimitives V1 registry is identifier-path-only


kfly8:
Catalogued under #1365 (parent: #1244).

## Problem

The `templatePrimitives` V1 registry on both Go and Mojo adapters is identifier-path-only and explicit. It registers a fixed set of well-known callees (`JSON.stringify`, `Math.floor`, `Math.ceil`, `Math.round`, `Number`, `String`) that are mappable to template helpers (`bf_json`, `bf_floor`, etc).

User-imported callees (`import { customSerialize } from './serialize'`) and chained-const callees that reference user imports are NOT registered, so the adapter can't render them server-side. The template-primitive conformance tests skip these cases.

## Cases affected

- `TemplatePrimitiveCaseId.USER_IMPORT_VIA_CONST` — bespoke user import not in the registry
- `TemplatePrimitiveCaseId.NO_DOUBLE_REWRITE_OF_PROPS_OBJECT` — uses `customSerialize` too

## Workaround

Precompute the result before passing into JSX:

```tsx
// Instead of: <div data-x={customSerialize(props.config)}>
const config = customSerialize(props.config)
// In JSX:    <div data-x={config}>
```

The compiler will lift `config` to a struct field at SSR time. The Go adapter renders it as a string; Mojo renders it via `<%= $config %>`. The callee runs at request handler time, not template-eval time.

## Fix direction

A user-extensible `templatePrimitives` map: the adapter accepts a `(callee, args) → template-emission` registry from user config, so projects can register their own callees against Go/Mojo template helpers they ship alongside.

The V1 contract is intentionally narrow to keep the surface small; expanding it should wait until at least one user has requested it.

## References

- Adapter docstrings:
  - `packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts:138-153`
  - `packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts:111-126`
- V1 surface definitions: `templatePrimitives` in `go-template-adapter.ts` / `mojo-adapter.ts`

PR #1444, opened 2026-05-20
fix(mojo/#1443): lower [a, b].filter(Boolean).join(' ') so the registry Slot renders server-side on Mojo

kfly8:
## Summary

Closes the Mojo side of #1443 for the registry `<Slot>` shape.

The registry `<Slot>` builds its merged className via:

```tsx
const mergedClass = [className, childClass].filter(Boolean).join(' ')
```

Pre-#1443 each link in the chain (array literal, `Boolean` callable filter, `.join`) hit a separate refusal gate on the Mojo adapter, so the scaffold `<Button>` / `<Card>` were unusable on Mojo. This PR lowers all three to Embedded Perl:

```perl
join(' ', @{[grep { $_ } @{[$className, $childClass]}]})
```

## Changes

- **Parser** (`packages/jsx/src/expression-parser.ts`): `array-literal` joins `ParsedExpr` as a first-class IR node with recursive `isSupported` over its elements. `.filter(Boolean)` becomes `higher-order` with a synthetic identity predicate so adapters reuse their existing higher-order paths instead of a callee-resolution branch we don't want to generalise (#1389). The `higher-order` `isSupported` arm now also checks the `object` — previously it didn't, which is what let `unsupported` sources slip into the recursion loop #1421/#1427 had to guard.
- **Shared emitter** (`packages/jsx/src/adapters/parsed-expr-emitter.ts`): `arrayLiteral(elements, emit)` joins the `ParsedExprEmitter` interface; the exhaustive `emitParsedExpr` switch forces every adapter to handle it.
- **Mojo adapter**: `[$a, $b]` Perl array refs, `.join(sep)` lowers to `join(sep, @{...})` (the default `${obj}->{join}` hash-lookup fallback would be invalid Perl).
- **Go adapter**: `arrayLiteral` records BF101 directly so the pre-#1443 refusal contract for `branch-local-filter-join` stays intact. Go's own Slot fix is the next PR in the stack.
- The #1421 recursion guard stays as defence in depth; the specific shape it covered no longer reaches the loop because the parser succeeds cleanly.

## Stack

This is **PR 1/N** in a stack against #1443. Follow-ups (each as a separate PR on top of this one):

- PR2: Go side of the same Slot chain
- PR3: destructured filter param + `function`-keyword filter (both adapters)
- PR4: nested higher-order in filter predicate
- PR5 (optional, needs design): `reduce` / `forEach` / `flatMap`

## Test plan

- [x] `bun test packages/jsx/` (+3 new tests for `.filter(Boolean)` and array-literal parsing)
- [x] `bun test packages/adapter-mojolicious/` (all 176 pass; 4 skipped because Perl/Mojolicious is not installed in the sandbox)
- [x] `bun test packages/adapter-go-template/` (200 pass — Go-side refusal for `branch-local-filter-join` preserved)
- [x] `bun test packages/adapter-tests/` (345 pass — cross-adapter conformance)
- [x] `tsc --noEmit` clean for `packages/jsx`, `packages/adapter-mojolicious`, `packages/adapter-go-template`
- [ ] Manual: scaffold a mojo app via `npm create barefootjs@latest --adapter mojo` and confirm the registry `<Button>` renders without BF101 (sandbox doesn't have perl)

The 4 pre-existing failures (3 alias-fidelity tests in `packages/jsx`, 4 context-provider tests in `packages/client`) are unchanged by this PR — verified by stashing and running the same suites against base.

---
_Generated by [Claude Code](https://claude.ai/code/session_013cEtcSYt37CdRguejqFeAt)_


packages/adapter-mojolicious/src/adapter/mojo-adapter.ts
```diff
@@ -1368,6 +1369,33 @@ export class MojoAdapter extends BaseAdapter implements IRNodeEmitter<MojoRender
...
+    case 'join': {
+      // arr.join(sep) → join(sep, @{arr}). The default `${obj}->{join}`
+      // hash-lookup fallback would emit invalid Perl, which is why the
+      // IR carves out a dedicated method node instead of routing
+      // through the generic call dispatcher.
+      const obj = emit(object)
+      const sep = emit(args[0])
+      return `join(${sep}, @{${obj}})`
+    }
+  }
```
kfly8, line 1408: [imo]What happens in non-join cases? I want to avoid silence.

packages/adapter-mojolicious/src/adapter/mojo-adapter.ts
```diff
@@ -1368,6 +1369,33 @@ export class MojoAdapter extends BaseAdapter implements IRNodeEmitter<MojoRender
...
 
+/**
+ * Lowering for `array-method` IR nodes (#1443) — shared between the
+ * filter and top-level emitters so the Embedded Perl form stays
+ * consistent regardless of which context the chain lands in. The
+ * exhaustive switch on `method` makes future additions (e.g.
+ * `.concat`, `.slice`) a TS compile error here, mirroring the drift
+ * defence on `ParsedExpr.kind`.
+ */
+function renderMojoArrayMethod(
```
kfly8: I don't think it needs to have "mojo" on it.

packages/adapter-mojolicious/src/adapter/mojo-adapter.ts
```diff
@@ -1570,6 +1627,24 @@ class MojoTopLevelEmitter implements ParsedExprEmitter {
     return arrayExpr
   }
 
+  arrayLiteral(elements: ParsedExpr[], emit: (e: ParsedExpr) => string): string {
+    // Perl array ref. Identifiers inside elements resolve through the
+    // top-level emitter so `[className, childClass]` becomes
+    // `[$className, $childClass]` (the registry Slot's chain in
+    // #1443). Empty `[]` stays as `[]` — a valid empty Perl array
+    // ref that grep/join handle naturally.
+    return `[${elements.map(emit).join(', ')}]`
```
kfly8, line 1648: What you're doing is too specific for the name.

PR #1456, opened 2026-05-21
fix(adapter-hono/#1453): land Hono SSR template at zero tsc errors

### LINKED ISSUE #1453: Hono adapter SSR template codegen leaks tsc errors into the scaffold's `public/components/ui/{button,slot}/index.tsx`
## Summary

After `npm create barefootjs@latest`, `tsc --noEmit` (or any IDE) raises ~12 errors in the generated `public/components/ui/{button,slot}/index.tsx`. Runtime is unaffected (esbuild strips `import type`), but the errors show up red across the editor surface from the first scaffold and undermine confidence at onboarding.

Surfaced in PR #1450. The user-authored side is now tsc-clean; this issue covers the residual generated-file errors.

## Repro

```sh
npm create barefootjs@latest --yes
cd my-app && npm install
npx tsc --noEmit
# → ~12 errors in public/components/ui/{button,slot}/index.tsx
```

## Three codegen gaps in `packages/adapter-hono/src/adapter/hono-adapter.ts`

1. **Relative imports don't survive the `public/` depth shift.** Registry source carries `import type { Child } from '../../../types'`, which resolves to `<root>/types/` from `components/ui/button/`. The compiler copies the import verbatim into `public/components/ui/button/index.tsx`, where the same path now resolves to the non-existent `<root>/public/types/`. Fix: rewrite by depth delta, or via `@/`-alias.

2. **Source-local type aliases are dropped from emit.** The Button source declares `type ButtonVariant`, `type ButtonSize`, `interface ButtonProps` and re-exports them. The generated file keeps the re-export AND references `ButtonProps` in `type ButtonPropsWithHydration = ButtonProps & {...}`, but the declarations themselves aren't carried forward → TS2304. Same for `SlotProps`. Fix: preserve referenced type declarations on emit (or rewrite refs to `import type` from the source).

3. **`variantClasses[variant]` raises TS7053.** The hydration wrapper widens `variant` / `size` to `any` at the lookup site, so `Record<ButtonVariant, string>[any]` errors. Fix: preserve the precise prop types past the destructure-after-spread rewrite.

## Why it matters

A user who opens a fresh scaffold in VS Code immediately sees ~12 red squiggles in files they didn't write and can't fix (regenerated on every `bf build`). Wrong first impression for an alpha. Closing this would get the Hono scaffold to a "zero tsc errors out of the box" guarantee.

## Scope

- Hono adapter only for now (other adapters use `dist/` and may differ — separate audit).
- `bf add`-produced source under `components/ui/<name>/` is already tsc-clean; bug is entirely on the emit side.

## Test plan

- Fixture under `packages/adapter-tests/fixtures/` asserting `tsc --noEmit` over a Button-shaped emit produces zero errors.
- Extend `hydration-props-type.test.ts` (added in #1450) for type-alias preservation + relative-import rewriting.

## References

- PR #1450 — narrowed the surface from ~23 to these residual ~12 errors and locked it with tests.

kfly8:
Closes #1453.

After `npm create barefootjs@latest` + `npx tsc --noEmit`, the scaffold's generated `public/components/ui/{button,slot}/index.tsx` raised ~12 errors — none in user-authored code, all in regenerated emit. This lands the Hono scaffold at the "zero tsc errors out of the box" guarantee called out in the issue.

## Three codegen gaps

**1. Relative imports didn't survive the `public/` depth shift.**
The user-authored `import type { Child } from '../../../types'` lives in `components/ui/button/index.tsx` and resolves to `<root>/types/`. The compiler copied it verbatim into `public/components/ui/button/index.tsx`, where the same path now points at the non-existent `public/types/` (TS2307).

`compileEntry` now post-processes every marked template through `rewriteRelativeImportsForOutput`, which re-anchors each `from '…'` / `import '…'` / `export … from '…'` relative path against the on-disk output dir:
- Component-to-component imports (`'../slot'`) stay the same because both ends are mirrored at matching depth.
- Non-component paths (`types/`, shared helpers) get re-relativised back to source.

**2. Source-local type declarations were dropped from emit.**
The Hono adapter's `generateTypes` filtered `metadata.typeDefinitions` to those whose names appear in `componentBody`. But the destructured-props branch synthesises `${Name}PropsWithHydration = ${propsTypeName} & {…}` *after* that scan, so the body literally never mentioned `ButtonProps` — its declaration was dropped (TS2304 on the synthesised alias), and `variant`/`size` then widened to `any` at every `Record[key]` lookup site (TS7053).

The seed set now also includes `propsTypeName` and every name listed in `metadata.namedExports` re-export blocks, so the transitive closure pulls in `ButtonProps → ButtonVariant / ButtonSize` and `export type { ButtonVariant, ButtonSize, ButtonProps }` points at concrete local declarations.

**3. Local-function predicates and branch-local `as` casts were stripped.**
Slot's `function isValidElement(element: unknown): element is { tag: unknown; props: Record<string, unknown> }` collapsed to `function isValidElement(element)`, and `const Tag = children.tag as any` collapsed to `const Tag = children.tag`. Both lost their type information because the analyzer rebuilt the signature from `ParamInfo` (which can't distinguish explicit `:unknown` from a missing annotation) and routed scope-variable initializers through `ctx.getJS` (which strips type casts).

Added source-verbatim fields:
- `FunctionInfo.typedParams` / `typedReturnType` — populated from the AST, consumed by `jsx-adapter` and `module-exports`.
- `IRIfStatement.scopeVariables[].typedInitializer` — populated in `jsx-to-ir`, consumed by `hono-adapter`'s `renderIfStatement`.

Each new field follows the existing `typedBody` / `typedValue` / `typedInitialValue` / `typedComputation` convention.

## End-to-end

`bf build` against a Button + Slot scaffold now produces a tsc-clean `public/components/`:
```sh
$ bf build && tsc --noEmit
$ echo $?
0
```

## Test plan

- [x] `packages/cli/src/__tests__/build.test.ts` — 6 unit tests for `rewriteRelativeImportsForOutput`: non-component depth shift, sibling-component preservation, bare specifiers untouched, `import type` / side-effect `import` / `export … from` forms, single vs. double quote preservation.
- [x] `packages/adapter-hono/src/__tests__/hydration-props-type.test.ts` — 3 new tests for type-alias preservation: `propsTypeName` seed, transitive closure through interface bodies, named re-exports.
- [x] Existing `hydration-props-type.test.ts` cases still pass (no regression).
- [x] Full `bun test packages/jsx packages/adapter-hono packages/cli` — 2455 pass; the 4 remaining failures are pre-existing on this branch (`./runtimes.generated` is a generated file the sandbox doesn't have).
- [x] End-to-end: `bf build` on a Button + Slot project → `tsc --noEmit` exits 0.

---
_Generated by [Claude Code](https://claude.ai/code/session_012rR93Fc92QGQCZ3VnXbupq)_

### DIFF
--- a/packages/cli/src/lib/build.ts
+++ b/packages/cli/src/lib/build.ts
@@ -924,9 +924,10 @@ export function effectiveNamesFor(
 }
 
 /**
- * Re-anchor every relative `from '…'` / side-effect `import '…'` in a
- * marked-template's source so the same import points at the same file
- * after the template is written to its destination on disk.
+ * Build a relative-import rewriter for a single source-file → emit-file
+ * pair. The returned function takes a relative module specifier (as
+ * written in the source) and returns the same specifier re-anchored
+ * to the emit's on-disk dir.
  *
  * `bf build` mirrors `<componentDir>/<rel>/index.tsx` to
  * `<templatesOutDir>/<rel>/index.tsx`. Crucially, only files under a
@@ -938,7 +939,7 @@ export function effectiveNamesFor(
  * resolves to the non-existent `public/types/` and tsc raises TS2307
  * across the scaffold (#1453).
  *
- * For each relative import:
+ * For each relative specifier:
  *   1. Resolve against the SOURCE file's directory → `srcAbs`.
  *   2. If `srcAbs` is under any componentDir, the build emits a mirror
  *      at `<templatesOutDir>/<rel-under-componentDir>` — point the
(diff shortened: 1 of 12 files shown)


packages/cli/src/lib/build.ts
```diff
@@ -923,6 +923,77 @@ export function effectiveNamesFor(
...
+    }
+    return srcAbs
+  }
+
+  // `\b(?:from|import)\s+` matches the keyword introducing a string
+  // specifier in: `import X from '...'`, `import type X from '...'`,
+  // `export { X } from '...'`, and side-effect `import '...'`. The
+  // capture group keeps the specifier's quote style intact.
+  return content.replace(
+    /(\b(?:from|import)\s+)(['"])(\.[^'"\n]+)\2/g,
```
kfly8: [imo] Isn't that a bit fragile?

PR #1687, opened 2026-05-31
bench(perf): deep-tree re-render baseline for one reactive leaf (#1373)

### LINKED ISSUE #1373: Perf: re-render cost on deep tree (depth 50+) with one reactive leaf
Parent: #1244 (Performance / scale)

## Goal

Measure re-render cost when a tree at depth 50+ has one reactive leaf. Establish current baseline; identify pathological growth before scaling problems land in user code.

## Steps

- Generate a recursive component tree of depth 50, 100, 200 with a single signal at the leaf.
- Measure:
  - mount cost
  - single-signal-update propagation time to the leaf
  - GC behaviour after unmount
- Compare against a shallow tree with the same total node count to isolate depth-specific cost.

## DoD

- Benchmark fixture under `bun bench`.
- Numbers recorded as a baseline (this issue body or `docs/perf.md`).
- If pathological growth is found, a follow-up fix issue.

## Outcome

Investigated in #1687 (closed without merging — no fix to land). Baseline numbers and full findings are in [this comment](https://github.com/piconic-ai/barefootjs/issues/1373#issuecomment-4585848297).

**No pathological growth.** Update propagation to the leaf is O(1) in depth (fine-grained reactivity notifies only the leaf signal's subscriber set, never the owner tree); mount is linear in node count with no depth-specific term; the subtree is GC'd cleanly after unmount. No follow-up fix issue required. (Out-of-scope note: recursive disposal is call-stack-bounded at ~10k–20k nesting, two orders of magnitude beyond this issue's 50–200 target.)

kfly8:
Closes #1373.

## What

Establishes the perf baseline requested by #1373: re-render cost of a deeply nested component tree (depth 50/100/200) that has a single reactive leaf.

- `benchmarks/deep-tree.ts` — fixture. Nested `createRoot` scopes (each models a component boundary) wrapping real DOM nodes, with one `createSignal` + `createEffect` at the deepest leaf. Measures mount, single-signal update propagation to the leaf, deep-vs-shallow at equal node count, and GC after unmount (WeakRef).
- `bun bench` script added to the root `package.json` (runs `reactive.ts` + `deep-tree.ts`).

The baseline numbers are recorded in the [#1373 issue comment](https://github.com/piconic-ai/barefootjs/issues/1373#issuecomment-4585848297). A committed `docs/perf.md` was intentionally dropped — the findings will be vetted more carefully before being turned into documentation.

## Findings (DoD)

| Metric | depth 50 | depth 100 | depth 200 |
|---|---|---|---|
| mount | ~0.19 ms | ~0.13 ms | ~0.21 ms |
| update → leaf | ~0.0017 ms | ~0.0010 ms | ~0.0008 ms |
| effect runs / update | 1.00 | 1.00 | 1.00 |

- **Update propagation is O(1) in depth, not O(depth).** Fine-grained reactivity notifies only the leaf signal's subscriber set — never the owner tree. A leaf update at depth 200 runs exactly one effect, same cost as a shallow tree.
- **Mount is linear in node count** with no depth-specific term (deep ≈ shallow at equal node count).
- **No leak after unmount** — root and leaf nodes are GC'd after `dispose()` + `Bun.gc()`.

**No pathological growth in the 50–200 range → no follow-up fix issue needed.**

### Known boundary (out of scope)

`disposeSubtree` / the mount walk are recursive, so reactive-scope nesting is bounded by the JS call stack: ~10k nesting is fine, ~20k throws `Maximum call stack size exceeded` — two orders of magnitude beyond the issue's target. Converting disposal to an explicit work-list would remove the limit if ever needed.

https://claude.ai/code/session_01Li59k6bpAzaYp85dN6aWjC


kfly8:
Closing without merging. The investigation confirmed no pathological growth (see #1373), so there's no fix to land — the benchmark fixture and `bun bench` script don't need to ship as part of this. Baseline numbers are recorded in the #1373 comment.

---

Issue #1448, opened 2026-05-20
Catalog: JS array &amp; string methods that don&apos;t yet lower to Mojo/Go template syntax


kfly8:
Tracking the gap between JS built-in `Array.prototype` / `String.prototype` methods and what the Mojo + Go template adapters can lower at SSR time. Surfaces during #1443 follow-up: once the registry `` chain landed, the natural next question is "which other JS-builtin shapes can users reach for without `/* @client */`?"

The Hono / CSR adapters evaluate JSX at runtime (full JS available), so they never hit these gates. The catalogue below is specifically about what reaches the **template-language** adapters via `ParsedExpr` lowering.

## Currently lowered

| Layer | Methods | IR kind |
|---|---|---|
| Higher-order on array | `filter`, `every`, `some`, `find`, `findIndex`, `findLast`, `findLastIndex` | `higher-order` |
| Non-higher-order array method | `join`, `includes`, `indexOf`, `lastIndexOf`, `at`, `concat`, `slice`, `reverse`, `toReversed` | `array-method` (#1443, #1457–#1462) |
| Sort method (structured comparator) | `sort`, `toSorted` | `array-method` + `SortComparator` (#1472, #1661) |
| Reduce fold (structured op) | `reduce`, `reduceRight` | `array-method` + `ReduceOp` (#1728, #1729) |
| Iteration shapes | `entries`, `keys`, `values` | `IRLoop.iterationShape` (#1600) |
| Non-higher-order string method | `includes`, `toLowerCase`, `toUpperCase`, `trim`, `split`, `startsWith`, `endsWith`, `replace`, `repeat`, `padStart`, `padEnd` | `array-method` (#1457, #1463–#1465, #1717–#1721) |
| Identifier-path callee | `JSON.stringify`, `String`, `Number`, `Math.floor`/`ceil`/`round` | `templatePrimitives` registry (#1188 / #1189) |

`.filter(Boolean)` is recognised in the parser and synthesised into the existing `higher-order` form (#1443). Destructured / function-keyword filter callbacks are also normalised at parse time (#1443 PR3).

## Unsupported array methods

Grouped by feasibility (rough — each entry needs its own design pass before landing).

### Tier A — Lowerable with mechanical work ✅ Complete

[...]

PR #1728, opened 2026-06-03
Lower .reduce(fn, init) arithmetic-fold catalogue (#1448 Tier C)

kfly8:
## Summary

First Tier C landing for #1448. `.reduce(fn, init)` — the only Tier C method with measurable demand in the codebase (`playlist.reduce((s, t) => s + t.duration, 0)`, view-count / visitor sums in the demo components) — now lowers on the template-language adapters. The shape catalogue mirrors the `.sort` precedent: a structured `ReduceOp` extracted once at parse time, then consumed by both adapters and a single runtime helper.

### Catalogue (lowered)

| Shape | Result |
|---|---|
| `arr.reduce((acc, x) => acc + x, 0)` | numeric sum over self |
| `arr.reduce((acc, x) => acc + x.field, 0)` | numeric sum over a struct field |
| `arr.reduce((acc, x) => acc * x.field, 1)` | numeric product |
| `arr.reduce((acc, x) => acc + x.field, '')` | string concatenation (string init flips `+` to concat) |
| single-`return` block body | unwrapped to the returned expression |

The accumulator must be the binary expression's left operand (`acc + x`, not `x + acc`), the per-item operand must be the item param or a single non-computed field access on it, and the init must be a number or string literal (negative numbers via prefix `-` allowed). Subtraction / division, deep field access, object-building reducers (`{...acc, [x.id]: x}`), 3- / 4-param reducer forms, and `.reduce(fn)` without an initial value all refuse with **BF101** + the `@client` escape hatch. `.reduceRight` stays refused entirely — the same helper can grow a `direction` arg later if non-commutative folds surface in practice.

### Pipeline

1. **Parser** (`expression-parser.ts`) — new `extractReduceOpFromTS` matches the catalogue from the raw TS AST (two-param arrow ⇒ standard `convertNode` rejects it). Off-catalogue 2-arg calls return an explicit `unsupported` with the accepted-shape catalogue in the message; `reduce` stays in `UNSUPPORTED_METHODS` so the no-init fall-through still refuses loudly.
2. **Emitter interface** (`adapters/parsed-expr-emitter.ts`) — new `reduceMethod()` arm. Adding it makes every `ParsedExprEmitter` implementer surface a TS compile error until they handle it (same drift defence sort uses).
3. **Go adapter** — `emitBfReduce(recv, op)` emits `bf_reduce <recv> "<op>" "<keyKind>" "<keyName>" "<type>" "<init>"`. Runtime `Reduce(items, op, keyKind, keyName, type, init)` folds left-to-right; numeric folds use float64, string folds reuse the existing `toString` rule.
4. **Mojo adapter** — `renderReduceMethod(recv, op)` emits `bf->reduce($recv, { op => …, key_kind => …, key => …, type => …, init => … })`. Perl `sub reduce` mirrors the Go semantics; non-numeric keys coalesce to 0 (numeric) / '' (concat) so the two stay byte-equal.

### Cross-adapter byte-equality

Four new conformance fixtures (`reduce-sum-field`, `reduce-sum-self`, `reduce-concat`, `reduce-product`) verified end-to-end through `go run` against the expected HTML. The same `expectedHtml` is shared across Hono / CSR / Go / Mojo adapters via the existing conformance harness.

**Known caveat** (documented in `spec/compiler.md` alongside the `bf_sort` `auto` divergence): float stringification can diverge from JS for inexact binary fractions (`0.1 + 0.2`); genuine integer sums — the common SSR case — agree across all three adapters.

### Spec

`spec/compiler.md` gains a "Reduce folds" subsection adjacent to "Sort comparators" listing the accepted catalogue and refusal conditions.

## Test plan

- [x] `bun test packages/jsx/src/__tests__/reduce-op.test.ts` — 14 parser-catalogue tests (positive shapes + 7 refusal shapes)
- [x] `bun test packages/jsx/src/__tests__/ir-reduce-op.test.ts` — IR survives the analyzer→IR boundary with the structured fold intact
- [x] `bun test packages/adapter-go-template/src/__tests__/` — 9 Go reduce tests pass (4 conformance fixtures end-to-end via `go run` + the existing `reduce (no init)` BF101 mirror)
- [x] `bun test packages/adapter-mojolicious/src/__tests__/` — 15 Mojo reduce tests pass (4 emit-shape pins + adapter-side compile path; Perl runtime exercised by CI)
- [x] `bun test packages/jsx/src/__tests__/` — 1840 pass, 4 pre-existing failures on `main` (reactive-primitive-resolver alias migration — unrelated)
- [x] `tsgo --noEmit` clean across jsx / adapter-go-template / adapter-mojolicious
- [x] `go test ./...` clean for the Go runtime

---
_Generated by [Claude Code](https://claude.ai/code/session_016hsRA9RLzvMgq2YgghQULe)_


kfly8:
Thanks for the review — all five findings were real cross-adapter byte-equality gaps in edge-case inits. Addressed in de5ffd0:

**Numeric init `1_000` / `0x10` / `0b10`** (classifyReduceInit + Go `Reduce`): the parser now stores TypeScript's canonical decimal form (`node.text`) rather than the raw source — `1_000`→`1000`, `0x10`→`16`, `1e3`→`1000` — so Go's `strconv.ParseFloat` and Perl fold identically instead of Go silently parsing to `0`.

**String init escapes** (both adapters' init slicing): the parser now stores the *decoded* contents (`node.text`), and refuses any string literal carrying an escape (`'\n'`, `'\u{…}'`, an escaped quote) at parse time (→ BF101 + `@client`). Accepted seeds are therefore escape-free single-line strings that embed byte-equal in both the Go-template `"…"` operand and the Perl single-quoted literal with no per-target unescaping. Round-trip emitters re-quote a string seed via `JSON.stringify`. The realistic concat seed (`''`) is unaffected.

**Go field-name initialisms** (`emitBfReduce` `capitalize(field)`): the module-level `capitalize` now applies the same Go-initialism rules as the adapter's `capitalizeFieldName` (`id`→`ID`), so the reflect lookup resolves the generated exported struct field instead of folding zero. This also corrects the pre-existing `emitBfSort` path. Additionally, the runtime `getFieldValue` gained an all-lowercase map fallback so JSON-decoded `map[string]any` keyed by the lowercase JS name (`"id"`) still resolves an all-caps projected key.

New coverage: Go runtime `Reduce` tests (incl. the initialism-on-lowercase-map case), a Go emit pin asserting the key is `ID` not `Id`, and parser tests for numeric normalization, decoded string seeds, and escape refusal.

---

PR #1814, opened 2026-06-07
feat(adapters): lower style={{…}} object literals to CSS (expectedDiagnostics audit)

kfly8:
**Stacked on #1813** (base = `claude/expected-diagnostics-audit`).

## Increment: `style-object-dynamic` / `style-3-signals` — ✅ Go + Mojolicious + Xslate

A JSX `style={{ backgroundColor: color, padding: '8px' }}` object literal was pinned to **BF101** on all three SSR adapters — a bare object literal in attribute position had no template form. It now lowers to a CSS string.

New shared **`parseStyleObjectEntries`** (`@barefootjs/jsx`):
- parses the object literal (wrapped in parens to force expression context, since a bare `{…}` parses as a block statement),
- kebab-cases each key (`backgroundColor` → `background-color`, `WebkitTransform` → `-webkit-transform`),
- classifies each value as a static string literal or a JS expression.

Each adapter assembles the CSS string, interpolating dynamic values in its own syntax:
- **Go**: `background-color:{{.Color}};padding:8px`
- **Mojo**: `background-color:<%= $color %>;padding:8px`
- **Xslate**: `background-color:<: $color :>;padding:8px`

Each value expression is pre-checked with `isSupported`, so an unsupported value — or an unsupported object shape (spread / shorthand / computed key) — keeps the existing BF101 refusal rather than emitting partial output. Dropped the two `style-*` entries from all three adapters' `expectedDiagnostics`.

## Test status
- Adapters (real Go 1.25.6 / Mojolicious 9.35 / Text::Xslate v3.5.9): **1301 pass / 27 skip / 0 fail**
- jsx: new `parseStyleObjectEntries` unit tests pass (the only jsx failures are the 4 pre-existing resolver-migration alias tests, present on base)
- `tsc` clean on all four packages

Hono reference snapshots unchanged.

https://claude.ai/code/session_01LaP8KV6yASUT37PzDMvYbW

---
_Generated by [Claude Code](https://claude.ai/code/session_01LaP8KV6yASUT37PzDMvYbW)_

PR #1818, opened 2026-06-07
feat(adapters): lower object-rest .map destructure (expectedDiagnostics audit)

kfly8:
**Stacked on #1814** (base = `claude/expected-diagnostics-audit-2`; chain: #1813 → #1814 → this).

## Increment: `rest-destructure-object-in-map` — ✅ Go + Mojolicious + Xslate

`tasks().map(({ id, title, ...rest }) => <li>{title}:{rest.flag}</li>)` (an object-rest param whose `rest` is read via member access) was pinned to **BF104** on all three SSR adapters. It now lowers — each binding resolves against a per-item loop variable:

- **Go**: `{{range $_, $bfItem := …}}`; the `rest` binding maps to the bare range var so the member emitter renders `rest.flag` → `$bfItem.Flag` (via a new `loopBindingStack`).
- **Mojo**: one Perl `my` local per binding (`my $rest = $bfItem;` so `$rest->{flag}` resolves).
- **Xslate**: equivalent Kolon `: my` binding locals.

A shared supportability gate (`destructureBindingsSupportable`) inspects the IR's `paramBindings` and only accepts **simple `.field` paths + an object-rest read via member access**. The other three rest-destructure fixtures stay refused (BF104), because the SSR `range`/`for` can't express them inline:
- `rest-destructure-object-spread-in-map` (`{...rest}`) needs a residual object excluding the consumed keys,
- `rest-destructure-array-in-map` (`[a, ...t]`) needs index/slice,
- `rest-destructure-nested-in-map` (`{ cells: [h, ...r] }`) needs nested index paths.

## Test status
- Adapters (real Go 1.25.6 / Mojolicious 9.35 / Text::Xslate v3.5.9): **1301 pass / 27 skip / 0 fail**
- `tsc` clean on all three packages

Hono reference snapshots unchanged.

https://claude.ai/code/session_01LaP8KV6yASUT37PzDMvYbW

### DIFF
--- a/.changeset/rest-destructure-object.md
+++ b/.changeset/rest-destructure-object.md
@@ -1,4 +1,5 @@
 ---
+"@barefootjs/jsx": patch
 "@barefootjs/go-template": patch
 "@barefootjs/mojolicious": patch
 "@barefootjs/xslate": patch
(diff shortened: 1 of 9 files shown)


kfly8:
Thanks @copilot — all 9 points addressed in bbba7e6:

- **Gate centralized** in `@barefootjs/jsx` as `isLowerableObjectRestDestructure`, replacing the three duplicated per-adapter helpers. It now walks the **whole** loop subtree (elements, components, conditionals, async, providers, template-literal parts), so `{...rest}` spreads and bare value uses (`String(rest)`, `{rest}`, `fn(rest)`) under any node type are detected and refused. Member-access reads stay supported; `foo.rest` is not a false positive (lookbehind).
- **Go `identifier()` ordering**: binding resolution now runs before module-const inlining, so a binding whose name collides with a module string const resolves to the loop item.
- **Go nested shadowing**: the synthetic range var is suffixed with nesting depth (`$bfItem0`, `$bfItem1`).
- **Mojo `loopBoundNames`**: the synthetic `bfItem` is now registered.
- **`.filter().map(destructure)`**: refused via the gate, so the filter-param rewrite never targets the destructure-pattern text (both Mojo and Xslate).

Adapter + jsx suites green against real Go 1.25.6 / Mojolicious 9.35 / Text::Xslate v3.5.9 (the 4 pre-existing `resolver migration … via checker` failures are unrelated — they fail on the base too).

---

PR #1889, opened 2026-06-11
feat(router): @barefootjs/router — partial-navigation client router (alpha)

kfly8:
## `@barefootjs/router` — automatic partial-navigation client router (alpha)

On a same-origin link click it swaps **only the `[bf-outlet]` content region** and re-hydrates the islands inside it, leaving the page shell mounted — Turbo-Drive/Frames-style, but the swap + re-hydration reuse the BarefootJS runtime instead of a new framework.

> Built on **#1893** (`@barefootjs/client` subtree-scoped hydrate + disposal), **merged to main**. This PR is rebased onto main; its diff is router-only. The reference app + design exploration are in **#1891** (stacked on this).

## What's here

- **Outlet-scoped swap** — intercept a same-origin `<a>`, fetch the full page, extract `[bf-outlet]` client-side, swap it, keep the shell mounted. **Backend-agnostic, no content-negotiation header** (a server fragment was considered and dropped: it only shaves gzip-compressible shell markup while hurting cache efficiency). Re-hydration uses the runtime's subtree-scoped `window.__bf_hydrate_within` (O(outlet)); disposal uses `window.__bf_dispose_within`. Both fall back gracefully if absent.
- **Module-aware swap** — loads a navigated-to island's JS (the response's same-origin `<script type="module">`) before re-hydrating, so an island whose module wasn't on the first page comes alive.
- **Last-wins navigation** — rapid clicks resolve to the latest target (abort guard after each await).
- **Prefetch** on hover (dwell), focus, and primary press (`pointerdown`), with `modulepreload` of island modules. Best-effort like Next.js — a failed prefetch isn't cached.
- **Snapshot cache** — stale-while-revalidate with a jittered refresh window and LRU eviction (`cacheFreshMs` / `cacheStaleMs` / `cacheCap`), making back/forward instant.
- **`searchParams()`** (`@barefootjs/router/signals`) — a reactive read of the URL query, branded `Reactive<…>` so the *existing* compiler reactivity analysis drives fine-grained island updates (no new compiler feature). A same-route query-only navigation updates the signal + URL **without swapping the outlet**; a pathname change still swaps.

## Package structure

The controller is split into focused modules so the package reads cleanly:

| Module | Responsibility |
|---|---|
| `types.ts` | shared interfaces (no logic) |
| `seams.ts` | `window.*` bridges to the optional client runtime |
| `cache.ts` | snapshot cache (stale-while-revalidate + LRU) |
| `outlet.ts` | read the response: extract the outlet + island modules |
| `router.ts` | the controller wiring those together |
| `search-params.ts` | the `searchParams()` environment signal |

## Notes

- **Alpha**: the core is in place; the API may still shift before 1.0.
- `@barefootjs/client` is an **optional peer** — the router core only touches it through `window.*` seams.
- 17 unit tests (happy-dom): outlet swap, last-wins, module-aware swap, prefetch (hover/focus/press), SWR + LRU cache, searchParams (no-swap query nav, popstate), non-intercept cases.

https://claude.ai/code/session_01SAkUPtiKEw2m6xKggAjNj2


kfly8:
Close to rethink it.

PR #1891, opened 2026-06-11
docs(router): router-blog reference app (stacked on #1889, not for merge)

kfly8:
## Reference implementation + design exploration — not for merge

A small blog that uses **`@barefootjs/router`** (PR #1889) for **automatic partial navigation**, grown into a **directional stress test** and a **design exploration**. This is the top of a 3-PR stack: **#1893** (`@barefootjs/client` subtree hydrate/dispose — the runtime capability) → **#1889** (router prototype) → **#1891** (this).

## Design exploration — IR-driven router ([`DESIGN.md`](https://github.com/piconic-ai/barefootjs/blob/claude/loving-turing-aatkma-blog/integrations/router-blog/DESIGN.md))

Goal: strong navigation performance by playing to what BarefootJS is built around — **moving work from runtime into the compiler/IR** — and **auto-wiring** what htmx-style routers make you wire by hand. Next/TanStack/Inertia/Turbo/htmx are references we learn from; the aim is a different mechanism that suits BarefootJS, with respect for the prior art.

`bench.ts` quantifies the runtime "force" the prototype pays per navigation (happy-dom, avg 200, growing shell, fixed 8-island outlet):

| shell nodes | full-doc walk (current rehydrate) | outlet-only walk (IR-scoped) | full parse | fragment parse |
|---|---|---|---|---|
| 50 | 0.125ms | 0.013ms | 2.84ms | 0.17ms |
| 1000 | 2.43ms | 0.009ms | 22.2ms | 0.16ms |
| 4000 | 9.53ms | **0.010ms** | 92.1ms | **0.13ms** |

Re-hydration (`rehydrateAll` walks the **whole document**) and the client parse both scale O(document); an IR-scoped outlet walk / fragment parse stay flat. The design (grounded in verified codebase anchors) proposes a **manifest-interpreter router**, **compiler-derived outlet + persistent islands** (the auto-wiring win), **signal-level data patches** for data-only navigations (borrowing Inertia's `only:` idea), and **CLI integration** (`routes.manifest.json` via the existing `postBuild` hook; a `bf routes` inspector).

## P0 — extracted to its own PR (#1893)

The two `@barefootjs/client` unlocks the design rests on are now an **independent PR**: subtree-scoped `rehydrateScope(root)` (O(outlet)) + `createRoot`-keyed `disposeScope(root)`. A general runtime capability (router/conditionals/loops/streaming all benefit), not router-specific. The router (#1889) consumes them and the disposal leak is fixed by construction.

[...]

PR #1910, opened 2026-06-14
feat(router): ship backend-agnostic partial navigation router

kfly8:
### Motivation
- Provide a production-ready client router that partial-updates a marked content region while keeping the shell mounted and reusing the BarefootJS runtime, instead of requiring a custom navigation protocol.
- Remove any `X-Barefoot-Navigate`/protocol-based approach so the router works with ordinary full-page HTML responses from any backend.
- Structure the implementation for clarity and release: a small public surface and separated internals (cache, outlet parsing, runtime seams, search-params signal).

### Description
- Add a new package `@barefootjs/router` exposing `startRouter`, `navigate`, and the optional `@barefootjs/router/signals` entry; internals split into `cache.ts`, `outlet.ts`, `router.ts`, `seams.ts`, `search-params.ts`, and `types.ts`.
- Remove/avoid any special navigation header or `X-Barefoot-Navigate` concept: the router fetches the full document and extracts the `[bf-outlet]` subtree client-side, and `BF_OUTLET` is added to `@barefootjs/shared` as the swappable marker.
- Implement prefetch (hover/focus/pointerdown) with `modulepreload`, an LRU stale-while-revalidate snapshot cache, module-aware import-before-hydrate, last-wins navigation, and a reactive `searchParams()` signal that can opt-in to avoid full swaps on query-only navigations.
- Address review feedback: query-only navigation now supersedes an older in-flight swap; disposal is awaited and falls back to `@barefootjs/client/runtime`; SSR, singleton-module, inline-script, focus/a11y, and modulepreload lifecycle limitations are documented.
- Clarify the router vision and outlet contract: the single outlet is a broad-to-narrow spectrum, so document-heavy sites can get Turbo-like navigation while apps with shared UI can preserve more shell continuity without adopting an SPA protocol.
- Add `@barefootjs/client` as an explicit `^0.11.0` development dependency for the signals entry. This intentionally avoids `workspace:*` because of the previous Changesets release-version incident, while the published optional peer remains `>=0.11.0`.
- Add a runnable blog example demonstrating progressive enhancement and backend independence, plus unit tests and a small contract test ensuring the router issues ordinary HTML requests without a custom header.

### Testing
- `bun test packages/router/__tests__` — 19 passed.
- `bun run --filter @barefootjs/router build` — passed.
- `bunx biome check packages/router` — passed.
- `git diff --check` — passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2dd210631083209b88483537dca790)


kfly8:
## Feedback from building a real-component blog on this router

I built a blog example on this branch where **every island is a real compiled `"use client"` component hydrated by the real `@barefootjs/client` runtime** (no hand-written stand-in like #1891). Reference app + 17-assertion Playwright verification (all green): **#1911** (`integrations/router-blog/`, stacked on this branch). The good news first: **`searchParams()` reactivity wired by the compiler out of the box** — an island that imports `searchParams` from `@barefootjs/router/signals` and reads it in a `createMemo` gets correct fine-grained DOM updates with **zero outlet swap** on `?sort=`/`?tag=` links. The `Reactive<>` brand approach works end-to-end. 🎉

Four integration gaps I hit, roughly in priority order:

### 1. The runtime seams aren't auto-installed; `setupStreaming()` is an undocumented prerequisite
`defaultRehydrate`/`defaultDispose` (`seams.ts`) read `window.__bf_hydrate_within` / `window.__bf_dispose_within`, but those are only set by `@barefootjs/client`'s `setupStreaming()` (`runtime/streaming.ts`). Nothing calls it by default — the compiled `hydrate()` path schedules its own walk but never installs the seams. So following the README's **`startRouter()`-only** usage gives you a router whose **disposal is a silent no-op** and whose re-hydration falls through to the dynamic-import fallback. I had to add a bootstrap that calls `setupStreaming()` before `startRouter()`. Suggestion: document this, or have `startRouter()` install the seams / have the client auto-install them.

### 2. `defaultDispose` has no fallback (unlike `defaultRehydrate`)
`defaultRehydrate` degrades `__bf_hydrate_within` → `__bf_hydrate` → `import('@barefootjs/client/runtime').rehydrateScope`. `defaultDispose` only does `window.__bf_dispose_within?.(outlet)` and silently no-ops if the seam is missing. Combined with #1, a developer who skips `setupStreaming()` gets limping-but-working re-hydration and **completely silent island leaks** — exactly the "disposal is the hard part" failure the design calls out. It should share the same dynamic-import fallback to `disposeScope`.

### 3. `searchParams()` has no SSR request-awareness, and the module-global signal is unsafe under concurrent SSR
On the server `searchParams()` returns `''` (the code comment admits this), so an island SSR'd for `/?tag=perf` renders the **unfiltered** list and only corrects after hydration — a flash and a likely hydration mismatch. To get correct SSR I import `setSearch` server-side and prime it from `c.req.url` before each render. But `searchString` is a **module-level signal shared across all requests in the process**, so per-request priming **races under concurrency** (request B clobbers request A mid-render). The "environment signal" needs request-scoped SSR state (via the adapter request context), not a module global. This is the biggest gap for the headline feature.

### 4. `@barefootjs/router/signals` must be a singleton module, or reactivity silently breaks
For `searchParams()` to drive an island, the router core (which calls `window.__bf_set_search`) and the island (which reads `searchParams()`) must share **one** instance of the signals module **and** the same reactive runtime. With standard `bf build` output the island imports `@barefootjs/router/signals` as a bare specifier, so it must be import-mapped to a **single** served bundle that keeps `@barefootjs/client/reactive` external (→ the same `barefoot.js`). Bundle it twice (e.g. once inside the bootstrap, once for the island) and you get two `searchString` signals — reactivity dies with no error. I worked around it with a dedicated `router-signals.js` bundle + import-map entry. A docs note ("signals must be a singleton") or a runtime warn when `__bf_set_search` is overwritten by a second instance would save real debugging time.

### Minor
- PR description mentions a `packages/router/example` blog and `packages/router/__tests__/blog-example.test.ts`, but neither is in the branch (only `router.test.ts`), and the local run is **18** tests, not 19. Looks like the example/test got moved out (`refactor(router): move blog example out of package PR`) and the description wasn't updated.
- `BF_OUTLET` is re-exported nicely, but the `<main bf-outlet>` marker is authored by hand — a typo just makes the router silently hard-navigate (only signal is a full reload). A tiny marker helper/component would harden that.

Happy to fold a minimal `setupStreaming()` mention + the `defaultDispose` fallback into this PR if useful. Full write-up in `integrations/router-blog/README.md` on #1911.

---

packages/router/src/search-params.ts
```diff
@@ -0,0 +1,54 @@
...
+ * ↔ `location`), so the value agrees across hydration with no flash and no
+ * seed-passing. On the server (no `window`) the initial value is `''`; an
+ * adapter-side SSR impl that reads the request is a follow-up.
+ */
+import { createSignal } from '@barefootjs/client/reactive'
+
+type Reactive<T> = T & { readonly __reactive: true }
+
+const [searchString, setSearchString] = createSignal(
+  typeof window !== 'undefined' ? window.location.search : '',
```
kfly8, line 30: **[medium] SSRに関する記述が実装と矛盾しています。**

このシグナルはクライアントでは `window.location.search`、サーバーでは（`window` がないため）`''` で初期化されます。そのため `/list?sort=price` を**直接ロード**した場合、サーバーは空のパラメータで描画し、クライアントは `sort=price` でハイドレートするので、`searchParams()` にバインドしたアイランドで実際にハイドレーションの不一致／フラッシュが発生します。

doc コメントの「server and client read the same source … so the value agrees across hydration with no flash」という表現は、この直接ロードのケースでは過大です（ファイル冒頭の docstring と README も同様）。挙動自体は既知の制約（adapter 側 SSR 実装は follow-up）として妥当ですが、「クライアント側のクエリナビゲーションでは一致するが、クエリ付き直接ロードは SSR seam が入るまでフラッシュする」程度に表現を弱めることを推奨します。

---

packages/router/src/router.ts
```diff
@@ -0,0 +1,334 @@
...
+  // isn't in use, fall through to a normal swap (legacy behavior).
+  if (
+    target.pathname === window.location.pathname &&
+    target.search !== window.location.search &&
+    setSearchSeam(target.search)
+  ) {
+    if (mode === 'push') window.history.pushState({ bfRouter: true }, '', target.href)
+    else if (mode === 'replace') window.history.replaceState({ bfRouter: true }, '', target.href)
+    return
+  }
```
kfly8, line 129: **[low] クエリのみのショートサーキットが、実行中のナビゲーションを上書きしません。**

この分岐は後段の `state.inflight?.abort()` より前に `return` します。outlet スワップが実行中にクエリのみのナビが発生すると、クエリ更新（signal + URL）は即時適用されますが、古いフェッチがその後に解決して `replaceChildren` までコミットし得るため、URL/コンテンツを上書きする可能性があります。

狭いレースですが、このクエリのみ経路に入る前に `state.inflight?.abort()` を呼ぶ（またはガードする）だけで防げます。

---

packages/router/src/router.ts
```diff
@@ -0,0 +1,334 @@
...
+    const current = document.querySelector(state.outletSelector)
+    if (!content || !current) {
+      // Target isn't part of this shell — fall back to a full load.
+      hardNavigate(finalUrl)
+      return
+    }
+
+    // Dispose the outgoing islands, then swap only the outlet.
+    state.dispose(current)
+    current.replaceChildren(...content.nodes)
```
kfly8, line 163: **[medium] スワップ後にフォーカス／アクセシビリティが管理されていません。**

`replaceChildren` でアウトレットを差し替えた後、フォーカスは失われ（切り離されたアンカーに残るか `<body>` に戻る）、`aria-live` によるルート変更の読み上げもありません。SPA ルーターでは定番の対応で、実際のアクセシビリティ上のギャップです。

最低限、README の「Limitations」にこの点を追記すること（現状スクロール復元のみ記載）、可能であればスワップ後にアウトレットへフォーカス移動＋ライブリージョンでの通知を検討してください。

---

packages/router/src/router.ts
```diff
@@ -0,0 +1,334 @@
...
+
+/**
+ * `modulepreload` the island modules a prefetched page carries: fetch +
+ * compile (no execute), so the click's `import()` runs instantly without
+ * a network wait — and without triggering hydration during the hover.
+ */
+function preloadModules(state: RouterState, html: string): void {
+  for (const src of extractOutlet(html, state.outletSelector)?.moduleSrcs ?? []) {
+    if (state.loadedModules.has(src) || state.preloaded.has(src)) continue
+    state.preloaded.add(src)
```
kfly8, line 225: **[low] `preloaded` と `modulepreload` リンクが無制限に増加します。**

`state.preloaded` セットと、`<head>` に追加される `<link rel="modulepreload">` ノードは、長いセッション中ずっと刈り取られません。ページキャッシュは LRU で上限がありますが、こちらにはありません。軽微なリークですが、上限を設けるか、対応するモジュール読み込み完了後にリンク要素を除去することを検討してください。

---

packages/router/src/outlet.ts
```diff
@@ -0,0 +1,60 @@
...
+ * the response has no outlet marker — that page belongs to a different shell,
+ * so the caller hard-navigates.
+ */
+export function extractOutlet(html: string, selector: string): OutletContent | null {
+  const doc = new DOMParser().parseFromString(html, 'text/html')
+  const outlet = doc.querySelector(selector)
+  if (!outlet) return null // not part of this shell → caller hard-navigates
+  const title = doc.querySelector('title')?.textContent ?? null
+  const moduleSrcs = [...collectModuleScripts(doc)]
+  const nodes = Array.from(outlet.childNodes).map((n) => document.importNode(n, true))
```
kfly8, line 40: **[low / docs] アウトレット内のインライン `<script>` は実行されません。**

`importNode` + `replaceChildren`（呼び出し側）ではインラインスクリプトは走らず、body 末尾の `<script type="module" src>` のみが `import()` されます。アイランドモデルとは整合しますが、コンテンツ領域にインライン初期化スクリプトを置く利用者にとっては落とし穴になりやすいです。

README の "How it reuses the runtime" テーブルに一文（「インラインスクリプトは差し替え時に実行されない」）を添えると親切です。

---

kfly8 (commented):
全体として設計がよく整理され（cache / outlet / seams / router / types の分割）、テストも充実した完成度の高い PR です。last-wins の `AbortController`、SWR + LRU キャッシュ、`window.*` seam によるオプショナルピア分離、同一オリジン限定のモジュール読み込みなど、安全側に倒した実装が好印象です。インラインで5件の指摘を付けました（medium 2 / low 3）。加えて1点、コード行に紐付かない指摘を以下に記載します。

**[medium] PR 説明文が、差分に存在しない成果物を記載しています。**

本文（Description / Testing）には次が記載されていますが、変更された15ファイルの中に含まれていません:
- 「a runnable blog example (`packages/router/example`)」
- 「an automated example test `packages/router/__tests__/blog-example.test.ts`」、および「executed 19 tests」

実際に存在するテストは `packages/router/__tests__/router.test.ts`（約18件）のみです。「カスタムヘッダを使わない」契約テストは含まれていますが、Testing セクションで言及されている blog-example のアサーションはありません。最終コミットから example が漏れたか、説明文が古いままかのいずれかと思われます。マージ前に説明文を実態と一致させる（または example を追加する）ことをお勧めします。

総評: **軽微な修正を前提に承認（Approve with minor changes）**。コアは堅実です。上記の説明文整合（#1）と `searchParams` の SSR 表現（#2）の2点だけ、マージ前にご対応いただければと思います。残りは任意です。

---

packages/router/src/search-params.ts
kfly8, line 30: 🤖 **Claude Code review** (posted on behalf of @kobaken)

**[medium] The SSR note contradicts the implementation.**

This signal is initialized from `window.location.search` on the client but `''` on the server (no `window`). So on a **direct load** of `/list?sort=price`, the server renders with empty params while the client hydrates with `sort=price` — a real hydration mismatch / flash for any island bound to `searchParams()`.

The docstring's "server and client read the same source … so the value agrees across hydration with no flash" (and the matching README wording) overstates this for the direct-load case. The behavior itself is a reasonable known limitation (the adapter-side SSR impl is a follow-up), but please soften the claim to something like: "agrees for client-side query navigations; a direct load with a query string flashes until the SSR seam lands."

---

packages/router/src/router.ts
kfly8, line 129: 🤖 **Claude Code review** (posted on behalf of @kobaken)

**[low] The query-only short-circuit doesn't supersede an in-flight navigation.**

This branch `return`s before the later `state.inflight?.abort()`. If a full outlet swap is in flight and a query-only navigation fires, the query update (signal + URL) applies immediately, but the older fetch can still resolve and commit through `replaceChildren` afterward, clobbering the URL/content.

Narrow race, but cheap to prevent: call `state.inflight?.abort()` (or guard) before taking this query-only path.

---

packages/router/src/router.ts
kfly8, line 163: 🤖 **Claude Code review** (posted on behalf of @kobaken)

**[medium] No focus / a11y management on swap.**

After `replaceChildren` replaces the outlet, focus is lost (it stays on a now-detached anchor or resets to `<body>`), and there's no `aria-live` route-change announcement. This is standard SPA-router territory and a real accessibility gap.

At minimum, document it in the README "Limitations" section (which currently lists only scroll restoration). Ideally, move focus to the outlet after the swap and announce the change via a live region.

---

packages/router/src/router.ts
kfly8, line 225: 🤖 **Claude Code review** (posted on behalf of @kobaken)

**[low] `preloaded` and the `modulepreload` links grow unbounded.**

The `state.preloaded` set and the `<link rel="modulepreload">` nodes appended to `<head>` are never pruned over a long session. The page cache is LRU-bounded, but this isn't. Minor leak — consider capping it, or removing each link element once its module has loaded.

---

packages/router/src/outlet.ts
kfly8, line 40: 🤖 **Claude Code review** (posted on behalf of @kobaken)

**[low / docs] Inline `<script>` inside the outlet won't execute.**

`importNode` + `replaceChildren` (on the caller side) doesn't run inline scripts; only body-end `<script type="module" src>` are `import()`-ed. This is consistent with the islands model, but it's an easy footgun for anyone putting an inline init script in the content region.

A one-line note in the README's "How it reuses the runtime" table ("inline scripts do not run on swap") would help.

---

kfly8 (commented):
🤖 **Review by Claude Code** (posted on behalf of @kobaken)

> _Note: this English review supersedes my earlier Japanese one ([review #4494033818](https://github.com/piconic-ai/barefootjs/pull/1910#pullrequestreview-4494033818)). The inline comments there can be ignored — same content, Japanese._

Overall a strong, well-structured PR: clean module split (cache / outlet / seams / router / types), solid test coverage, and a number of safe-by-default choices — last-wins via `AbortController`, the SWR + LRU cache, optional-peer isolation through `window.*` seams, and same-origin-only module loading. I left 5 inline comments (2 medium / 3 low). One more finding that isn't tied to a code line:

**[medium] The PR description references artifacts that aren't in the diff.**

The body (Description / Testing) mentions:
- "a runnable blog example (`packages/router/example`)"
- "an automated example test `packages/router/__tests__/blog-example.test.ts`", and "executed 19 tests"

None of these are among the 15 changed files. The only test file present is `packages/router/__tests__/router.test.ts` (~18 tests). The "ordinary HTML, no custom header" contract test is included, but the blog-example assertions described under Testing are not. Either the example was dropped from the final commit or the description is stale — please reconcile the description with what actually ships (or add the example) before merge.

**Recommendation: approve with minor changes.** The core is solid. Before merge I'd just address the description mismatch (above) and soften the `searchParams` SSR claim (inline). The rest are optional. Submitting as a non-blocking `COMMENT` rather than `REQUEST_CHANGES`.

---

kfly8:
### Additional finding: `@barefootjs/router/signals` SSR import fails to resolve `@barefootjs/client/reactive` (packaging)

Hit while running the example on a fresh clone (repro confirmed locally by deleting `packages/router/node_modules/@barefootjs/client`):

```
error: Cannot find module '@barefootjs/client/reactive' from
  /…/packages/router/src/search-params.ts
Bun v1.3.0 (macOS arm64)
```

**Cause.** `@barefootjs/router/signals` (`search-params.ts`) **statically** imports `@barefootjs/client/reactive` at module top level, but `@barefootjs/client` is declared only as an **optional peer** of `@barefootjs/router`. Some bun versions (e.g. 1.3.0) then don't symlink it into `packages/router/node_modules`, so any **SSR/Node** load of the signals entry can't resolve `@barefootjs/client/reactive` relative to `packages/router` — even after a full `bun run build`. (Bun 1.3.11 happened to link the optional peer, which is why it didn't reproduce everywhere.)

This is the packaging counterpart of the "signals needs `@barefootjs/client`" point: the **core** is legitimately client-optional, but the **`/signals` entry requires client**, and `peerDependenciesMeta.optional` is package-wide so it can't express "optional for core, required for `/signals`".

**Fix I'm using** (one line, keeps the core client-optional for external consumers):

```diff
   "devDependencies": {
     "@happy-dom/global-registrator": "^20.0.11",
-    "typescript": "^5.0.0"
+    "typescript": "^5.0.0",
+    "@barefootjs/client": "workspace:*"
   }
```

A workspace devDependency makes bun link `@barefootjs/client` into `packages/router/node_modules` reliably across versions, so the `/signals` SSR import resolves. Committed on #1911 (`fix(router): link @barefootjs/client so the /signals SSR import resolves`) so the example runs from a clean clone — please fold into #1910 if it looks right. Alternatives if you'd rather not add a devDep: document that `/signals` consumers must install `@barefootjs/client`, or bundle the reactive dep into the published `signals` entry.

---

packages/router/src/search-params.ts
kfly8: Addressed. I updated the `searchParams()` doc comment to say that the server-side initial value is empty because `window` is unavailable, and that direct loads with a query string may render the empty-query state until hydration until an adapter-side request-scoped SSR seam lands. The README now documents the same limitation.

packages/router/src/router.ts
kfly8: Addressed. The same-route query-only path now calls `state.inflight?.abort()` and treats the query-only navigation as the winner, so an older full navigation cannot commit a stale outlet swap afterward. I also added a regression test for this race.

packages/router/src/router.ts
kfly8: Addressed in docs for this PR. I added the focus-management / route-change-announcement gap to the README Limitations section. I intentionally left behavior changes for a follow-up because the a11y API and default focus target should be designed deliberately.

packages/router/src/router.ts
kfly8: Addressed in docs for this PR. I documented that modulepreload links and the preload dedupe set are session-lived while the page snapshot cache is LRU-bounded. Capping or cleanup can be handled as a follow-up.

packages/router/src/outlet.ts
kfly8: Addressed. The README now notes that inline scripts inside the outlet do not execute on swap and that island setup should live in the external module scripts emitted by BarefootJS.

packages/router/src/search-params.ts
kfly8: Addressed. I softened both the `searchParams()` doc comment and README wording: client-side query navigations stay reactive without a swap, but direct query-string loads can show the empty-query SSR state until hydration until the SSR seam lands.

packages/router/src/router.ts
kfly8: Addressed. The query-only branch now aborts any in-flight structural navigation, and the new regression test verifies that an older full navigation cannot overwrite the newer query-only URL/content state.

packages/router/src/router.ts
kfly8: Addressed in docs for this PR. The README Limitations section now calls out that focus management and route-change announcements are not implemented yet; implementation should be a follow-up.

packages/router/src/router.ts
kfly8: Addressed in docs for this PR. The README Limitations section now documents that modulepreload links and their dedupe set are session-lived; cleanup or caps can be handled later.

packages/router/src/outlet.ts
kfly8: Addressed. The README runtime section now explicitly says inline scripts inside the outlet do not execute on swap.

kfly8:
Close to rethink it.

Issue #1897, opened 2026-06-11
Mojo/Xslate adapters: composed site/ui demo corpus parity gaps (#1467 Phase 3)


kfly8:
## Summary

Per-adapter parity state of the **composed `site/ui` demo corpus** (`tabs`, `accordion`, plus future Phase 2c fixtures) on the template-language adapters, observed with real perl (Mojolicious 9.46 / Text::Xslate 3.5.9). Go's children-drop gap is tracked separately in #1896.

## Mojolicious

- **`accordion`** — refused at build time: the `AccordionItemContext.Provider` value is an object literal with arrow-function members (`{ open: () => props.open ?? false, … }`), no EP lowering → BF101 (pinned in `expectedDiagnostics`, same family as `radio-group`).
- **`tabs`** — compiles clean but **fails at render time** inside `BarefootJS::render_child` (`Mojo::Template` exception), i.e. the EP-side child-render helper can't service the demo-shaped `{{template}}`-equivalent call. Pinned in `skipJsx`.

## Xslate

Xslate renders the composed demos *almost* to Hono parity — children of imported components do render (unlike Go #1896). The remaining divergences:

- **`tabs`**: `aria-selected="1"` / `"0"` vs Hono's `"true"` / `"false"` — Perl boolean stringification of a prop-driven ARIA attribute. (The Mojo adapter routes ARIA booleans through `bf->bool_str` at compile time; Xslate appears to miss that path for prop-forwarded values.) Also text-node entity escaping: `'` emitted raw vs Hono's `&#39;`.
- **`accordion`**: the nested sibling's SVG (`ChevronDownIcon`, reached via accordion's own `../icon` import) is missing from the trigger output — nested-sibling template registration gap in the harness/adapter; plus `id=""` emitted for an undefined `id` prop where Hono omits the attribute; plus the same text-escaping divergence.

Both are pinned in `skipJsx` with a docstring pointing here (`accordion` could alternatively pin BF101 — the sibling compile does emit it — but the render-level divergences above are the durable signal worth keeping visible).

## Coverage note

All pinned fixtures keep full coverage at the Hono SSR conformance layer and the real-browser fixture-hydrate layer (#1467); these pins only affect cross-adapter SSR parity, which is the #1467 Phase 3 work item.

kfly8:
## Status update — PR #1904 up (stacked on #1903)

[...]

PR #1975, opened 2026-06-24
refactor(go-template): Phase 4 decomposition — GoEmitContext seam + first clusters

kfly8:
## Summary

Phase 4 of the Go html/template adapter cleanup, aiming at the ideal decomposition where `GoTemplateAdapter` becomes a thin orchestrator and each logical domain lives in its own module. Phases 1–3 (#1974) extracted pure helpers, types, and per-compile state; this PR establishes the **architecture** the rest of the decomposition builds on and lands the first self-contained clusters.

**Behaviour is unchanged (byte-identical):** generated Go template output is not altered.

## Why a seam is needed

The adapter's lowering is deeply mutually recursive (expression lowering ↔ condition lowering ↔ rendering), and the `ParsedExprEmitter` / `IRNodeEmitter` / `AttrValueEmitter` interface methods *must* stay on the class (the shared dispatchers call `adapter.literal()`, `adapter.emitElement()`, …). So a cluster pulled into its own module still needs to call back into the shared per-compile state and the recursive entry points.

`GoEmitContext` is that seam: a **narrow interface** exposing the per-compile `state` plus the recursive entry points (`parseLiteralExpression`, `convertExpressionToGo`, `convertConditionToGo`). Extracted modules are free functions taking a `GoEmitContext`; the adapter implements it and passes `this`. Modules depend on this interface, not the 8k-line class — so cross-module coupling is explicit and the modules are unit-testable against a stub. The surface is kept minimal: a member is added only when an extracted module genuinely needs it.

## Changes

| Module | Contents | Coupling |
|---|---|---|
| `analysis/component-tree.ts` | Pure IR structural walks: `hasClientInteractivity`, `hasEventsInTree`, child/nested-component discovery | none (no state, no rendering) — plain free functions |
| `emit-context.ts` | The `GoEmitContext` interface | — |
| `expr/helper-inline.ts` | Local arrow-const helper inlining at a call site (`inlineLocalHelperCall` + splice-safety guards + `substituteHelperParams`) | via `GoEmitContext` |
| `expr/url-builder.ts` | `URLSearchParams` builder helpers → `bf_query` lowering (#1897) | via `GoEmitContext` |

Main file: **8,048 → 7,533 lines**.

## Verification (each commit)

- `adapter-go-template` unit: ✅ 551 pass / 0 fail
- `adapter-tests` conformance (byte-identical SSR): ✅ 786 pass / 0 fail
- `tsgo --noEmit`: ✅ &nbsp;·&nbsp; `biome check`: ✅

## Roadmap (subsequent PRs, on this seam)

`type-codegen` → `value-lowering` → `memo-lowering` → `spread-codegen` → `props-codegen` → `array-lowering`, with node rendering (`renderElement` / `renderLoop` / `renderComponent` / `renderAttributes`) remaining as the orchestrator core. Each is its own byte-identical, test-verified PR.

---

### DIFF
--- a/.changeset/go-adapter-phase4-decompose.md
+++ b/.changeset/go-adapter-phase4-decompose.md
@@ -1,0 +1,10 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Continue decomposing the Go html/template adapter (Phase 4). Internal-only, output byte-identical (verified by the adapter unit + conformance suites); no behavioural or public-API change.
+
+- `analysis/component-tree.ts` — pure IR structural walks (`hasClientInteractivity`, nested/child-component discovery) moved out as free functions that read no adapter state.
+- `emit-context.ts` — introduce `GoEmitContext`, the narrow interface (per-compile `state` + the recursive entry points) that extracted emit modules depend on instead of the concrete adapter. The adapter implements it and passes `this`.
+- `expr/helper-inline.ts` — local arrow-const helper inlining at a call site.
+- `expr/url-builder.ts` — `URLSearchParams` builder helpers → `bf_query` lowering.
--- a/packages/adapter-go-template/src/adapter/analysis/component-tree.ts
+++ b/packages/adapter-go-template/src/adapter/analysis/component-tree.ts
@@ -1,0 +1,176 @@
+/**
+ * IR component-tree analysis for the Go html/template adapter.
+ *
+ * Pure structural walks over the IR — no adapter instance state, no rendering.
+ * Extracted from `go-template-adapter.ts` (Phase 4 decomposition). Only the two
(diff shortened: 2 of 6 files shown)


kfly8:
Addressed the Copilot review in cd4306a: reworked the seam so the exported adapter's public surface is unchanged. `state` / `parseLiteralExpression` / `convertExpressionToGo` / `convertConditionToGo` are `private` again and the class no longer `implements GoEmitContext`; instead the adapter holds a private `emitCtx: GoEmitContext` built over those private members and passes `this.emitCtx` to the extracted modules. This keeps the changeset's "no public-API change" accurate. Output stays byte-identical (unit 551, conformance 786).

---

PR #1977, opened 2026-06-25
refactor(jsx,go-template): carry IRExpression.parsed; reuse it in the Go adapter

kfly8:
## Summary

Continues the **"IR carries the semantics, adapters emit from it"** direction (after the merged memo PR #1976), this time for text-interpolation expressions. The IR now stores the parsed expression tree for each `{expr}` node, so SSR adapters emit from it instead of each re-parsing the string at emit time — and a multi-adapter build (Go + Mojo) parses each expression once, not per adapter.

**Behaviour is unchanged (byte-identical).** The only public-API change is additive: `IRExpression` gains an optional `parsed` field.

## Changes

**`@barefootjs/jsx`**
- `IRExpression.parsed?: ParsedExpr`. `jsxToIR` walks the produced tree once and attaches `parseExpression(expr.trim())` to every text-interpolation node (`attachParsedExpressions`). **Best-effort and fallback-safe**: a node this walk doesn't reach (or one with an empty `expr`) simply has no `parsed`, and the adapter falls back to parsing — so under-coverage is never a behavioural change.

**`@barefootjs/go-template`**
- `convertExpressionToGo` takes an optional `preParsed` tree; `renderExpression` passes `expr.parsed`. A rendered interpolation reuses the IR's parse (`preParsed ?? parseExpression(trimmed)`) instead of calling `parseExpression` again. The string-based early returns (null/undefined, static record index, inlined consts, helper/url lowering) are unchanged and still run first — `preParsed` is consulted only at the single parse site after them.

## Tradeoff

Expressions the adapter resolves via an early return (e.g. `x ?? undefined`, inlined consts) are now parsed at IR-build even though the adapter skips parsing them. Those are a minority, so the extra build-time parsing is negligible next to moving the parse to Phase 1 and sharing it across adapters.

## Verification

- `adapter-go-template` unit: ✅ 552 / 0
- `adapter-tests` conformance (byte-identical SSR, real analyzer→IR→adapter pipeline): ✅ 786 / 0
- `@barefootjs/jsx` suite: ✅ 2181 / 0
- `tsgo --noEmit` ✅ · `biome check` ✅

## Follow-ups (same direction)

Extend `parsed` to attribute expressions (`ExpressionAttr`) and conditions, then signal `initialValue` and local-const values; and have the Mojo adapter consume `IRExpression.parsed` / `MemoInfo.parsed` too. Each is its own byte-identical PR.

---

### DIFF
--- a/.changeset/ir-expression-parsed.md
+++ b/.changeset/ir-expression-parsed.md
@@ -1,0 +1,9 @@
+---
+"@barefootjs/jsx": patch
+"@barefootjs/go-template": patch
+---
+
+Carry the parsed expression tree in the IR for text-interpolation nodes, so SSR adapters emit from it instead of each re-parsing the string at emit time (and a multi-adapter build parses it once, not per adapter). Output byte-identical; the only public-API change is additive (`IRExpression` gains an optional `parsed` field).
+
+- `@barefootjs/jsx`: `jsxToIR` now walks the produced tree and attaches `IRExpression.parsed` (`parseExpression(expr.trim())`) to every text-interpolation node. Best-effort — a node left without `parsed` (or an empty expr) just falls back to adapter-side parsing, so it is never a behavioural change.
+- `@barefootjs/go-template`: `convertExpressionToGo` takes an optional pre-parsed tree and `renderExpression` passes `expr.parsed`, so a rendered interpolation reuses the IR's parse instead of calling `parseExpression` again. The string-based early returns (null/undefined, static record index, inlined consts, helper/url lowering) are unchanged and still run first.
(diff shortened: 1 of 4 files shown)


kfly8:
Both addressed in 64a2ebf:

1. **Walk coverage** — `attachParsedExpressions` now also recurses the expression nodes stored off the main `children` array: `IRAsync.fallback`, `IRLoop.childComponent.children`, `IRLoop.nestedComponents[*].children`, and `IRLoop.flatMapCallback.fragments[*].ir`. (Still fallback-safe — this only widens the optimization; output stays byte-identical.)
2. **Doc** — reworded the `IRExpression.parsed` comment to state it's optional by design (best-effort walk, under-coverage possible) so consumers always fall back to parsing `expr`.

Verification unchanged: unit 552, conformance 786, jsx green, biome ✅.

---

Issue #2018, opened 2026-06-27
Introduce a lightweight ParsedExpr evaluator for higher-order callbacks; unify reduce/sort/map/filter lowering; retire adapter-side re-parsing (terminal sweep)


kfly8:
## Background

The terminal sweep (#2006) drove emit-time `ts.createSourceFile` / `parseLiteralExpression` out of the Go adapter almost entirely. The only real call sites left on `main` are **2, both in `expr/url-builder.ts`**. Everything else flows through structured IR (`ParsedExpr` / `ParsedExpr2`) and the adapter only emits.

That last case exposed the real, broader question: **how should the compiler treat block bodies and, especially, higher-order callbacks?** Expressions are carried structurally by the IR; statement sequences and callback bodies are not. Today the adapter handles them by *recognizing specific patterns* — `bf_sort` matches particular comparator shapes, `bf_reduce` matches `+`/`*` folds with `self`/`field` projections, block memos match named idioms (#1897, #1945, #2015). That is recognition without abstraction — an ad-hoc list that grows with each new shape.

This issue defines the abstraction and the boundary.

## Principle: purely-functional expressibility is the single support criterion

Whether something can be supported is decided by one question — *can it be written as a pure value computation?* — not by which named pattern it matches.

| Construct | Treatment |
| --- | --- |
| HTML structure control (iterate/condition to **emit elements**, no state accumulation) | **Statement** — `IRLoop` / `IRConditional` (already today) |
| Value computation (operators, ternary, calls, member access, …) | **Expression**, lowered to **template-native** syntax (keeps templates readable) |
| Value accumulation | **fold / reduce** (a loop that mutates local state is always a fold) |
| **Higher-order callback bodies** (`reduce` / `sort` / `map` / `filter` / `find` `(…) => expr`) | **Evaluated by a lightweight `ParsedExpr` evaluator** in each backend runtime (see below) |
| Value-producing `if` / early-return; pure `let`-binding sugar | Normalized to ternary / inlined into the expression |
| Side-effecting / mutable APIs, raw `for` accumulation, `break`, I/O | **`unsupported`** → diagnostic (BF101) |

## Key move: a lightweight `ParsedExpr` evaluator, scoped to higher-order callbacks

[...]

PR #2025, opened 2026-06-28
refactor(mojo,xslate): lower spreads from IR ParsedExpr, drop emit-time re-parse (#2018)

kfly8:
## Summary

Tracks **issue #2018**: drive the Mojolicious and Xslate adapters' expression lowering from the IR-carried structured `ParsedExpr` tree instead of re-parsing source text with `ts.createSourceFile` at emit time — achieving **zero live `createSourceFile` calls** in both adapters' `src` (test files excluded; only explanatory comment references remain, matching the go-template precedent).

This applies to mojo/xslate exactly what go-template already achieved:
- **U5** — conditional spread lowered from `SpreadAttr.parsed` (`go-template/.../spread/spread-codegen.ts`).
- **U6** — inline object-literal → map/hashref from `ExpressionAttr.parsed` (`go-template/.../value/value-lowering.ts`, `objectLiteralToGoMap`).
- **Roadmap A** — structured literal lowering (`go-template/.../value/parsed-literal-to-go.ts`).

## Changes (5 emit-time re-parse sites across 3 files)

- `packages/adapter-mojolicious/src/adapter/spread/spread-codegen.ts` — `conditionalSpreadToPerl` and `objectLiteralExprToPerlHashref` now take `ParsedExpr | undefined` (was `expr: string`); `objectLiteralToPerlHashref` / `recordIndexAccessToPerl` consume the structured tree.
- `packages/adapter-xslate/src/adapter/spread/spread-codegen.ts` — same for the Kolon (`conditionalSpreadToKolon`, `objectLiteralExprToKolonHashref`, …).
- `packages/adapter-mojolicious/src/adapter/value/parsed-literal.ts` — the orphaned `parsePureStringLiteral` (the file's only `ts.createSourceFile`) is **removed**: module-scope pure-string-const inlining is owned by the shared `collectModuleStringConsts` in `@barefootjs/jsx` (consumed via `moduleStringConsts`), so the adapter-local copy had **zero callers**. Removing it is behaviour-preserving and reaches `createSourceFile = 0`. The two live helpers (`isStringTypeInfo`, `isBareStringLiteral`) are retained.

Call sites in `mojo-adapter.ts` / `xslate-adapter.ts` pass `value.parsed` for the spread / component-prop expressions, and `parseExpression(initTrimmed)` (the analyzer's own entry, not `ts.createSourceFile`) for the local-const resolution branch — mirroring go-template's same path.

## Byte-identity

The condition and scalar values are re-stringified with `stringifyParsedExpr` and routed back through `convertExpressionToPerl` / `convertExpressionToKolon`, which re-parse — so the emitted Perl/Kolon is byte-identical to the former AST-text path. The `ts.factory` rebuild in `recordIndexAccess*` only reconstructs the `IDENT[KEY]` node the shared `parseRecordIndexAccess` accepts; no source-text re-parse. **`@barefootjs/jsx` is unchanged** — only existing IR fields/exports are consumed.

## Verification

- `bun test packages/adapter-mojolicious/` → **527 pass / 0 fail**
- `bun test packages/adapter-xslate/` → **353 pass / 0 fail**
- `bunx tsgo --noEmit` (both packages) → clean
- `bunx biome check` (both adapter dirs) → clean
- `git grep -nE "\.createSourceFile\(|parseLiteralExpression\("` over both adapters' `src` (tests excluded) → **0 live calls**

Local Perl/Xslate interpreters are absent, so real-render conformance cases skip locally (CI runs them); IR/structure conformance passes.

---

### DIFF
--- a/.changeset/mojo-xslate-parsedexpr.md
+++ b/.changeset/mojo-xslate-parsedexpr.md
@@ -1,0 +1,6 @@
+---
+"@barefootjs/mojolicious": patch
+"@barefootjs/xslate": patch
+---
+
+Lower conditional-spread and inline object-literal expressions from the IR-carried structured `ParsedExpr` tree instead of re-parsing source with `ts.createSourceFile` at emit time (#2018, mirroring go-template's U5/U6/Roadmap-A). Behaviour and output are unchanged — the condition and scalar values still route through `convertExpressionToPerl` / `convertExpressionToKolon`, which re-parse, so the emitted Perl/Kolon stays byte-identical. The now-orphaned `parsePureStringLiteral` (superseded by the shared `collectModuleStringConsts`) was removed from the Mojo adapter.
--- a/packages/adapter-mojolicious/src/adapter/mojo-adapter.ts
+++ b/packages/adapter-mojolicious/src/adapter/mojo-adapter.ts
@@ -867,10 +867,13 @@ export class MojoAdapter extends BaseAdapter implements IRNodeEmitter<MojoRender
       }
       // Inline object-literal child prop (carousel's `opts={{ align: 'start' }}`):
       // lower to a Perl hashref so the child can serialize it (`data-opts`),
-      // instead of refusing the bare object with BF101. (#1971 Perl) Cheap `{`
-      // guard so the common non-object case skips the AST parse.
-      if (value.expr.trim().startsWith('{')) {
-        const hashref = objectLiteralExprToPerlHashref(this.spreadCtx, value.expr)
+      // instead of refusing the bare object with BF101. (#1971 Perl) Read the
+      // IR-carried structured `ParsedExpr` tree (#2018) instead of re-parsing
+      // `value.expr` with `ts.createSourceFile`; the lowering returns null for
+      // any non-object-literal shape, so the common non-object case falls
+      // straight through to the bare-expression path below.
+      if (value.parsed) {
+        const hashref = objectLiteralExprToPerlHashref(this.spreadCtx, value.parsed)
         if (hashref !== null) return `${perlHashKey(name)} => ${hashref}`
       }
       return `${perlHashKey(name)} => ${this.convertExpressionToPerl(value.expr)}`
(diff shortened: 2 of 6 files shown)


kfly8 (commented):
Reviewed against the design principles and the go-template precedent. `createSourceFile = 0` is achieved in both adapters, byte-identity is exercised by the live-interpreter `test` checks (both green), and `@barefootjs/jsx` is untouched. Good, well-scoped track-#2018 step. (Can't formally Approve my own PR — leaving as a comment: LGTM.)

## Matches the go-template model
- **object-literal → hashref** is lowered structurally from the `ParsedExpr` tree (`objectLiteralToPerlHashref`), mirroring go-template's `objectLiteralToGoSpreadMap`.
- **`recordIndexAccessToPerl`** rebuilds the `IDENT[KEY]` node with `ts.factory` and feeds the shared `parseRecordIndexAccess` — no source-text re-parse — exactly the go-template `recordIndexAccessToGoMap` approach.
- Removing the orphaned `parsePureStringLiteral` (0 callers; module-string-const inlining is owned by `collectModuleStringConsts` in jsx) is the right behaviour-preserving way to reach `createSourceFile = 0`.

## One design note (not blocking)
The condition and non-structural scalar values are lowered via `ctx.convertExpressionToPerl(stringifyParsedExpr(expr.test))` — i.e. the IR's `ParsedExpr` is re-stringified and re-parsed inside `convertExpressionToPerl`. So while the direct `ts.createSourceFile` call is gone, a re-parse still happens one layer down.

This differs from go-template, which avoids the round-trip two ways: (a) it lowers the spread condition structurally via `conditionToGoBool(expr.test, ir)` (a spread-specific, type-aware ParsedExpr→bool path, no `convertExpressionToGo` round-trip), and (b) its `convertExpressionToGo(jsExpr, out?, preParsed?)` accepts a `preParsed` tree so callers never re-stringify.

For this PR's scope (drop emit-time `createSourceFile`) the round-trip is a reasonable transitional step and you've documented it. The clean finish is in #2018: give `convertExpressionToPerl` / `convertExpressionToKolon` a `preParsed?: ParsedExpr` param (matching `convertExpressionToGo`) so the carried tree is threaded straight through instead of stringify→re-parse. The seam docstrings already point at #2018 for this — maybe add a one-line `// TODO(#2018)` at the `stringifyParsedExpr(...)` call sites so the round-trip is easy to find and remove later.

Net: the structural parts match go-template, the round-trip is a documented, transitional consequence of the string-only seam, and it's the right thing to retire in #2018. Nice work.

---

kfly8:
Addressed the design note in `09882fe`: added a `// TODO(#2018)` at each of the four `convertExpressionTo{Perl,Kolon}(stringifyParsedExpr(…))` round-trip sites (the spread condition + the non-structural scalar value, in both adapters), pointing at the `preParsed?: ParsedExpr` finish (cf. `convertExpressionToGo(jsExpr, out?, preParsed?)`) so the round-trip is easy to find and retire later.

Left the round-trip itself in place as the documented transitional step for this PR's scope — the structural lowering (`objectLiteralTo*Hashref`, `recordIndexAccessTo*` via `ts.factory` + `parseRecordIndexAccess`) already matches go-template, and threading the carried tree straight through belongs with the `convertExpressionTo*` signature change in #2018. Comment-only, so output stays byte-identical (mojo 464 / xslate 339, 0 fail); tsgo + biome clean.

---

PR #2026, opened 2026-06-28
refactor(mojo,xslate): thread preParsed through spread converters (#2018)

kfly8:
## Summary

Removes the residual `ParsedExpr` round-trip (stringify → re-parse) in the spread lowering of the Mojolicious and Xslate adapters, tracked by #2018.

The conditional-spread / object-literal spread codegen (`spread/spread-codegen.ts`) previously took the IR-carried `ParsedExpr` tree, re-stringified it with `stringifyParsedExpr(...)`, and routed it back through `convertExpressionToPerl` / `convertExpressionToKolon`, which then **re-parsed** the text. `ts.createSourceFile` was already at 0 after #2025, but this re-parse remained.

This change aligns the Perl/Kolon converters with go-template's `convertExpressionToGo(jsExpr, out?, preParsed?)`: both converters now accept an optional `preParsed?: ParsedExpr` and use the carried tree directly (`const parsed = preParsed ?? parseExpression(...)`), so the spread call sites thread the tree straight through with no round-trip.

## Changes

- **`emit-context.ts`** (mojo + xslate): extended the `convertExpressionToPerl` / `convertExpressionToKolon` seam declarations (on the `*SpreadContext` and `*MemoContext` interfaces) to `(expr: string, preParsed?: ParsedExpr)`. The `#2018` "should also accept preParsed" TODO docstrings are updated to describe the now-realized behavior.
- **`mojo-adapter.ts` / `xslate-adapter.ts`**: the converter implementations use `preParsed` directly when supplied instead of re-parsing; the `spreadCtx` / `memoCtx` getters forward the new arg.
- **`spread/spread-codegen.ts`** (mojo + xslate): the 4 round-trip sites (each adapter's conditional-spread condition + non-structural object value) now pass the carried tree as `preParsed` and drop the `stringifyParsedExpr(...)` first argument (passing `''`, since the converter derives diagnostic text from the tree). The `stringifyParsedExpr` call backing the BF101 "indexes a record map…" message is **retained** (display-only).

## Byte-identity

Guaranteed: `preParsed` is the same tree that `parseExpression(stringifyParsedExpr(tree))` produced, so the AST emitter receives identical input and emits identical Perl/Kolon. The only string still derived from text — the BF101 "Expression not supported" message — is reconstructed via `stringifyParsedExpr(parsed)` when `preParsed` is set, which equals the former `${trimmed}` source. `ts.createSourceFile` stays at 0 (verified: all remaining references are comments).

## Verification

- `bun test packages/adapter-mojolicious/` → **527 pass / 0 fail**
- `bun test packages/adapter-xslate/` → **353 pass / 0 fail**
- `bunx tsgo --noEmit` (both packages) → clean
- `bunx biome check` (both adapter dirs) → clean

Real Perl/Xslate render conformance runs in CI.

---

### DIFF
--- a/.changeset/mojo-xslate-preparsed-seam.md
+++ b/.changeset/mojo-xslate-preparsed-seam.md
@@ -1,0 +1,8 @@
+---
+"@barefootjs/mojolicious": patch
+"@barefootjs/xslate": patch
+---
+
+Remove the spread-lowering `ParsedExpr` round-trip in the Mojolicious and Xslate adapters (#2018).
+
+The conditional-spread / object-literal spread codegen previously re-stringified the IR-carried `ParsedExpr` tree (`stringifyParsedExpr`) and routed it back through `convertExpressionToPerl` / `convertExpressionToKolon`, which re-parsed the text. The seam now matches go-template's `convertExpressionToGo(jsExpr, out?, preParsed?)`: the converters accept an optional `preParsed?: ParsedExpr` and thread the carried tree straight through, eliminating the stringify→re-parse round-trip. Output is byte-identical (the carried tree is exactly what re-parsing the stringified text produced). `stringifyParsedExpr` is retained only for BF101 diagnostic message text.
--- a/packages/adapter-mojolicious/src/adapter/emit-context.ts
+++ b/packages/adapter-mojolicious/src/adapter/emit-context.ts
@@ -76,12 +76,14 @@ export interface MojoSpreadContext {
   /**
    * Lower a JS expression to its Perl form (the core recursive entry).
    *
-   * String-in today. Once the IR-carries-semantics work (#2018) lands, this
-   * seam should also accept an already-parsed `ParsedExpr` — cf. go-template's
-   * `convertExpressionToGo(jsExpr, out?, preParsed?)` — so the future Perl
-   * evaluator can thread a structured tree instead of re-parsing source text.
+   * When the IR already carries a structured `ParsedExpr` tree, pass it as
+   * `preParsed` so the converter threads it straight through instead of
+   * re-parsing `expr` — mirrors go-template's
+   * `convertExpressionToGo(jsExpr, out?, preParsed?)`. With `preParsed` set,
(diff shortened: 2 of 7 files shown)


kfly8 (commented):
Reviewed — this cleanly retires the round-trip flagged in #2025 and brings the Perl/Kolon converters to full parity with go-template's `convertExpressionToGo(jsExpr, out?, preParsed?)`. (Can't Approve my own PR — comment: LGTM.)

## Verified
- **Round-trip gone.** The 4 spread sites now call `convertExpressionTo*('', expr.test)` / `('', val)`, threading the IR tree as `preParsed` with no `stringify → re-parse`. The only `stringifyParsedExpr` left is the BF101 record-map message (display-only), as intended.
- **`''` first-arg is safe.** The impl is `if (preParsed) { parsed = preParsed } else { const trimmed = expr.trim(); if (trimmed === '') return "''"; parsed = parseExpression(trimmed) }` — when `preParsed` is set the `''` is never parsed; and even the no-`preParsed` `''` path is guarded. Equivalent to go-template's `preParsed ?? parseExpression(trimmed)`.
- **Byte-identical.** `preParsed` is the same tree `parseExpression(stringifyParsedExpr(tree))` produced, so the AST emitter input is unchanged. The one text-derived output — the BF101 "Expression not supported" message — is reconstructed as `preParsed ? stringifyParsedExpr(parsed) : expr.trim()`, which equals the former `${trimmed}`. So both emitted code and diagnostics are unchanged.
- **`createSourceFile` stays 0**; no jsx changes; `check`/`dist`/`deno-check`/`biome`/`smoke` + one `test` shard green (second render shard still running — expected green like #2025).

## Design
This is the clean finish I noted on #2025: the seam now carries `preParsed` exactly like `convertExpressionToGo`, so mojo/xslate spread lowering no longer discards the IR's structure only to rebuild it. With this, all three adapters share the same "IR carries the tree, no emit-time re-parse" contract. 👍

Once the second `test` shard goes green this is good to merge.

---

PR #2037, opened 2026-06-29
refactor(expr): collapse ParsedExpr2 into a single generic ParsedExpr (#2018 P5)

kfly8:
## Summary

Final phase of the EXPR2 migration (#2018). Retires the **folded** expression model now that the runtime evaluator drives every higher-order callback body on both SSR backends (Go `eval.go`, Perl `Evaluator.pm`, landed in P1–P3), and unifies the compiler on **one generic `ParsedExpr`**.

Net **−1545 lines**: the dual-model machinery (`ParsedExpr2`, the folding extractors, the structured fold kinds) is gone.

## Parser / IR (`packages/jsx`)
- Higher-order callbacks (`.filter`/`.find`/`.findIndex`/`.findLast`/`.findLastIndex`/`.every`/`.some`/`.sort`/`.toSorted`/`.reduce`/`.reduceRight`/`.flatMap`) now parse to a generic `call` whose argument is a generic multi-param `arrow`. Added `arrow` + `regex` to the union; **deleted** the folded kinds `higher-order`, `arrow-fn`, and the structured sort/reduce/flatMap `array-method` variants (kept the non-callback `array-method` + `flat`).
- **Deleted** `ParsedExpr2`, `parseExpression2`, `convertNode2`, the three bridges, and the `extract{Sort,Reduce,FlatMap}*FromTS` extractors. Ported the sort-comparator catalogue to `sortComparatorFromArrow` (walks the generic arrow body) as the `localeCompare` fallback. Added `asCallbackMethodCall` / `CALLBACK_METHODS` / `tsNodeToParsedExpr`.
- Emitter: replaced `higherOrder`/`sortMethod`/`reduceMethod`/`flatMapMethod`/`arrowFn` with a single `callbackMethod` arm + `arrow`/`regex`; the dispatch recognises a callback call once via `asCallbackMethodCall` (`tsc` exhaustiveness is the safety net).
- `IRLoop` carries the generic comparator arrow (`IRLoopSort`) for the `.sort().map()` hoist; the analyzer attaches one generic `parsed` tree per const, retiring the Go-only `parsed2`.

## Adapters (go-template / mojolicious / xslate)
- `callbackMethod` is **eval-first**: serialize the arrow body → `bf_*_eval`; fall back to the structured `bf_sort`/`bf->sort` only for a `localeCompare` comparator (`sortComparatorFromArrow`); reduce/flatMap are eval-only (the catalogue always serializes). Refit the eval-emit helpers to take `(body, params)`. Migrated the Go ctor lowering from `parsed2` to the generic `parsed` tree.

## Invariant
**Behaviour-neutral.** Emitted SSR template text changes (`bf_sort …` → `bf_sort_eval … "<json>"`), but **rendered HTML is identical**. Template-text pins regenerated in-PR; rendered-HTML conformance fixtures unchanged.

One deliberate, principled **capability expansion** surfaced by the collapse (consistent across Go + Mojo + Xslate): a `flatMap` projection with a literal element (`i => [i.name, "x"]`) now lowers via the evaluator instead of refusing (BF101) — it serializes cleanly and renders Go==Perl==JS. No previously-supported input changes.

## Verification
- `tsc --noEmit` clean: jsx + all 3 adapters.
- `bun test`: jsx **2178 ✓ / 0 fail**, go-template **552 ✓**, mojolicious **475 ✓**, xslate **356 ✓**.
- adapter-tests conformance (CSR + per-adapter HTML) **789 ✓ / 0 fail**; Go runtime `go test ./...` ✓.
- Changeset added (jsx + 3 adapters, patch).

---


kfly8:
Thanks @Copilot — addressed the review:

- **`isStringExpr` per-call `Set` (go-template)** and **`callbackMethod` per-call `Set` (mojolicious)** → fixed in bf1746c9: both hoisted to module-level `ReadonlySet` constants (`STRING_METHODS` / `PREDICATE_METHODS`), reused across calls. Behaviour-identical; go/mojo `tsc` + suites green. (go-template's `PREDICATE_METHODS` was already a `static readonly`; the remaining `new Set(...)` sites are per-call dynamic dedup sets, not hot-path constants.)

- **Xslate predicate Kolon-lambda fallback degrading nested higher-order calls** → tracked as #2038 rather than fixed here. This lossy fallback is **pre-existing** (the folded model fell to the identical lambda path when a predicate body contained a nested `higher-order`), so this PR faithfully preserves it and stays behavior-neutral. Surfacing BF101 for such predicates is a deliberate compile-behavior change (a currently-compiling component becomes an explicit `@client`), so it belongs in its own PR with a render-conformance fixture — see #2038 for the repro (`thread-demo.tsx`, which is `"use client"` and outside the SSR render gate) and the proposed `containsHigherOrder`-driven guard.

---

kfly8:
Second review pass addressed in c696977c:

- **`IRLoopSort.arrow` typed too wide (types.ts)** → narrowed to `Extract<ParsedExpr, { kind: 'arrow' }>`. All consumers already read `.params`/`.body`; tsc clean across jsx + 3 adapters (type-only, no runtime change).
- **Unsupported-sort message vs actual catalogue (jsx-to-ir.ts)** → message now also lists the relational-ternary sign comparator and `||`-chained multi-key forms that `sortComparatorFromArrow` accepts, so an accepted comparator is no longer described as rejected.
- **`emitSortEval` / `emitReduceEval` defaulting missing param names (go-emit.ts)** → both now guard on `params.length >= 2` and return `null` on wrong arity, failing over to the structured fallback / BF101 cleanly instead of binding the wrong env — matching the Perl adapters' arity guard.

Verified: jsx (`unsupported-expression` + `ir-sort-comparator`) 35✓, go-template suite 523✓/0 fail, `tsc --noEmit` clean in all four packages.

---

kfly8:
Third review pass:

- **`renderSortEval` / `renderReduceEval` arity (xslate array-method.ts)** → fixed in 35800991: both now guard `params.length < 2` and return null, matching the Go/Mojo emitters. xslate suite 356✓/0 fail, tsc clean.

- **`matchTrailingSlashReplace` using `pattern.text` (ctor-lowering.ts:41)** → this one is a **false positive**, no change. Unlike `StringLiteral`, a `RegularExpressionLiteral`'s `.text` *does* include the surrounding `/.../`. Verified empirically:
  ```ts
  // src: x.replace(/\/+$/, '')
  pattern.text      === "/\\/+$/"   // i.e. the literal /\/+$/  (with slashes)
  pattern.getText() === "/\\/+$/"
  ```
  So `pattern.text !== '/\\/+$/'` matches correctly and the `strings.TrimRight` lowering works; it's also covered by the existing green go-template tests.

---

Issue #2048, opened 2026-06-29
queryHref follow-up: multi-value keys / arrays (Q4)


kfly8:
Follow-up to #2042 (the `queryHref` umbrella — DoD fully shipped across #2044/#2045/#2046/#2047). These two open questions were enhancements beyond the original scope.

> **Status:** Q5 (encoder parity) is **done** — shipped in #2049. Q4 (multi-value keys) remains open and is the only thing tracked here now.

## Q4 — multi-value keys / arrays (open)

`queryHref` currently takes a flat `{ key: string | null | undefined }` object — one value per key, `URLSearchParams.set` semantics (a repeated key overwrites). There is no way to express a key that appears multiple times (`?tag=a&tag=b`), i.e. `URLSearchParams.append`.

Open design points:
- API shape — allow `string[]` as a value (`{ tag: ['a', 'b'] }`)? Empty-array / empty-string-member handling (skip empties, matching the scalar truthy-omit)?
- Each SSR adapter's helper would need an append path: go `bf_query`, the shared Perl `query` helper (Mojo/Xslate).
- Conformance: byte-identical to `URLSearchParams` for the array case (order preserved, each member form-encoded). The shared golden helper vectors (`packages/adapter-tests/helper-vectors`, fn `query`) added in #2049 are the natural place to pin the array cases once implemented.

Out of scope for v0; only worth doing once a real component needs repeated keys.

## Q5 — encoder parity between go `bf_query` and `URLSearchParams` ✅ Done (#2049)

go's `bf_query` encoded via `url.QueryEscape`, which diverged from `URLSearchParams` (client / Perl `query`) on exactly two characters — `~` (`%7E` vs kept) and `*` (kept vs `%2A`). #2049 replaced it with a `formEscape` matching `URLSearchParams` byte-for-byte and promoted the `query` helper to the shared golden vectors, so all four backends (Hono ≡ Go ≡ Mojo ≡ Xslate) now render query strings identically. Refs: #2042, #2044, #2045.

kfly8:
Closing as **not planned**.

- **Q5** (encoder parity) shipped in #2049 — all four backends (Hono ≡ Go ≡ Mojo ≡ Xslate) now render query strings byte-for-byte identically, covered by the shared golden `query` vectors.
[...]

Issue #2042, opened 2026-06-29
Add a pure `queryHref` API (the functional query-string builder) → retire the imperative URLSearchParams-builder recognizer


kfly8:
Follow-up to #2039 / #2040 (and #2018).

> **Status (umbrella issue — kept open):** the API + go-template SSR landed in #2044 (merged). The remaining work is in flight as a stack of three PRs: **Mojo/Xslate (Perl) parity** (#2045), **migrating existing usages** (#2046), and **retiring the `urlBuilder` recognizer** (#2047 — the actual "delete the concrete code" payoff). See the checklist below.

## Background

#2039 retired the go-template adapter's emit-time re-parse of the `URLSearchParams` "href builder" idiom by recognizing it at analysis time and carrying it as IR (`ConstantInfo.urlBuilder`, see `packages/jsx/src/url-builder-shape.ts`). That works, but the recognizer is an **ad-hoc, domain-specific matcher** for one imperative source shape:

```tsx
const hrefFor = (sort: string, tag: string) => {
  const u = new URLSearchParams()
  if (sort !== 'date') u.set('sort', sort)
  if (tag) u.set('tag', tag)
  const s = u.toString()
  return s ? `${root}?${s}` : root
}
```

#2040 (block → expression normalization) will **not** subsume this idiom: `new URLSearchParams()` + conditional `u.set(...)` is mutable-stdlib-object accumulation, which #2040 explicitly puts on the *refuse (BF101)* side (mutable local state / side-effecting APIs), not the normalize side. The `.set()`-accumulation → query-string semantics is domain knowledge (`bf_query`), not a generic block→expression transform — so a general normalizer can't delete the recognizer.

The clean way to **delete the concrete code** (rather than keep recognizing imperative source) is to give users a pure, functional API to express the same thing — the option #2039 itself flagged: *"expressible structurally, or a pure `queryHref` API"*.

## Goal

Add a pure `queryHref` helper to `@barefootjs/client` (the **build** counterpart to the existing `searchParams()` **reader**), so query-href construction is written functionally from the start — a structured `call` + `object-literal`, both already first-class IR. This removes the need to recognize the imperative `URLSearchParams` builder, letting the `urlBuilder` recognizer (and the adapter's builder-lowering path) be retired.

## API (shipped in #2044)

```tsx
import { queryHref } from '@barefootjs/client'

const hrefFor = (sort: string, tag: string) =>
  queryHref(base, {
[...]

PR #2049, opened 2026-06-29
fix(go-template): form-encode bf_query to match URLSearchParams (#2048)

kfly8:
## Summary

Closes the **Q5** follow-up split out of #2042 (→ #2048): aligns the go-template `bf_query` query encoder with the browser's `URLSearchParams` so a `queryHref(base, { … })` renders byte-for-byte identically across the go-template, Mojolicious, and Xslate SSR adapters and the Hono client.

`Query` (the `bf_query` runtime helper in `runtime/bf.go`) encoded keys/values with Go's `url.QueryEscape`. That agrees with `URLSearchParams` on everything **except** two characters:

| char | `URLSearchParams` (client, Perl `query`) | go `url.QueryEscape` |
|------|------------------------------------------|----------------------|
| `~`  | `%7E`                                    | `~` (kept)           |
| `*`  | `*` (kept)                               | `%2A`                |

Since each SSR adapter is conformance-compared byte-for-byte against the Hono render, a query value containing `~` or `*` rendered differently on the go adapter than on the other three backends.

## Changes

- **`runtime/bf.go`** — add `formEscape` (`application/x-www-form-urlencoded`: keep the unreserved set `A-Z a-z 0-9 * - . _`, space → `+`, every other byte → `%XX` uppercase, byte-wise UTF-8) and use it for both key and value in `Query`. This is the same byte set the Perl `_form_escape` helper already uses. `url.QueryEscape` is no longer used by `Query` (the `url` import stays — `url.Values` / `url.ParseQuery` are used elsewhere in the file).
- **`runtime/bf_test.go`** — extend `TestQuery` with the `~` / `*` / UTF-8 / `%` vectors, mirroring the Perl helper's `t/query.t` so all four backends share the same expectations.

## Behavior

Only the encoding of `~` and `*` in query keys/values changes (now `%7E` / `*`). All other characters — including space → `+`, `&` → `%26`, and UTF-8 — were already identical. No component in the repo emits a `~`/`*` query value today (`PostList` uses plain `sort`/`tag` strings), so there's no fixture or snapshot fallout.

## Test plan

- `go test ./...` in `packages/adapter-go-template/runtime` green (incl. the new `TestQuery` vectors).
- `gofmt` / `go vet` clean.
- Verified the expected strings against Node's `URLSearchParams` (`a~b*c` → `a%7Eb*c`, `100%~free*` → `100%25%7Efree*`, `café` → `caf%C3%A9`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01RkK14npEtF2TnALVYt1nec)_

### DIFF
--- a/.changeset/go-bfquery-form-encode.md
+++ b/.changeset/go-bfquery-form-encode.md
@@ -1,0 +1,7 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Encode `bf_query` keys/values with `application/x-www-form-urlencoded` (matching the browser's `URLSearchParams` and the Perl `query` helper) instead of Go's `url.QueryEscape`, so a `queryHref(base, { … })` renders byte-for-byte identically across the go-template, Mojolicious, and Xslate SSR adapters and the Hono client (#2048, follow-up to #2042).
+
+The two encoders agreed on everything except `~` and `*`: `url.QueryEscape` keeps `~` and percent-encodes `*`, whereas `URLSearchParams` percent-encodes `~` → `%7E` and keeps `*`. The new `formEscape` keeps the unreserved set `A-Z a-z 0-9 * - . _`, turns a space into `+`, and percent-encodes every other byte as `%XX` (uppercase, byte-wise UTF-8) — so query values containing `~` or `*` now match the other backends exactly.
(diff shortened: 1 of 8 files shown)


kfly8 (commented):
No blocking findings.

Priority: P0/P1 issues: none found.

I checked the Go `bf_query` encoder against `URLSearchParams` semantics for the changed `~` / `*` cases, reviewed the shared helper-vector coverage and the Go-specific empty-value divergence, and verified the Perl UTF-8 vector decoding path. The split responsibility for empty values looks intentional and covered by both the adapter lowering tests and the Go divergence pin.

Validation run:
- `env GOCACHE=/private/tmp/barefootjs-go-build-cache go test ./...` in `packages/adapter-go-template/runtime`
- `perl -Ilib t/query.t` in `packages/adapter-perl`
- `perl -Ilib t/helper_vectors.t` in `packages/adapter-perl`
- `bun test packages/adapter-tests/src/__tests__/helper-vectors.test.ts`
- `bun test packages/adapter-go-template/src/__tests__/query-href.test.ts packages/adapter-mojolicious/src/__tests__/query-href.test.ts packages/adapter-xslate/src/__tests__/query-href.test.ts packages/client/__tests__/query-href.test.ts`

Issue #2040, opened 2026-06-29
block → expression normalization (let-inline, early-return/if → ternary; raw for/mutable/break → BF101)


kfly8:
Carved out of #2018 (stage 5). The evaluator-unification core of #2018 (stages 2–3) landed and merged (P0–P5). This is the remaining "block bodies" half of the boundary defined in #2018.

## Goal

Apply the same **purely-functional-expressibility** boundary to *block* bodies (memo/derived/callback blocks), so the value-producing-but-statement-shaped forms normalize into expressions and the genuinely-imperative ones become an explicit, diagnosable `unsupported`.

## Scope (from #2018, stage 5)

- **Normalize to expression** (so they flow through the existing `ParsedExpr` path / the evaluator for callback bodies):
  - pure `let`-binding sugar → inline into the expression.
  - value-producing `if` / early-`return` → ternary in value position.
- **Refuse with a diagnostic (BF101):**
  - raw `for` accumulation / mutable local state, `break`, side-effecting / I/O APIs.
  - (A loop that mutates local state to accumulate a value is a fold — already expressible; a loop that does anything else is `unsupported`.)

## Notes

- This is **statement**-level normalization; it must not carry a general statement AST in the IR (a #2018 non-goal). The aim is to *normalize into the existing expression surface or diagnose*, not to grow a procedure interpreter.
- Pairs with the existing block-memo idiom recognition (#1897 / #1945 / #2015) — this generalizes/replaces those ad-hoc recognizers the same way the evaluator replaced the `bf_sort`/`bf_reduce` catalogue.

## Definition of done

- Value-producing `let` / `if` / early-return blocks lower without per-idiom recognition.
- Imperative blocks emit BF101 with an actionable message (and the client still runs them natively via `@client`).
- Conformance + render parity unchanged for the currently-supported corpus.

Refs: #2018.

kfly8:
Closing as completed — the block → expression normalization work has landed and merged.

**Definition of done**

- ✅ Value-producing `let` / `if` / early-return blocks lower without per-idiom recognition. The general `foldBlockToExpr` (#2051) replaced the ad-hoc block-memo recognizers, which were then retired one by one: filter predicate (#2052), guard-const memo (#2053), and searchParams object memo (#2054).
[...]

PR #2051, opened 2026-06-30
feat(jsx): normalize value-producing block-bodied callbacks to expressions (#2040)

kfly8:
Carved from #2018 stage 5 — the **block bodies** half of the purely-functional-expressibility boundary. Closes the callback portion of #2040.

## What

A higher-order callback (`.sort` / `.reduce` / `.find` / `.some` / `.every` / `.flatMap` …) written with a `{ … }` block body now **normalizes to a single expression** when its body is purely-functionally expressible, instead of being refused with *"only single-`return` block-body functions are supported"*:

- **let-inline** — pure `const` bindings inline into the expression that uses them.
- **early-return / value `if`** — `if (c) return A; return B` (incl. `if/else` and `else if` chains) folds to a ternary in value position.

The folded result flows through the existing `ParsedExpr` surface (the #2018 callback evaluator / template-native lowering), so **no new IR statement shapes are carried** — matching the issue's non-goal of growing a procedure interpreter.

Genuinely **imperative** block bodies — raw `for` / `while`, `break`, local re-assignment / mutable state, side-effecting or I/O calls — have no value-position lowering and stay `unsupported`, so adapters surface **BF101** with an actionable message (rewrite an accumulation loop as `.reduce(...)`, or move the body to a `/* @client */` value that runs natively on the client).

## How

- New public helper `foldBlockToExpr(ParsedStatement[])` in `expression-parser.ts` performs the fold (let-inline substitution + `if`/early-return → ternary, with branch-termination analysis for the fall-through/early-return idiom).
- `convertNode`'s arrow path uses it for multi-statement block bodies. The single-`return` fast path is unchanged, so existing output is **byte-identical**.

## Scope / follow-up

This lands the **callback** half of #2040. The **memo / derived** block-recognizer replacement (#1897 / #1945 / #2015) is deliberately **not** in this PR: the Go adapter's memo lowering checks `MemoInfo.parsed` before its `parsedBlock` recognizers (`memo-compute.ts`), while `template-interp.ts` checks `parsedBlock` first — so populating `parsed` from a fold would reroute memo emission and needs per-adapter (Go + shared Perl) parity work, ideally alongside the mojo/xslate decomposition from #2018 track D. `foldBlockToExpr` is the reusable core that work will build on.

## Testing

- `foldBlockToExpr` unit tests (let-inline, chained let, early-return ternary, `if/else`, `else if` chain, fall-through refusal, imperative refusal).
- Updated `expression-parser` / `unsupported-expression` tests that previously asserted the old "single-`return` only" restriction.
- Updated `docs/core/rendering/jsx-compatibility.md` (the let-inline sort comparator now lowers; the ❌ example is now a genuinely imperative comparator).
- Verified green: `@barefootjs/jsx`, `go-template`, `mojolicious`, `xslate` unit suites. Conformance suite unchanged — the only failures (carousel cross-adapter render) are **pre-existing on `main`** (confirmed by re-running with the change stashed).

Refs: #2040, #2018.

---

### DIFF
--- a/packages/jsx/src/expression-parser.ts
+++ b/packages/jsx/src/expression-parser.ts
@@ -2585,63 +2585,194 @@ function statementsTerminate(stmts: ParsedStatement[]): boolean {
 }
 
 /**
- * Substitute free identifiers in `expr` with their bound expression from `env`
- * (the let-inline half of {@link foldBlockToExpr}). A nested `arrow`'s own
- * params shadow the environment for its body, so a bound name that collides with
- * a param is left untouched inside that body. Mirrors the structural walk used
- * by `substituteDestructuredFields`; every `ParsedExpr` kind is handled so a new
- * kind surfaces as a compile error here rather than silently skipping
- * substitution.
+ * Refusal reason when inlining a `const` binding would capture a free variable
+ * of its initializer under a nested callback parameter of the same name. The
+ * let-inline substitution is not a hygienic (alpha-renaming) substitution, so
+ * rather than silently miscompile the callback we refuse and adapters surface
+ * BF101.
  */
-function substituteIdents(
+export const CAPTURE_BLOCK_REASON =
+  'Block body cannot be normalized: inlining a `const` binding would capture ' +
+  'one of its free variables under a nested callback parameter of the same ' +
+  'name (e.g. `const x = a; … list.map(a => a + x)`). Rename the inner ' +
+  'parameter, or move the body to a `/* @client */` value so it runs natively ' +
+  'on the client.'
+
+/**
+ * Refusal reason when a `const` initializer may have side effects (it contains
+ * a function/method call) and the binding is NOT used exactly once on a single
+ * runtime path. Let-inline substitutes the initializer at each use site, which
+ * would drop the effect (zero uses) or duplicate it (multiple uses / a use
+ * inside a callback that runs per element) — exactly the side-effecting shape
+ * the fold must refuse rather than miscompile.
+ */
+export const IMPURE_INLINE_BLOCK_REASON =
+  'Block body cannot be normalized: a `const` whose initializer may have side ' +
+  'effects (a function or method call) is used zero or more than once, so ' +
+  'inlining it would drop or duplicate the effect. Bind a pure value, use the ' +
+  'binding exactly once, or move the body to a `/* @client */` value so it ' +
+  'runs natively on the client.'
+
+/**
+ * Whether an expression is provably free of side effects by syntax alone, so it
+ * is safe to inline at any number of use sites. Conservative: any function or
+ * method call (or a value the parser couldn't represent) is treated as possibly
+ * impure. Member access is treated as pure, matching `substituteDestructuredFields`
+ * and the rest of the compiler's expression handling.
+ */
+function isPureInit(e: ParsedExpr): boolean {
+  switch (e.kind) {
+    case 'identifier':
+    case 'literal':
+    case 'regex':
+      return true
+    case 'member':
+      return isPureInit(e.object)
+    case 'index-access':
+      return isPureInit(e.object) && isPureInit(e.index)
+    case 'binary':
+    case 'logical':
+      return isPureInit(e.left) && isPureInit(e.right)
+    case 'unary':
+      return isPureInit(e.argument)
+    case 'conditional':
+      return isPureInit(e.test) && isPureInit(e.consequent) && isPureInit(e.alternate)
+    case 'template-literal':
+      return e.parts.every(p => p.type !== 'expression' || isPureInit(p.expr))
+    case 'array-literal':
+      return e.elements.every(isPureInit)
+    case 'object-literal':
+      return e.properties.every(p => isPureInit(p.value))
+    // A call / method call may be effectful or non-deterministic; an arrow value
+    // can capture impurity; `unsupported` is opaque. Treat all as possibly impure.
+    case 'call':
+    case 'array-method':
+    case 'arrow':
+    case 'unsupported':
+      return false
+  }
+}
+
+/**
+ * Maximum number of times `name` is evaluated on a single runtime path through
+ * `expr`. A `conditional` only ever evaluates one branch, so its branch cost is
+ * the max (not the sum) of the two arms plus the always-evaluated test. A use
+ * inside a nested `arrow` body may run any number of times (the callback is
+ * invoked per element / comparison), so it counts as "more than once" — modelled
+ * as `Infinity` — which forces an impure binding referenced from a callback to
+ * be refused. Used to decide whether inlining a possibly-impure init is
+ * evaluation-count-preserving.
+ */
+function maxUsesPerPath(name: string, expr: ParsedExpr): number {
+  const walk = (e: ParsedExpr): number => {
+    switch (e.kind) {
+      case 'identifier':
+        return e.name === name ? 1 : 0
+      case 'literal':
+      case 'regex':
+      case 'unsupported':
+        return 0
+      case 'member':
+        return walk(e.object)
+      case 'index-access':
+        return walk(e.object) + walk(e.index)
+      case 'binary':
+      case 'logical':
+        return walk(e.left) + walk(e.right)
+      case 'unary':
+        return walk(e.argument)
+      case 'conditional':
+        return walk(e.test) + Math.max(walk(e.consequent), walk(e.alternate))
+      case 'template-literal':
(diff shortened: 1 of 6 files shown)


packages/jsx/src/expression-parser.ts
```diff
@@ -2518,6 +2542,194 @@ function parseStatement(
...
+          kind: 'template-literal',
+          parts: e.parts.map(p =>
+            p.type === 'expression' ? { type: 'expression', expr: walk(p.expr) } : p,
+          ),
+        }
+      case 'arrow': {
+        // Params shadow the outer let-bindings inside the body.
+        const inner = new Map(env)
+        for (const p of e.params) inner.delete(p)
+        return { kind: 'arrow', params: e.params, body: substituteIdents(e.body, inner) }
```
kfly8: **Variable capture: substitution is not hygienic.** Deleting the arrow's own params from `env` prevents the params from being substituted, but it does *not* protect the free identifiers in a substituted-in expression from being captured by those params.

Confirmed against the PR build:

```ts
foldBlockToExpr(parse('{ const x = a; return list.map(a => a + x) }'))
// => list.map(a => a + a)      ❌  should be  list.map(a => a + <outer a>)
```

The outer `a` bound to `x` gets captured by the inner `map` param `a`, silently miscompiling the callback. Reachable through the real arrow path, e.g. `(a, b) => { const x = a.f; return list.map(a => a + x) }`.

A correct fold needs to refuse (or alpha-rename) when a substituted expression contains a free identifier that any shadowing param in the substitution target would capture.

---

packages/jsx/src/expression-parser.ts
```diff
@@ -2518,6 +2542,194 @@ function parseStatement(
...
+  env: ReadonlyMap<string, ParsedExpr>,
+): { ok: true; expr: ParsedExpr } | { ok: false; reason: string } {
+  if (stmts.length === 0) {
+    return { ok: false, reason: IMPERATIVE_BLOCK_REASON }
+  }
+  const [head, ...rest] = stmts
+  switch (head.kind) {
+    case 'var-decl': {
+      const next = new Map(env)
+      next.set(head.name, substituteIdents(head.init, env))
```
kfly8: **Purity of `const` inits is not actually enforced, so let-inline changes evaluation count.** The docstring (L2667) says "The binding is pure by construction", but `parseStatement` accepts any expression init, including a call. Let-inline then substitutes the init at each use site, which is only sound when the init is pure.

Confirmed against the PR build:

```ts
foldBlockToExpr(parse('{ const d = next(); return d + d }'))
// => next() + next()     ❌  double-evaluates an effectful/non-deterministic call

foldBlockToExpr(parse('{ const _ = log(); return a - b }'))
// => a - b               ❌  unused binding's side effect is dropped entirely
```

For the typical pure comparator this is harmless, but the compiler silently miscompiles when the init is effectful or non-deterministic (`Math.random()`, a counter, a logger) — exactly the "side-effecting / I/O call" shape the PR claims to refuse. Consider either restricting let-inline to inits provably free of calls (or used at most once), or rejecting call-bearing inits as imperative so BF101 fires instead of emitting wrong code.

---

kfly8 (commented):
Reviewed the block→expression fold (`foldBlockToExpr` + `substituteIdents`). The fast-path/byte-identical design is nice and the early-return → ternary folding looks correct. Two soundness gaps in the let-inline substitution, both confirmed against this branch's build — details inline.

---

kfly8:
Both soundness gaps fixed in `5b01e7f`. Thanks — both were real miscompiles, confirmed and now covered by regression tests.

**1. Variable capture (hygiene).** The let-inline walk now tracks enclosing callback params and refuses (→ BF101) when inlining a binding would place a free variable of its value under a nested param of the same name. `{ const x = a; return list.map(a => a + x) }` and the `(a, b) => { const x = a.f; return list.map(a => a + x) }` arrow-path variant now both refuse instead of folding to `a + a`.

**2. Impure init duplication / drop.** `foldBlockToExpr` is restructured to fold the rest first (leaving each binding free), measure its **per-path** use count, then inline. A possibly-impure init (one containing a call) is inlined **only when used exactly once on a single runtime path**; zero uses (dropped effect) or >1 (incl. a use inside a callback that runs per element) now refuse. A pure-by-syntax init (identifiers, literals, member access, operators) still inlines at any count, so the canonical `const f = filter(); …` signal-getter and member-access cases are unchanged. The L2667 docstring's "pure by construction" claim is gone.

Concretely:
- `{ const d = next(); return d + d }` → refused (was `next() + next()`)
- `{ const _ = log(); return a - b }` → refused (dropped effect)
- `{ const d = next(); return arr.map(x => x + d) }` → refused (per-element eval)
- `{ const d = next(); return d }` → still folds to `next()` (eval-count preserved)
- `{ const f = filter(); if (f === 'active') return !t.done; return true }` → still folds

Verified green: `@barefootjs/jsx` (only the 4 pre-existing checker-alias failures remain), `go-template`, `mojolicious`, `xslate` unit suites.

---

packages/jsx/src/expression-parser.ts
```diff
@@ -2518,6 +2542,344 @@ function parseStatement(
...
+      // the rest are inlined by the enclosing `var-decl` frames; references to
+      // `head.name` inside `head.init` cannot occur (a `const` can't read itself).
+      const restFold = foldBlockToExpr(rest)
+      if (!restFold.ok) return restFold
+      const uses = maxUsesPerPath(head.name, restFold.expr)
+      // A possibly-impure init is only safe to inline when it is evaluated
+      // exactly once on a single path — same as the original block. A pure init
+      // is safe at any count (including zero: dropping a pure unused binding is
+      // observationally identical).
+      if (!isPureInit(head.init) && uses !== 1) {
```
kfly8: Both earlier findings are now correctly refused (capture / double-eval / dropped-unused) — nice. One gap remains in this same impure-init check.

`maxUsesPerPath` returns the **max** uses across runtime paths, so the guard only catches *duplication* (`uses > 1`) and the *uniformly-unused* case (`uses === 0`). It misses an impure init that is used on **some** paths but not others — `max` is `1`, yet the binding is dropped on the path that doesn't use it. The original block evaluates `const d = next()` **unconditionally exactly once**, so moving it into a single ternary arm drops the effect on the other arm.

Confirmed against this branch's build:

```ts
foldBlockToExpr(parse('{ const d = next(); if (c) return d; return 0 }'))
// => c ? next() : 0      ❌  next() is skipped when c is false (original always calls it once)

foldBlockToExpr(parse('{ const d = next(); if (c) return 0; return d }'))
// => c ? 0 : next()      ❌  next() is skipped when c is true
```

This is the same "dropped effect" class the fix targets, just reached through a conditional rather than an unused binding. The sound condition for inlining a possibly-impure init is "evaluated exactly once on **every** path" (min == max == 1), not "max == 1". Tracking a min-uses-per-path alongside `maxUsesPerPath` and refusing when `min !== 1 || max !== 1` would close it. (A pure init is unaffected — drop/duplicate is unobservable there.)

---

kfly8 (commented):
Re-reviewed after `5b01e7f`. The capture fix (`inlineBinding` + `CAPTURE_BLOCK_REASON`) and the impure-init guard look solid for the duplication and unused-binding cases — both original findings are now refused, and pure / used-once inits still fold as before. One residual gap in the impure-init use-count check (impure init used conditionally) — detail inline.

---

kfly8:
Fixed in `e63d726`. You're right — `max == 1` only rules out duplication; an impure init used on *some* paths but not others still drops the effect on the skipped path, even though the original block runs the initializer unconditionally once.

Replaced `maxUsesPerPath` with `usesPerPath` returning `{ min, max }`, and a possibly-impure init now inlines only when `min === 1 && max === 1` (evaluated exactly once on **every** path). `min` accounts for every skippable position:
- a ternary arm (used in one side only) → `min` 0
- a short-circuited `&&` / `||` / `??` right operand → contributes to `max`, not `min`
- a nested callback body → `min` 0 / `max` Infinity

Confirmed:
- `{ const d = next(); if (c) return d; return 0 }` → refused (was `c ? next() : 0`)
- `{ const d = next(); if (c) return 0; return d }` → refused
- `{ const d = next(); return c && d }` → refused (short-circuit drop)
- `{ const d = next(); return d && c }` → still folds (`next()` always runs once, left operand)
- pure inits and the canonical `const f = filter(); if (f === 'active') …` (read once in the always-evaluated test) → unchanged

Added regression tests for the then-only / else-only / short-circuit drops plus the unconditional-left case. Suites green: jsx (only the 4 pre-existing checker-alias failures), go-template, mojolicious, xslate.

---

Issue #1665, opened 2026-05-29
Module-level helper called inside a reactive JSX child throws 'ReferenceError: <helper> is not defined' at hydration


kfly8:
## Summary

Calling a **module-level helper function from inside a reactive JSX child expression** (one compiled into a `createEffect`, e.g. via `cond() && helper()` inside a `.map`) throws `ReferenceError: <helper> is not defined` at hydration. The identical call inside a **non-reactive** child works, so the helper IS emitted — it just isn't in scope of the generated reactive effect.

## Environment
- `@barefootjs/cli@0.4.0`, adapter: `@barefootjs/hono`
- Runtime: Bun + `wrangler dev`

## Repro
```tsx
'use client'
import { BrandLogo } from './brand-logo'
import type { ThemeId } from './types'

const THEMES = [{ id: 'piconic' }, { id: 'hono' }] as const

function themeLogo(id: ThemeId) {        // module-level helper, returns JSX
  if (id === 'hono') return <BrandLogo />
  return <BrandLogo />
}

export function Header(props: { sel: ThemeId }) {
  return (
    <span>
      {/* reactive child: `&&` reads props.sel -> compiled into a createEffect */}
      {THEMES.map(t => props.sel === t.id && themeLogo(t.id))}
    </span>
  )
}
```

**Observed (browser, at hydration):**
```
ReferenceError: themeLogo is not defined
    at .../App.client.js:...    (inside Array.map)
    at runEffect (.../barefoot.js)
    at createEffect (.../barefoot.js)
    at initHeader (...)
```

**Contrast:** the same `{themeLogo(t.id)}` used in a *non-reactive* child (a plain list `.map` rendered once) resolves `themeLogo` correctly. Only the reactive-effect path fails to capture the module-level binding.

## Expected
Module-level functions referenced inside reactive child expressions should be reachable from the generated effect scope (same as in non-reactive children).

## Workaround
Inline the element instead of calling the helper in reactive children:
```tsx
{props.sel === 'hono' && <BrandLogo />}
```
(which also forces duplicating the id→logo mapping that the helper was meant to centralize).

## Note
Possibly related: bare function calls returning elements in JSX children are not re-evaluated reactively (only ternary / `&&` / `.map` are treated as reactive control flow). The reactive `&&` form here is the documented-reactive shape, yet it can't reach the helper.

PR #2053, opened 2026-06-30
feat(jsx): fold guard-and-return-const block memos to expressions (#2040)

kfly8:
**PR-B of the #2040 memo follow-up stack.** Base: **`main`** (PR-A #2052 merged; rebased — this PR is now just PR-B's single commit on top of main).

## Context

PR-A retired the filter block-condition renderers. PR-B does the same for the **guard-and-return-const memo** (#1897 / #1945) — `() => { const k = getter(); if (!k) return MODULE_CONST; … }`.

As discussed on the issue: a memo recognizer isn't just shape-matching — it **partial-evaluates** the body with signals at their initial values to bake the SSR value. The clean fix (per the DoD) is to fold the block to an expression and let the existing evaluator handle it, not to grow the statement matcher.

## How

- **Analyzer fold** — a complete, value-producing block memo is folded into a single `MemoInfo.parsed` expression via `foldBlockToExpr`, in a post-pass after all signals/memos are collected. The reactive-getter set is the purity oracle, so an idempotent `const k = getter()` read may be inlined on several branches. Incomplete/unfoldable blocks leave `parsed` undefined and keep the `parsedBlock` fallback.
- **Go recognizer rewrite** — `resolveBlockBodyMemoModuleConst` now reads the folded `parsed` conditional (`!getter() ? MODULE_CONST : <derived>`) instead of walking `var-decl`/`if`/`return` statements with a local-var→signal map. The bespoke statement matcher is gone; the guard-falsy-init → module-const baking is unchanged. All three call sites (initial value, field type, loop-item type) pass `memo.parsed`.

The folded shape for the canonical case:
```
(!sel()) ? ALL : ALL.filter(t => t === sel())
```
`sel` starts `null` (falsy) ⇒ `!sel()` is `true` ⇒ SSR value is the module const `ALL`, baked to a Go slice exactly as before.

## Testing

- New analyzer tests: guard memo folds to a `conditional` `parsed`; imperative (`for`-loop) block leaves `parsed` unset.
- **Adapter conformance (Go + Perl): 780 pass, 9 fail** — the 9 are the pre-existing carousel failures on `main`.
- **Go / Mojo / Xslate adapter unit suites: green** (558 / 843 pass, 0 fail).
- **jsx suite: 2163 pass**, only the 4 pre-existing checker-alias failures.

Note: `parsedBlock` is still consumed by the searchParams-object memo recognizer (PR-C) and the template-literal block memo (template-interp); it is removed in a later stack PR.

Refs: #2040, #2018.

### DIFF
--- a/packages/adapter-go-template/src/adapter/memo/memo-compute.ts
+++ b/packages/adapter-go-template/src/adapter/memo/memo-compute.ts
@@ -295,7 +295,7 @@ export function computeMemoInitialValueOrNull(
   // is falsy: `() => { const k = getter(); if (!k) return MODULE_ARRAY; … }`.
   // When the signal starts null the SSR value is that array; its literal value
   // (not the identifier) is passed to the baker so it reduces to a Go slice.
-  const blockReturn = resolveBlockBodyMemoModuleConst(ctx, memo.parsedBlock, signals)
+  const blockReturn = resolveBlockBodyMemoModuleConst(ctx, memo.parsed, signals)
   if (blockReturn !== null && blockReturn.constValue && blockReturn.constType) {
     return convertInitialValue(ctx,
       blockReturn.constValue,
--- a/packages/adapter-go-template/src/adapter/memo/memo-compute.ts
+++ b/packages/adapter-go-template/src/adapter/memo/memo-compute.ts
@@ -295,7 +295,7 @@ export function computeMemoInitialValueOrNull(
   // is falsy: `() => { const k = getter(); if (!k) return MODULE_ARRAY; … }`.
   // When the signal starts null the SSR value is that array; its literal value
   // (not the identifier) is passed to the baker so it reduces to a Go slice.
-  const blockReturn = resolveBlockBodyMemoModuleConst(ctx, memo.parsed, signals)
+  const blockReturn = resolveBlockBodyMemoModuleConst(ctx, memo, signals)
   if (blockReturn !== null && blockReturn.constValue && blockReturn.constType) {
     return convertInitialValue(ctx,
       blockReturn.constValue,
@@ -367,8 +367,8 @@ export function computeComparisonTernaryGo(
   propsParams: { name: string; type?: TypeInfo; defaultValue?: string }[],
   propFallbackVars: ReadonlyMap<string, PropFallbackVar>,
 ): string | null {
-  // Only an expression-bodied conditional matches; a block-bodied memo has no
-  // `parsed`, so it returns null.
+  // Matches any `conditional` `parsed` — an expression-bodied ternary or, since
+  // #2040, a block-bodied memo whose value `if` / early-return folded to one.
   if (!parsed || parsed.kind !== 'conditional') return null
   const cond = parsed.test
   if (cond.kind !== 'binary' || !(cond.right.kind === 'literal' && cond.right.literalType === 'string')) {
--- a/packages/jsx/src/analyzer.ts
+++ b/packages/jsx/src/analyzer.ts
@@ -21,6 +21,7 @@ import {
   membersToProperties,
   isComponentFunction,
   isArrowComponentFunction,
+  collectReactiveGetterNames,
 } from './analyzer-context.ts'
 import { createError, createWarning, ErrorCodes } from './errors.ts'
 import path from 'node:path'
@@ -286,9 +287,7 @@ export function analyzeComponent(
   // parser dropped a statement it couldn't represent) or one that doesn't fold
   // (imperative residue) leaves `parsed` undefined and consumers keep their
   // existing `parsedBlock` fallback.
-  const reactiveGetterNames = new Set<string>()
-  for (const s of ctx.signals) reactiveGetterNames.add(s.getter)
-  for (const m of ctx.memos) reactiveGetterNames.add(m.name)
+  const reactiveGetterNames = collectReactiveGetterNames(ctx.signals, ctx.memos)
   for (const memo of ctx.memos) {
     if (memo.parsed || !memo.parsedBlock || !memo.parsedBlockComplete) continue
     const folded = foldBlockToExpr(memo.parsedBlock, { pureCallNames: reactiveGetterNames })
--- a/packages/jsx/src/analyzer.ts
+++ b/packages/jsx/src/analyzer.ts
@@ -8,7 +8,7 @@
 
 import ts from 'typescript'
 import type { ImportSpecifier, TypeInfo, ParamInfo, ReactiveFactoryInfo } from './types.ts'
-import { parseExpression, parseBlockBodyTolerant } from './expression-parser.ts'
+import { parseExpression, parseBlockBodyTolerant, foldBlockToExpr } from './expression-parser.ts'
 import { rewriteBarePropRefs } from './prop-rewrite.ts'
 import { incrementCounter } from './instrumentation.ts'
 import {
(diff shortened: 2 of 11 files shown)


kfly8 (commented):
## レビュー: guard-and-return-const ブロックメモの式へのフォールド (#2040 PR-B)

per-idiom な statement マッチャを廃し、解析器の一般的な fold (`foldBlockToExpr`) で `MemoInfo.parsed` に正規化してから Go レコグナイザに食わせる方向性は altitude として良い改善です。クライアント JS は `memo.computation` を見ているため影響なし、Mojo/Xslate のメモ経路も無関係であることを確認しました。

その上で3点コメントしています（深刻度順）:

1. **[regression / 要確認]** `memo-compute.ts:298` — ガードと並んで不純なローカル束縛があると fold が拒否され、SSR の const ベイクが `main` に対して後退（初期描画でリストが空に）。fixture 未カバー。
2. **[scope / invariant]** `analyzer.ts:295` — `parsed` を全 foldable ブロックに設定することで、`isStringTernaryMemo` / `memoInitialFromParsedBody` / `computeComparisonTernaryGo` など「ブロックメモに parsed は無い」前提の消費者の挙動も変化（多くは改善方向だがテスト無し・docstring の不変条件が偽に）。
3. **[cleanup]** `analyzer.ts:289` — リアクティブゲッター集合の構築が既存 `getReactiveGetterNames` と重複。

1 が唯一 `main` より厳密に悪い出力を生む可能性のある項目ですが、現行 fixture が触れない形に依存するため要確認（plausible）です。それ以外は概ね妥当で、設計の方向性は支持します。

---

packages/adapter-go-template/src/adapter/memo/memo-compute.ts
```diff
@@ -295,7 +295,7 @@ export function computeMemoInitialValueOrNull(
   // is falsy: `() => { const k = getter(); if (!k) return MODULE_ARRAY; … }`.
   // When the signal starts null the SSR value is that array; its literal value
   // (not the identifier) is passed to the baker so it reduces to a Go slice.
-  const blockReturn = resolveBlockBodyMemoModuleConst(ctx, memo.parsedBlock, signals)
+  const blockReturn = resolveBlockBodyMemoModuleConst(ctx, memo.parsed, signals)
```
kfly8: **[regression / needs verification] An impure local binding alongside the guard drops the SSR const bake**

Switching from `memo.parsedBlock` to `memo.parsed` means any block that `foldBlockToExpr` refuses to fold ends up with `memo.parsed === undefined`, so `resolveBlockBodyMemoModuleConst` here returns null and nothing is baked.

```tsx
const visible = createMemo(() => {
  const k = sel()
  const z = helper()                 // non-reactive (not in pureCallNames)
  if (!k) return ALL
  return ALL.filter(t => g(t, z))     // z used in only one arm
})
```

Here `parsedBlockComplete === true`, but `z = helper()` is "impure and not used exactly once on every path", so `foldBlockToExpr` returns `{ok:false}` at `expression-parser.ts:2896` and `memo.parsed` stays unset. As a result `computeMemoInitialValueOrNull` returns null → the SSR initial value falls back to a nil/zero slice, whereas the old statement-walker on `main` baked `ALL`. The first server render shows an empty list (it recovers after client hydration).

The old walker ignored `const z` and baked `ALL`, so this is a behavior regression vs. `main`. No conformance fixture covers this shape, so the green suite doesn't catch it — recommend adding one under `packages/adapter-tests/fixtures/`.

---

packages/jsx/src/analyzer.ts
```diff
@@ -277,6 +277,24 @@ export function analyzeComponent(
...
+  // parser dropped a statement it couldn't represent) or one that doesn't fold
+  // (imperative residue) leaves `parsed` undefined and consumers keep their
+  // existing `parsedBlock` fallback.
+  const reactiveGetterNames = new Set<string>()
+  for (const s of ctx.signals) reactiveGetterNames.add(s.getter)
+  for (const m of ctx.memos) reactiveGetterNames.add(m.name)
+  for (const memo of ctx.memos) {
+    if (memo.parsed || !memo.parsedBlock || !memo.parsedBlockComplete) continue
+    const folded = foldBlockToExpr(memo.parsedBlock, { pureCallNames: reactiveGetterNames })
+    if (folded.ok) memo.parsed = folded.expr
```
kfly8, line 294: **[scope / invariant] Setting `memo.parsed` on every foldable block also changes other consumers beyond the guard-const recognizer**

Every foldable block memo now gets `parsed` set, but several Go-adapter consumers were written assuming "a block-bodied memo has no `parsed`", and that assumption is now false:

- `memo/memo-type.ts:61` — docstring literally states *"A block-bodied memo has no `parsed`, so it returns false."* (`isStringTernaryMemo`)
- `memo/memo-compute.ts:370` — same assumption (`computeComparisonTernaryGo`)
- `memo/memo-compute.ts:271` — `memoInitialFromParsedBody(memo.parsed)` runs **before** `resolveBlockBodyMemoModuleConst`

Example: `() => { if (vert()) return A; return B }` (A/B are module-scope string consts) folds to `vert() ? A : B`. On `main` the SSR initial value was the zero value; now `memoInitialFromParsedBody` bakes `B`, and `isStringTernaryMemo` flips the inferred Go field type from `interface{}` to `string`. These changes mostly look *more* correct, but they're outside this PR's stated scope (the guard-const recognizer), untested, and they silently falsify the two docstring invariants above. Recommend adding conformance fixtures for early-return string-const block memos and updating the two docstrings.

---

packages/jsx/src/analyzer.ts
```diff
@@ -277,6 +277,24 @@ export function analyzeComponent(
...
+  // #2040: fold a complete, value-producing block-bodied memo into a single
+  // expression so it flows through the same `parsed` path as an expression-bodied
+  // memo, instead of relying on per-idiom block recognizers (#1897 / #1945 /
+  // #2015). Runs after all signals/memos are collected so the reactive-getter
+  // set (used as the purity oracle — an idempotent signal/memo read may be
+  // inlined on several branches) is complete. An incomplete block (the tolerant
+  // parser dropped a statement it couldn't represent) or one that doesn't fold
+  // (imperative residue) leaves `parsed` undefined and consumers keep their
+  // existing `parsedBlock` fallback.
+  const reactiveGetterNames = new Set<string>()
```
kfly8: **[cleanup] The reactive-getter set duplicates an existing helper**

```ts
const reactiveGetterNames = new Set<string>()
for (const s of ctx.signals) reactiveGetterNames.add(s.getter)
for (const m of ctx.memos) reactiveGetterNames.add(m.name)
```

This is logically identical to `getReactiveGetterNames` in `jsx-to-ir.ts:196-203` (`signals → .getter` / `memos → .name`, over the same `AnalyzerContext`). The definition of "what counts as a reactive getter" now lives in two places; if a third reactive kind is ever added, one site can drift and mis-classify purity for the fold. Suggest extracting `collectReactiveGetterNames(signals, memos)` and calling it from both.

---

kfly8 (commented):
## Review: fold guard-and-return-const block memos to expressions (#2040, PR-B)

_(English version of my earlier review — please refer to these comments; the prior Japanese ones can be ignored.)_

Replacing the per-idiom statement matcher with a general analyzer fold (`foldBlockToExpr` → `MemoInfo.parsed`) and feeding that to the Go recognizer is a good altitude improvement. I confirmed the client JS path is unaffected (it consumes `memo.computation`, not `parsed`), and the Mojo/Xslate memo paths are untouched.

Three comments, most-severe first:

1. **[regression / needs verification]** `memo-compute.ts:298` — when an impure local binding sits alongside the guard, the fold is refused and the SSR const bake regresses vs. `main` (empty list on first render). Not covered by any fixture.
2. **[scope / invariant]** `analyzer.ts:295` — setting `parsed` on every foldable block also changes consumers (`isStringTernaryMemo` / `memoInitialFromParsedBody` / `computeComparisonTernaryGo`) that assume block memos have no `parsed`. Mostly toward more-correct output, but untested and it falsifies two docstring invariants.
3. **[cleanup]** `analyzer.ts:289` — the reactive-getter set duplicates the existing `getReactiveGetterNames` helper.

Only #1 can produce strictly worse output than `main`, and it depends on a memo shape no current fixture exercises (hence plausible rather than observed). Otherwise the direction is sound and I'm in favor of the approach.

---

kfly8:
All three addressed in `ce43523`. Thanks — point 1 was a real regression; I reproduced it before fixing.

**1. [regression] impure tail binding drops the const bake** — confirmed: `{ const k = sel(); const n = ext(); if (!k) return ALL; return ALL.filter(t => t.length === n).concat(ALL.slice(n)) }` gives `parsedBlockComplete === true` but `foldBlockToExpr` refuses (`n` impure, used twice on one path) → `parsed` unset → no bake on `main`+this-PR-as-was. Fixed by making `resolveBlockBodyMemoModuleConst` take the memo and try the folded `parsed` conditional first, then **fall back to the tolerant `parsedBlock` guard-prefix scan**. The bake depends only on the guard prefix, so the fallback reproduces `main` exactly for blocks the fold can't fully normalize. Added a regression test that bakes `[]interface{}{"a","b","c"}` for this exact shape.

**2. [scope/invariant] `parsed` now set on block memos** — updated the three docstrings that claimed "block memos have no `parsed`" (`MemoInfo.parsed`, `isStringTernaryMemo`, `computeComparisonTernaryGo`) and added a test for the early-return string-const block memo (`() => { if (vert()) return A; return B }` → bakes `B`, field type `string`). The behavior changes are the "more correct" direction you noted and are now covered.

**3. [cleanup] duplicate getter set** — extracted `collectReactiveGetterNames(signals, memos)` into `analyzer-context.ts`; both the analyzer fold and `jsx-to-ir`'s `getReactiveGetterNames` call it now.

Verified: Go adapter unit **561 pass / 0 fail** (one `#1665` loop-item-conditional flake under parallel `go run` load — passes in isolation), conformance **780 pass / 9 carousel-only**, jsx **2165 pass** (4 pre-existing checker-alias). The fold direction is unchanged; the fallback only catches blocks that genuinely can't fold.

---

PR #2054, opened 2026-06-30
feat(go-template): fold searchParams object memos to expressions (#2040)

kfly8:
**Stack PR-C of the #2040 memo follow-up.** Base: **`claude/2040-B-memo-guard-const-fold` (PR #2053)** — review/merge PR-B first; this diff is only PR-C's commit on top.

## What

Retires the last per-idiom block-memo recognizer's statement walk — the searchParams-derived object memo (#2015) — by lowering it through the general fold.

```tsx
createMemo(() => {
  const sp = searchParams()
  return { sort: asSortKey(sp.get('sort')), tag: sp.get('tag') ?? '' }
})
```

## How

- `computeObjectMemoInitialValue` previously walked `parsedBlock` for `const sp = searchParams()` bindings + a terminal `return { … }`. It now folds the block with `foldBlockToExpr`, adding `searchParams` to the purity oracle — an idempotent request-query read, safe to inline at each `sp.get('k')` site — and lowers the resulting object-literal. `sp` inlines to `searchParams()`.
- `lowerCtorExpr` now recognises a `searchParams().get('k')` receiver (the inlined form) in addition to the `const sp` env form.
- `foldBlockToExpr` / `predicateTernaryToLogical` / `FoldBlockOptions` are exported from `@barefootjs/jsx` so the adapter can fold.

searchParams-purity knowledge stays in the Go adapter (where `searchParamsLocals` already lives — covering aliased imports), so there's no analyzer coupling.

## Side benefit

An object memo calling `searchParams().get('k')` **directly** (no `const` binding) now lowers too — the old walk required the binding to populate its env. A block that doesn't fold to an object literal returns null → the same nil fallback as before.

## Testing

- New tests: direct-`searchParams()` object memo, aliased `searchParams as sp` import.
- **Go adapter unit: 563 pass / 0 fail.**
- **Conformance (Go + Perl): 780 pass, 9 fail** — the 9 are the pre-existing carousel failures on `main`.
- **jsx: 2165 pass**, only the 4 pre-existing checker-alias failures.

Note: `parsedBlock` is still consumed by the template-literal block memo (template-interp) and the guard-const fallback; its full removal from the IR is the remaining stack PR (PR-D).

Refs: #2040, #2018, #2015.

---

### DIFF
--- a/.changeset/searchparams-object-fold.md
+++ b/.changeset/searchparams-object-fold.md
@@ -1,0 +1,10 @@
+---
+"@barefootjs/jsx": patch
+"@barefootjs/go-template": patch
+---
+
+Lower the searchParams-derived object memo (#2015) through the general fold instead of a bespoke statement walk (#2040, PR-C of the memo follow-up stack).
+
+`computeObjectMemoInitialValue` previously walked `parsedBlock` for `const sp = searchParams()` bindings + a terminal `return { … }`. It now folds the block with `foldBlockToExpr`, adding `searchParams` to the purity oracle (an idempotent request-query read, safe to inline at each `sp.get('k')` site), and lowers the resulting object-literal. `sp` is inlined to `searchParams()`, so `lowerCtorExpr` now recognises a `searchParams().get('k')` receiver in addition to the `const sp` env form. `foldBlockToExpr` is exported from `@barefootjs/jsx`.
+
+This drops the statement-shape matching (var-decl scan + last-return check) for the object memo, and as a side benefit lowers an object memo that calls `searchParams().get('k')` directly without a `const` binding. A block that doesn't fold to an object literal returns null → the same nil fallback as before. Render parity verified by the Go adapter conformance + unit suites.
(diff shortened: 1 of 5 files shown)


kfly8 (commented):
No blocking findings.

Priority: P0/P1 issues: none found.

I reviewed the object-memo folding path, especially the new `foldBlockToExpr(..., { pureCallNames: searchParamsLocals })` use and the expanded `searchParams().get('k')` constructor lowering. The fallback behavior still looks conservative: unfolded/control-flow shapes that do not normalize to an object literal continue to return `nil`, and the direct/aliased `searchParams` cases are covered by the added tests.

Validation run:
- `bun test packages/adapter-go-template/src/__tests__/derived-state-memo.test.ts`
- `bun test packages/jsx/src/__tests__/expression-parser.test.ts packages/jsx/src/__tests__/serialize-parsed-expr.test.ts`
- `env GOCACHE=/private/tmp/barefootjs-go-build-cache bun test packages/adapter-go-template/src/__tests__`

PR #2055, opened 2026-06-30
feat(jsx): fold searchParams object memos in the analyzer; value-only fold mode (#2040)

kfly8:
**Stack PR-D of the #2040 memo follow-up.** Base: **`claude/2040-C-searchparams-object-fold` (PR #2054)** — review/merge PR-C first; this diff is only PR-D's commit on top.

## What

Moves the searchParams object-memo fold into the **analyzer** (the compiler normalizes once), and adds the **value-only fold mode** that memo SSR baking needs. Reduces `parsedBlock` consumers 3 → 2.

## Honest scope note

I set out to fully remove `parsedBlock` (the statement AST) from the IR — the "Expr but the IR still carries statements" concern. **It can't be fully removed**, and the probe shows why: the guard-const memo
```
{ const k=sel(); const n=ext(); if(!k) return ALL; return ALL.filter(t=>t.length===n).concat(ALL.slice(n)) }
```
folds *all-or-nothing*. The SSR value only needs the `consequent` (`ALL`, guard falsy), but the **alternate** uses `n` inside a per-element callback — value-unsafe to duplicate — so `foldBlockToExpr` (even `valueOnly`) refuses the whole block. The guard-const fallback therefore genuinely needs the tolerant `parsedBlock` for blocks whose *alternate* (irrelevant to the SSR value) can't fold. So `parsedBlock` stays in the IR with two justified consumers (guard-const fallback + template-literal memo).

## How

- **`foldBlockToExpr({ valueOnly })`** — a memo's `parsed` is consumed for its VALUE only (effects run on the client via the verbatim runtime code), so a possibly-impure `const` used on some paths but not others is safe to inline: the value on every path is unchanged; only an effect on an unused-value path is dropped. Duplication of a non-deterministic call stays refused.
- The analyzer's memo fold uses `valueOnly` and adds the `searchParams()` env-signal local(s) to the purity oracle (idempotent request-query read), so the object memo folds to an `object-literal` `parsed` at analysis time.
- `computeObjectMemoInitialValue` reads that folded `parsed` directly; the per-adapter re-fold (PR-C) and its `parsedBlock` dependency are removed.
- `searchParamsLocalNames` accepts `{ imports }`, shared by the analyzer and the adapters (one definition of the env-signal import recognition).

## Testing

- **Go adapter unit: 563 pass / 0 fail.**
- **Conformance (Go + Perl): 780 pass, 9 fail** — the 9 are the pre-existing carousel failures on `main`.
- **jsx: 2165 pass**, only the 4 pre-existing checker-alias failures.

Refs: #2040, #2018, #2015.

---

### DIFF
--- a/.changeset/memo-fold-value-only.md
+++ b/.changeset/memo-fold-value-only.md
@@ -1,0 +1,12 @@
+---
+"@barefootjs/jsx": minor
+"@barefootjs/go-template": patch
+---
+
+Fold the searchParams object memo in the analyzer (value-only mode); the Go adapter reads the folded `parsed` directly (#2040, PR-D of the memo follow-up stack).
+
+`foldBlockToExpr` gains a `valueOnly` option: for a memo's SSR initial value the `parsed` tree is consumed for its VALUE only — effects run on the client via the verbatim runtime code — so a possibly-impure `const` used on some paths but not others is safe to inline (the value on every path is unchanged; only an effect on a path that doesn't use the value is dropped). Duplication of a non-deterministic call (used more than once on a path) is still refused.
+
+The analyzer's memo fold now (a) uses `valueOnly` and (b) adds the `searchParams()` env-signal local(s) to the purity oracle — an idempotent request-query read — so a `const sp = searchParams()` object memo folds to an `object-literal` `parsed` at analysis time. The Go adapter's `computeObjectMemoInitialValue` then reads that folded `parsed` directly and the per-adapter re-fold is removed; `searchParamsLocalNames` is reusable from `{ imports }` (one definition, shared by the analyzer and the adapters).
+
+This drops `parsedBlock` from the object-memo path. Two justified consumers remain (the template-literal block memo and the guard-const fallback for blocks whose unfoldable *alternate* branch — irrelevant to the SSR value — prevents a whole-block fold), so `parsedBlock` stays in the IR for now. Render parity verified: Go adapter unit + conformance (Go + Perl) green, jsx suite carries only the pre-existing checker-alias failures.
--- a/packages/adapter-go-template/src/adapter/memo/memo-value.ts
+++ b/packages/adapter-go-template/src/adapter/memo/memo-value.ts
@@ -10,7 +10,6 @@
  *     delegating value lowering to `lowerCtorExpr`.
  */
 
-import { foldBlockToExpr } from '@barefootjs/jsx'
 import type { ParsedExpr, ParsedStatement, TypeInfo } from '@barefootjs/jsx'
 
 import type { GoEmitContext } from '../emit-context.ts'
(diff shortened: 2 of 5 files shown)


kfly8 (commented):
No blocking findings.

Priority: P0/P1/P2 issues: none found.

I reviewed the new analyzer-level `valueOnly` fold path, the shared `searchParamsLocalNames({ imports })` refactor, and the removal of the Go adapter's per-adapter re-fold for searchParams object memos. The important boundary still looks intact: `valueOnly` only relaxes dropped effects, while duplicated possibly-impure initializers remain refused; incomplete tolerant parses still leave `parsed` unset; and the remaining `parsedBlock` consumers continue to cover the guard-const and template-literal fallback paths.

Validation run:
- `bun test packages/jsx/src/__tests__/analyzer.test.ts packages/jsx/src/__tests__/expression-parser.test.ts`
- `bun test packages/adapter-go-template/src/__tests__/derived-state-memo.test.ts`
- `env GOCACHE=/private/tmp/barefootjs-go-build-cache bun test packages/adapter-go-template/src/__tests__`

kfly8:
Closing — not merging this.

A design discussion after this PR was opened converged on a policy that makes this change the wrong direction. In short:

- This PR pulls **`searchParams`-by-name recognition** into the analyzer's fold purity oracle (`searchParamsLocalNames({ imports })` added to `pureCallNames`). That's a *specific API name* special-cased in the compiler core — exactly the kind of thing the #2040 work has been removing, not adding.
- Its stated payoff (fold searchParams object memos into `parsed`, toward removing `parsedBlock`) doesn't actually land the cleanup: `parsedBlock` still can't be removed (the guard-const impure-tail fallback genuinely needs it, since `foldBlockToExpr` is all-or-nothing and an unfoldable *alternate* blocks the whole fold even under `valueOnly`).
- Under the agreed policy, SSR value embedding is **best-effort** and "how hard the compiler tries to embed" is an implementation detail, not a correctness boundary. The searchParams object memo already embeds fine via the **adapter-local** path on `main` (#2054's `computeObjectMemoInitialValue` re-fold with `ctx.state.searchParamsLocals`), which keeps searchParams knowledge in the Go adapter (its natural home) instead of the compiler core.

The right long-term shape (captured in a new RFC, see #ISSUE_PLACEHOLDER) is to make `searchParams` a **structurally-recognized signal** (`const [searchParams, setSearchParams] = createSearchParams()`) so the analyzer classifies it like any `createSignal` — no name special-casing anywhere — and to move `queryHref` out of core into the router layer. That removes the API-name coupling at the root rather than adding more of it here.

`main` stays at the #2052 / #2053 / #2054 state, which already retired the three statement-walk recognizers (filter / guard-const / searchParams-object) via the general fold. Nothing here is lost by closing.

---

kfly8:
RFC filed: #2057 (the `#ISSUE_PLACEHOLDER` above). Closing this PR in favor of that direction.

---

Issue #2057, opened 2026-07-01
RFC: keep API-name specialization out of the compiler core — env signals as structural signals, `queryHref` → router layer


kfly8:
## Background

The #2040 follow-up (retiring the per-idiom block-memo recognizers via the general `foldBlockToExpr`) surfaced a broader design question about **where API-specific knowledge lives**. Two runtime APIs — `searchParams` (env signal) and `queryHref` (query-URL builder) — are recognized *by name* in the compiler core:

- `searchParams` is allow-listed in `CLIENT_EXPORTS` and lowered specially (env-signal method calls, and — in the now-closed #2055 — added to the analyzer's fold purity oracle by name).
- `queryHref` is allow-listed and lowered by name to the adapter query helper (`queryHrefLocalNames` → `bf_query`).

This is exactly the kind of ad-hoc, name-based specialization the #2040 work has been removing from the block-body path. This RFC proposes removing it at the root so the **compiler core carries only general mechanisms**, and API sugar lives where it semantically belongs.

## Guiding principle

**The compiler core provides *mechanisms*; *conveniences* build on top.** Concretely, the core should carry no specific runtime-API names. It offers:

- structural recognition of reactivity (`createSignal` / `createMemo` shapes),
- the fold = "purely-functional value expressibility" surface (#2040),
- lowering of *standard platform* primitives (the existing Array/String method catalogue, etc.),

…and everything API-specific (routing, query building) is expressed in terms of those, in the layer it belongs to.

This also inherits the SSR value-embedding policy just documented in the spec (#2056): SSR reproduction is best-effort (embed if you can, else zero value + hydrate), loud only where there is no safe fallback.

### 1. `searchParams` → a structurally-recognized signal

Replace the bare global env-signal getter (`searchParams()` returning a mutable reader with a `.set()` API) with a `createSignal`-shaped factory:

```ts
const [searchParams, setSearchParams] = createSearchParams()
```

- **Read** (`searchParams()`): a reactive getter — idempotent within a render, so pure for the fold. Recognized *structurally* (same path as `createSignal`), so the analyzer's reactive-getter set includes it **without any `searchParams`-name allow-list**. This is the clean fix for the #2055 smell.
- **Write** (`setSearchParams(...)`): the single imperative navigation path — a signal-setter shape. Removes the overlapping mutable `.set()` on the reader (the DX-confusing third write path).
[...]

PR #2061, opened 2026-07-01
feat(#2057)!: move queryHref to the router layer (part 3)

kfly8:
Part 3 of the #2057 stack. Moves `queryHref` out of the compiler-core client package into `@barefootjs/router`, so the compiler core carries **no specific runtime-API names**. With `searchParams` already structural (part 1) and the lowering registry in place (part 2), this completes the RFC's "core = mechanisms, router = routing sugar" split.

## Runtime move (breaking)

`queryHref` (+ its `QueryParams` / `QueryParamValue` types) moves `@barefootjs/client` → `@barefootjs/router`. Behavior is unchanged — still a pure function that also runs during SSR.

```tsx
// before
import { queryHref } from '@barefootjs/client'
// after
import { queryHref } from '@barefootjs/router'
```

## Build recognition — explicit, declarative `plugins`

Recognition of `queryHref(base, { … })` (which lowers it to each adapter's query helper — `bf_query` / `bf->query` / `$bf.query`) is now owned by the router layer and declared in the build config:

```ts
// barefoot.config.ts
import { createConfig } from '@barefootjs/go-template/build'
import { queryHrefPlugin } from '@barefootjs/router/plugins'

export default createConfig({
  components: ['...'],
  plugins: [queryHrefPlugin],
})
```

- **`plugins` config field** (familiar, à la Vite/Rollup) instead of a magic side-effect import — you can see what's registered.
- **`queryHrefPlugin`** is named for the API it lowers but distinct from the runtime `queryHref`, so the two imports can't be confused (and passing the wrong one is a TS type error).
- Each adapter's `createConfig` registers the plugins into **its own bundled `@barefootjs/jsx` instance** — the same one the adapter reads at emit time. (Registering from the CLI's separate instance wouldn't be seen — a module-identity subtlety.)
- Without a registered plugin, a `queryHref(...)` call hits the support gate (BF101): the core no longer knows the name.

## Browser resolution (fixes the e2e regression)

An island's compiled client JS emits `import { queryHref } from '@barefootjs/router'`. When `@barefootjs/router` is installed, the CLI now:
- copies its standalone bundle to **`router.js`** next to `barefoot.js` (client external → `barefoot.js`, `@barefootjs/shared` inlined — self-contained), and
- **rewrites the island's `@barefootjs/router` import specifier to the served `./router.js`** (TS-AST, exactly like `@barefootjs/client` → `./barefoot.js`), plus an importmap entry.

So `queryHref` resolves in the browser identically to the client runtime. (Without this, islands failed to hydrate and query-only nav fell back to a full region swap — the e2e failure this PR resolves.)

## Core removals

- `'queryHref'` dropped from `CLIENT_EXPORTS`; `queryHrefLocalNames` / `QUERY_HREF_SOURCES` removed from `@barefootjs/jsx` (the router owns that resolution). `matchQueryHrefCall` / `queryHrefArgs` stay as generic guard-list helpers the router plugin builds on.
- `BuildOptions` gains a `plugins?: LoweringPlugin[]` field.
- Migration: shared blog sample + all 10 integration configs updated; `queryHref` dropped from the Hono client-shim re-export.

## Verification

- Full suites, **0 failures**: conformance (byte-exact) 789, jsx 2237, go 566, mojo 549, xslate 368, hono 452, client 372, router 56.
- End-to-end integration build (chi): 0 errors, `bf_query` in the SSR template, `router.js` served, and the island resolves `queryHref` from `./router.js`.
- `tsgo --noEmit` clean across jsx, client, router, and all adapters.
- Grep-verified: no bare `'queryHref'` / `'searchParams'` recognition remains in `@barefootjs/jsx`.

See the changeset for the full migration note.


kfly8:
Superseded by #2063. We decided `queryHref` should **stay in `@barefootjs/client`** (it's a pure URL-query helper with no structural routing coupling), and — since its runtime is core — its lowering is recognised **directly by each adapter** rather than through the plugin registry. That drops this PR's runtime move, `router.js` serving, and import rewriting entirely. The `LoweringPlugin` registry itself lives on in #2063 as the extension seam for non-core lowerings (with a sample plugin + tests). Closing.

---

Issue #2157, opened 2026-07-07
[P1] Ruby (barefoot_js 0.18.0): register_components_from_manifest renders every registry component's children EMPTY — derive_vars_from_defaults looks up symbol-keyed props with a String propName


kfly8:
_Migrated from kfly8-sandbox/sample-barefootjs-20260706#14 by @kfly8_

## Environment
- Published `barefoot_js` 0.18.0 (RubyGems) + `@barefootjs/erb` 0.18.0 (npm), Ruby 3.3.6, Sinatra 4.2

## Summary
Using the **documented** Ruby setup path — `bf.register_components_from_manifest(MANIFEST)` then `BACKEND.render_named('Counter', bf, seed)` — every registry child component renders with **empty children**. A `<Counter>` whose buttons say `+1 / -1 / Reset` renders three empty `<button>` elements. (Hydration still attaches, so the client eventually repaints, but SSR output is wrong, the page flashes empty, and it breaks no-JS / SEO / tests.)

## Root cause (one line)
`BarefootJS::Context.derive_vars_from_defaults` (lib/barefoot_js.rb):
```ruby
prop_name = d[:propName]                                   # => "children"  (a String)
extra[name] =
  if !prop_name.nil? && props.key?(prop_name) && !props[prop_name].nil?
    props[prop_name]
  else
    d[:value]                                              # => nil
  end
```
The compiled ERB passes child props with **symbol** keys — the generated `Counter.erb` emits:
```erb
<%= bf.render_child('button', { _bf_slot: 's4', children: __bf_children_s4 }) %>
```
so `props = { children: "+1" }` (symbol `:children`). But `d[:propName]` is the **String** `"children"` (a JSON string value, not a key). `props.key?("children")` is therefore always `false`, so every entry with a `propName` falls through to its static default (`value: null`) — the caller's real value is discarded.

## Minimal proof
```ruby
d = manifest[:"ui/button/index"][:ssrDefaults][:children]   # {propName:"children", value:nil}
props = { children: "+1" }
props.key?(d[:propName])         # => false   (String "children" vs symbol :children)
props.key?(d[:propName].to_sym)  # => true
BarefootJS::Context.derive_vars_from_defaults(
  manifest[:"ui/button/index"][:ssrDefaults], props)[:children]  # => nil   (should be "+1")
```

## Fix (one line)
Coerce `propName` to a symbol before the lookup:
```ruby
prop_name = d[:propName]&.to_sym
```
[...]

PR #2161, opened 2026-07-07
test: add render-stage cross-adapter conformance contract (#2158)

### LINKED ISSUE #2158: [proposal] adapter-tests: add a RENDER-stage cross-adapter conformance suite (compile-clean ≠ renders-correctly)
_Migrated from kfly8-sandbox/sample-barefootjs-20260706#15 by @kfly8_

## Motivation
`packages/adapter-tests` validates the **compile** stage (JSX → marked template + IR) and the compatibility matrix reports "62 × N compile clean". But every adapter bug found across the 0.17–0.18 review was at the **render** stage — only visible when the real backend runtime executes the template:

| Bug | Adapter | compile-clean matrix | a render check |
|---|---|---|---|
| registry children render EMPTY (symbol/String propName) | Ruby ERB | ✅ | ✗ children non-empty |
| loop rendering a child → iterates `.Checkboxs` | Go | ✅ | ✗ no-error-marker |
| `<Input placeholder=...>` → non-existent struct fields | Go | ✅ | ✗ won't build/render |
| multi-component 500 (`No renderer for toast_provider`) | Perl | ✅ | ✗ no-error-marker |
| scaffold 500 / silent empty-200 | Perl, Go | ✅ | ✗ non-empty body |

Compile-clean is necessary but not sufficient. The matrix's own "496/496" claim is measuring the wrong stage.

## Proposal
Add a render-stage conformance test to `packages/adapter-tests` that boots each integration server (reuse the existing per-adapter harness that `scaffold.contract.ts` / `dev-reload.contract.ts` already rely on), fetches a shared fixture, and asserts one golden contract on the produced HTML — same fixture, same assertions, every adapter.

### Contract v1 (stock Counter: 3 `<Button>` children, signal `count`, memo `doubled`)
1. `counter-ssr-initial` — `count` SSRs its initial value, not blank/NaN.
2. `child-slot-children-nonempty` — `<Button>` slot children render (`+1`/`-1`/`Reset`). **Ruby 0.18.0 fails this.**
3. `hydration-scripts-present` — client bundle + component script registered once.
4. `no-render-error-marker` — no `Template error` / `No renderer registered` / `can't evaluate field` / `Server Error` in the HTML.
5. `nonempty-body` — guards the silent empty-200 failure mode.

Text comparisons **decode HTML entities first** so a legitimate encoding difference (Go `html/template` emits `+` as `&#43;`; others emit it literally) is treated as equivalent rendered text — the reference harness surfaced exactly this divergence on its first run and it should be normalized, not flagged.

### Fixture set should grow to cover the other found-bug shapes
- loop body that renders a child component (catches #11)
- `<Input>` with HTML attributes `placeholder` / `value` (catches #12)
- a multi-component module page: Toast / Dialog / Tabs (catches #13)
- a scaffold-boot smoke fixture: `create → build → GET / is 200 & non-empty` per adapter, from the **packed tarball** (catches #1/#2/#7 and the "works in monorepo, broken as shipped" class)

### Result on the published 0.18.0 stack
The reference harness reports **PASS — 0 failing checks across 5 reachable adapters** (go/chi, perl/mojo, ruby/erb, python/jinja, rust/minijinja) — with the one-line Ruby `derive_vars_from_defaults` fix applied. Without that fix, `child-slot-children-nonempty` fails on Ruby, demonstrating the suite catches the exact class the compile matrix misses.

### Reference harness
Full source is in `conformance/render-contract.mjs` (+ README) on the eval branch — a self-contained `node render-contract.mjs` that targets running adapter servers and exits non-zero on any failing check. It's written to drop into `packages/adapter-tests` as `assertRenderContract(html)` once wired to the in-tree server-boot harness.

kfly8:
Closes #2158. Was stacked on #2159 (the #2157 Ruby fix, now merged); retargeted to `main`.

## What

Every adapter bug found across the 0.17–0.18 review was invisible at the compile layer — the compatibility matrix read 496/496 clean while the render stage was broken. This PR adds the render-stage counterpart to the compile conformance corpus, wired through the repo's existing conventions (reworked per review — no bespoke fixture or per-adapter test files).

### `fixtures/counter-buttons.ts` — a regular corpus fixture

The stock Counter (signal + memo, same shape as `fixtures/counter.ts`) composed with a children-forwarding `<Button>` child — the exact shape whose children rendered EMPTY on Ruby before #2157. Defined with `createFixture`, registered in `fixtures/index.ts`, `expectedHtml` generated by `scripts/generate-expected-html.ts`. As a corpus member it also joins cross-adapter HTML conformance, marker conformance, and the SSR-hydration contract like every other fixture — no skip-list entries needed anywhere.

### `src/render.contract.ts` — Contract v1

`assertRenderContract(html, opts?)` with five named checks, each failure message prefixed with its check id (mirroring `scaffold.contract.ts`'s step-naming):

1. `counter-ssr-initial` — the initial signal value SSRs as `0`, not blank/NaN
2. `child-slot-children-nonempty` — `+1` / `-1` / `Reset` render; no empty `<button>` (the #2157 regression pin)
3. `hydration-markers-present` — exactly one root `bf-s="test"` (corpus default scope, parameterizable) + ≥3 child scopes
4. `no-render-error-marker` — no `Template error` / `No renderer registered` / `can't evaluate field` / `Server Error` / Python traceback
5. `nonempty-body` — guards the silent empty-200 failure class

Text comparisons decode HTML entities first (`decodeHtmlEntities`, exported), so Go `html/template` emitting `+` as `&#43;` is treated as equivalent rendered text, not flagged. Malformed numeric references (out-of-range codepoints, lone surrogates) pass through undecoded instead of throwing — the function runs on arbitrary backend HTML including error pages (per review, with unit coverage).

### Execution — through the single mandatory entry point

The contract runs as a new suite inside `runAdapterConformanceTests` (with a typed `skipRenderContract` opt-out and the same `onRenderError` runtime-absent skip policy as the JSX suite). **Every adapter package's existing call site picks it up automatically** — Go, Mojolicious, Xslate, ERB, Jinja, minijinja (Rust), Twig, Blade — each in its own CI workflow where its real runtime is installed. A minimal test in adapter-tests covers the Hono reference leg (Hono has no runner call site) plus the entity-decoder units.

### ERB harness realignment (the piece that makes the suite bite)

`packages/adapter-erb/src/test-render.ts` used to seed child vars with a harness-local `defaults.merge(child_props)` shortcut over simplified `ssrDefaults` — sidestepping `derive_vars_from_defaults` entirely, which is exactly why #2157 was invisible in-tree. The generated child renderers now run the production sequence (`BarefootJS::Context.derive_vars_from_defaults` + `child_props.merge(extra)`, the same calls `register_components_from_manifest` makes in the published gem) against verbatim `ssrDefaults` entries. Only `@barefootjs/erb` gets a changeset — the other adapter packages no longer change at all.

## 🎯 First real catch, on the maiden CI run

The suite's very first CI run caught a live render-stage divergence: the fixture's original `'btn ' + className` **compiled clean on all 9 adapters but fataled at render time on Twig and Blade** (`PHP Fatal error: Unsupported operand types: string + string`) — JS's polymorphic `+` lowers to PHP's numeric-only `+`. Filed as #2163; the fixture uses a template literal (`` `btn ${className}` ``), which every adapter lowers through its dedicated concat path.

## Verification

- **Stack proof**: with the #2157 fix reverted, the runner's ERB render-contract test fails on `[child-slot-children-nonempty]` — and now the corpus HTML conformance for `counter-buttons` (plus 17 other child-component fixtures) fails too, because the realigned harness exercises the production path everywhere. With the fix: ERB suite 394 tests / 0 fail.
- Twig & Blade contract legs verified against **real** composer-installed PHP runtimes locally; Jinja, minijinja (cargo-built), go-template suites green.
- adapter-tests suite green (the 3 `divergences.test.ts` failures in the local sandbox reproduce without this change — local composer-vendor layout artifact, green in CI).

## Out of scope (follow-ups from the issue)

- Fixture growth: loop-body-renders-a-child, `<Input>` HTML attrs, multi-component module page (Toast/Dialog/Tabs), scaffold-boot smoke from the packed tarball.
- A full integration-server harness (HTTP round-trip, script-registration/full-page checks).
- #2163 — the Twig/Blade `+`-on-strings lowering gap this suite surfaced.

https://claude.ai/code/session_01RuFiQiCJGSp2inWdjrXPz2


kfly8:
Heads-up: reworking this now to align with the existing conformance conventions as requested — moving the fixture into `fixtures/counter-buttons.ts` (registered in the corpus, generated `expectedHtml`) and running the contract through `runAdapterConformanceTests` so every adapter picks it up automatically (the 5 hand-placed per-adapter tests and the bespoke cross-adapter test go away). Also folding in Copilot's two comments (`decodeHtmlEntities` invalid-code-point hardening, Go version wording). Push coming shortly — hold off merging until then. 🙏

---

kfly8:
Alignment rework pushed (f9eec2e) — ready for review again.

- Fixture is now a regular corpus member: `fixtures/counter-buttons.ts` (`createFixture`, registered in the index, generated `expectedHtml`) — it also joins HTML/marker/hydration conformance automatically, with zero skip-list entries.
- The contract executes through `runAdapterConformanceTests` as a new suite (typed `skipRenderContract` opt-out, same `onRenderError` skip policy as the JSX suite) — the 5 per-adapter test files and the bespoke cross-adapter test are gone; every adapter's existing call site picks it up with no wiring. A minimal adapter-tests test keeps the Hono reference leg.
- Both Copilot comments addressed: `decodeHtmlEntities` now validates numeric references (out-of-range / lone-surrogate → pass through undecoded, with unit coverage), and the stale Go-version doc text went away with the deleted file.
- Changeset shrank to `@barefootjs/erb` only — no other adapter package changes anymore.

Re-verified locally: ERB 394/0 (stack proof still bites — reverting the #2157 fix fails the runner's contract on `child-slot-children-nonempty`), Twig/Blade against real composer-installed runtimes, Jinja/minijinja/go-template suites green.

---

PR #2187, opened 2026-07-09
Support JSX element as a non-children prop (named slot) across all 8 template adapters

kfly8:
## Summary

Fixes the `jsx-element-prop` render divergence (#2168 sweep) on all 8 template adapters: a JSX element passed as a non-`children` prop (`<Card header={<strong>Title</strong>}>` — the slot / render-prop-lite pattern) previously rendered an empty slot or dropped the value silently.

- Every adapter already has a mechanism to forward the reserved `children` slot from a parent template into a child render (a captured buffer slice, a `{% set %}` block, a Kolon macro, a Go struct field, ...). This PR reuses that exact same mechanism for a JSX-valued prop under any OTHER name, keyed by the prop's own name instead of `children` — no new shared capture path invented.
- **Go**: bakes the value the same way real children are baked (`extractTextChildren` / `extractHtmlChildren`, falling back to `extractScopedHtmlChildren` when the root needs the parent's runtime scope id) and emits it as its own struct field.
- **Jinja / Twig / Rust (minijinja)**: a `{% set captureName %}...{% endset %}` block per named slot, passed as a dict/hash entry.
- **Text::Xslate**: a Kolon `macro NAME -> () { ... }` per named slot, called immediately in the hash literal.
- **Blade**: a PHP output-buffering capture (`ob_start()` / `ob_get_clean()`), wrapped in `$bf->backend->mark_raw(...)`.
- **Mojolicious**: a `begin %>...<% end` capture passed into `render_child`'s named-arg list. The shared `BarefootJS.pm` runtime's `render_child` now materializes *every* prop value (previously only the reserved `children` key) — a no-op for any value that isn't a captured CODE ref, so this generalizes safely to Text::Xslate too.
- **ERB**: the same output-buffer-slice capture already used for `children`, but ERB's `<%=` (unlike every other adapter's template tag) has no built-in "safe string" wrapper to bypass escaping on read-back — so the runtime gains one: a new `BarefootJS::SafeString` marker class, returned by `Backend::Erb#mark_raw` (previously an identity no-op) and recognized by `Context#h` to skip re-escaping already-forwarded HTML.

`jsx-element-prop` graduates from a render divergence to a passing render on all 8 template adapters; `ui/compat.lock.json` and the divergence declarations are updated accordingly.

Stacked on #2186 (math-methods).

## Test plan

- [x] `bun run build` (full monorepo) + `GOTOOLCHAIN=go1.25.6 bun run compat:lock` — clean diff, only `jsx-element-prop` entries removed
- [x] Go adapter suite (real `go test` via `GOTOOLCHAIN=go1.25.6`): 725 pass, 0 fail
- [x] Jinja adapter suite (real Jinja2): 516 pass, 0 fail
- [x] Text::Xslate adapter suite (real Xslate): 517 pass, 0 fail
- [x] Twig adapter suite (real Twig): 516 pass, 0 fail
- [x] Blade adapter suite (real Blade): 516 pass, 0 fail
- [x] Rust (minijinja) adapter suite (real `bf-render` binary): 516 pass, 0 fail
- [x] ERB adapter suite (real stdlib ERB): 490 pass, 2 skip, 0 fail, plus full Ruby unit-test suite (`test/*.rb`)
- [x] Mojolicious adapter suite (real Mojolicious): 687 pass, 15 skip, 0 fail, plus full Perl unit-test suite (`t/*.t`)
- [x] `bun run lint`

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZXzQWvSKLUUrTAZ3vph1j)_

### DIFF
--- a/.changeset/jsx-element-prop-named-slots.md
+++ b/.changeset/jsx-element-prop-named-slots.md
@@ -1,0 +1,22 @@
+---
+"@barefootjs/jinja": patch
+"@barefootjs/mojolicious": patch
+"@barefootjs/xslate": patch
+"@barefootjs/twig": patch
+"@barefootjs/blade": patch
+"@barefootjs/rust": patch
+"@barefootjs/erb": patch
+"@barefootjs/go-template": patch
+"@barefootjs/perl": patch
+---
+
+Support a JSX element passed as a non-`children` prop (`<Card header={<strong>Title</strong>}>`, the slot / render-prop-lite pattern) on all 8 template adapters. Every adapter already had a mechanism to forward the reserved `children` slot from a parent template into a child render (a captured buffer slice, a `{% set %}` block, a Kolon macro, a Go struct field, ...); named JSX-valued props reuse that exact same mechanism, keyed by the prop's own name instead of `children`, rather than inventing a new shared capture path.
+
+- **Go**: bakes the value the same way real children are baked (`extractTextChildren` / `extractHtmlChildren`, falling back to `extractScopedHtmlChildren` when the root needs the parent's runtime scope id) and emits it as its own struct field.
+- **Jinja / Twig / Rust (minijinja)**: a `{% set captureName %}...{% endset %}` block per named slot, passed as a dict/hash entry.
+- **Text::Xslate**: a Kolon `macro NAME -> () { ... }` per named slot, called immediately in the hash literal.
+- **Blade**: a PHP output-buffering capture (`ob_start()` / `ob_get_clean()`), wrapped in `$bf->backend->mark_raw(...)` so the child's `{{ }}` doesn't re-escape it.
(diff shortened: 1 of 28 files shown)


kfly8:
Fixed in bc92620d: the capture identifier (variable/macro name) on all 7 flagged adapters (Jinja, Twig, minijinja, Text::Xslate, Blade, Mojolicious, ERB) is now purely counter-based, never derived from the raw prop name or `comp.slotId`. The hash/dict KEY passed to the child still carries the real prop name through each adapter's existing key-quoting helper (also fixed a related gap in the Xslate port, which built that key from the bare prop name without `kolonHashKey`). Added a unit test per adapter pinning a hyphenated prop name (`data-slot`) against this regression. Go was not affected — its struct-field emission already skips a hyphenated prop name when there's no rest bag to route it into.

---

PR #2188, opened 2026-07-09
Fix optional-chaining-prop divergence on Go and ERB

kfly8:
## Summary

Fixes the `optional-chaining-prop` render divergence (#2168 sweep) on 2 of 8 template adapters: `user?.name ?? 'anonymous'` on an optional object-shaped prop previously exited/raised at render time on Go and Ruby ERB.

- `?.`'s "guard the nil receiver" semantics were discarded entirely at parse time — `ts.isPropertyAccessExpression`'s `questionDotToken` was never read, so `user?.name` and `user.name` produced the identical `ParsedExpr` shape. This happened to work on 6 adapters by accident of their host language/helper semantics (Jinja/Twig/minijinja/Xslate's generic `[]`/`.` accessor already swallows a nil/undef receiver; Blade already routes every member access through the null-safe `data_get()` helper), but Go's literal `.Field` dot-chain panics on a nil interface/pointer, and Ruby's `obj[:key]` raises `NoMethodError` on nil.
- Added an `optional: boolean` field to the shared `member` `ParsedExpr` variant (set from the `?.` token, threaded through every rewrite/copy site so it survives destructure and callback-body rewrites) and to the `ParsedExprEmitter.member()` interface — a required field, so every adapter's implementation had to be visited.
- **Go**: an `optional` access routes through the runtime's existing nil-safe reflection getter (`bf_get`/`getFieldValue`, `bf.go`, already used for map-rooted context chains) instead of a literal `.Field` dot-chain.
- **ERB**: an `optional` access emits Ruby's native safe-navigation form (`obj&.[](:key)`) instead of plain `obj[:key]`.
- The other 6 adapters (Jinja, Twig, minijinja, Xslate, Blade, Mojolicious) accept-and-ignore the new parameter — no behavior change, since their existing lowering is already null-safe.
- Scoped to the single `?.`-written hop, matching the current fixture — does not (yet) propagate JS's whole-chain short-circuit through a following plain `.` hop (`a?.b.c`); noted as a known limitation on the `ParsedExpr` docstring.
- Design reviewed with Opus before implementation (alternative considered and rejected: making the defensive form the *unconditional* default for all Go/ERB member access — rejected because the runtime's nil-safe getter also does fuzzy case-insensitive key resolution, so routing every access through it would mask genuine wrong-field-name bugs and change key-resolution order for the non-optional case).

`optional-chaining-prop` graduates from a render divergence to a passing render on both adapters; `ui/compat.lock.json` and the divergence declarations are updated accordingly.

Stacked on #2187 (jsx-element-prop).

## Test plan

- [x] `bun run build` (full monorepo, to refresh adapter `dist/` bundles) + `GOTOOLCHAIN=go1.25.6 bun run compat:lock` — clean diff, only `optional-chaining-prop` entries removed
- [x] Go adapter suite (real `go run` via `GOTOOLCHAIN=go1.25.6`): 726 pass, 0 fail (re-confirmed after an earlier transient run showed unrelated flaky timeouts under load — reproduced clean twice)
- [x] ERB adapter suite (real stdlib ERB): 492 pass, 2 skip, 0 fail
- [x] Jinja / Twig / Rust (minijinja) / Text::Xslate / Blade / Mojolicious adapter suites (real backends): all green, unchanged pass counts
- [x] Shared `@barefootjs/jsx` package suite: 2336 pass, 25 skip, 2 todo, 0 fail (two golden-shape assertions updated for the new `optional` field)
- [x] `bun run lint`

---
_Generated by [Claude Code](https://claude.ai/code/session_017ZXzQWvSKLUUrTAZ3vph1j)_

### DIFF
--- a/.changeset/optional-chaining-prop.md
+++ b/.changeset/optional-chaining-prop.md
@@ -18,6 +18,6 @@ Go and ERB act on the flag:
 - **Go**: an `optional` access routes through the runtime's existing nil-safe reflection getter (`bf_get`/`getFieldValue`, `bf.go`) instead of a literal `.Field` dot-chain, which panics evaluating a field on a nil interface/pointer (`nil pointer evaluating interface {}.Name`).
 - **ERB**: an `optional` access emits Ruby's native safe-navigation form (`obj&.[](:key)`) instead of plain `obj[:key]`, which raises `NoMethodError` on a `nil` receiver.
 
-Both routes only guard the single hop actually written with `?.` — a following plain `.c` after an optional `a?.b` is not (yet) short-circuited, matching JS's whole-chain semantics; see the `member` variant's docstring.
+Both routes only guard the single hop actually written with `?.` — a following plain `.c` after an optional `a?.b` is not (yet) short-circuited, so this does not yet match JS's whole-chain short-circuit semantics; see the `member` variant's docstring.
 
 `optional-chaining-prop` graduates from a render divergence to a passing render on both adapters.
--- a/.changeset/optional-chaining-prop.md
+++ b/.changeset/optional-chaining-prop.md
@@ -1,0 +1,23 @@
+---
+"@barefootjs/jsx": patch
+"@barefootjs/jinja": patch
+"@barefootjs/mojolicious": patch
+"@barefootjs/xslate": patch
+"@barefootjs/twig": patch
+"@barefootjs/blade": patch
+"@barefootjs/rust": patch
+"@barefootjs/erb": patch
+"@barefootjs/go-template": patch
+---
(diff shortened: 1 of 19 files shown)


kfly8:
Both addressed in 8ccc1bc5:
- Fixed the self-contradictory changeset sentence — the single-hop scoping does NOT match JS's whole-chain short-circuit semantics.
- Added a focused `parseExpression` test pinning `optional: true` on `user?.name`, `obj?.['key']`, and `arr?.[0]` (plus `optional: false` on the non-optional `arr[0]` counterpart).

---

Issue #2205, opened 2026-07-11
Sibling-template registration for loop-body components isn't exercised by the conformance harness (BF103) — likely fixable, not by-design


kfly8:
Refs #2204 (parent tracker).

## Symptom

`static-array-children`, `todo-app`, and `todo-app-ssr` refuse to compile on all 8 non-Hono template adapters with a `BF103` diagnostic: a sibling-imported component used inside a `.map()` loop body lowers to a cross-template call (Go: `{{template "X" .}}`, and each other adapter's equivalent partial/include construct), which only resolves if the child template is registered on the same template instance at render time.

## Why #2069 calls this "by design, no fix planned" — and why that looks stale

The Go adapter's own check (`packages/adapter-go-template/src/adapter/go-template-adapter.ts:346-354`, `checkImportedLoopChildComponents`) is gated:

```ts
// The barefoot CLI compiles a source dir together onto one
// *template.Template, so it sets `siblingTemplatesRegistered` and the
// check would be noise; stand-alone `compileJSX` callers leave it
// unset and get the loud build-time error.
if (!options?.siblingTemplatesRegistered) {
  this.checkImportedLoopChildComponents(ir)
}
```

`packages/compat/src/engine.ts` confirms `bf build` (real usage, and the `bun run compat` 558-cell component matrix) always passes `{ siblingTemplatesRegistered: true }` — this diagnostic is genuinely silenced for every real project.

But the **fixture-corpus conformance runner** — `collectFixtureDiagnostics` in `packages/adapter-tests/src/jsx-runner.ts:405-426` — never sets this flag for either the parent or child compile calls. So every multi-file, loop-body-component fixture hits BF103 in the conformance suite regardless of whether it would actually work in a real build.

That's the same class of bug as #2202 ("Go adapter: nested-loop-triple-depth divergence" — actually a test-harness key-capitalization bug, not a real adapter limitation) fixed earlier in this same sweep: the **test harness**, not the compiler, is the actual gap.

## What's still unverified (the reason this isn't marked a duplicate of a simple flag flip)

[...]

Issue #2204, opened 2026-07-11
Render Conformance (fixture corpus): tracking issue for the 9 outstanding gaps


kfly8:
## What this tracks

The [Compatibility Matrix](https://barefootjs.dev/docs/advanced/compatibility-matrix) page's **"Render Conformance (fixture corpus)"** section currently reports 230/239 fixtures rendering to reference parity on every adapter, with 9 fixtures showing a gap on one or more of the 8 non-Hono template adapters (blade, erb, go-template, jinja, minijinja, mojolicious, twig, xslate). This issue is the parent tracker for those 9, following the split-out convention established by #2069 (each entry gets its own dedicated issue carrying `known-limitation`, cross-linked here).

Audited 2026-07-11 against `ui/compat.lock.json` (regenerated from latest `main`, zero diff — the deployed page is current).

## Sub-issues

| Fixture(s) | Adapters | Diagnostic | Issue |
|---|---|---|---|
| `array-map-function-reference` | all 8 | BF101 | (this catalog, see sub-issue) |
| `dangerous-inner-html` | all 8 | BF101 | (this catalog, see sub-issue) |
| `static-array-children`, `todo-app`, `todo-app-ssr` | all 8 | BF103 | (this catalog, see sub-issue) |
| *(string `+` on Twig/Blade)* | twig, blade | render-time fatal | #2163 |
| *(string `.length` host-native)* | go-template, erb, jinja, minijinja, mojolicious, xslate | render divergence | #2196 |

`filter-nested-callback-predicate` and `filter-nested-find-predicate` (also visible on the matrix) are **not** included here — both are already fully tracked under the closed #2038, which explicitly delivered the current BF101 behavior as its fix (converting a silent SSR lossy-lambda bug into a loud refusal); the residual "can't lower this shape at all" is the documented, intentional end state of that issue, not a new gap.

## Note on #2069's "by design, no fix planned" characterization

#2069 (closed) lists the sibling-template-registration BF103 (`static-array-children` / `todo-app` / `todo-app-ssr`) under "By design / cosmetic (documented, no fix planned)". Investigating this catalog turned up evidence that characterization is stale — see the dedicated sub-issue for the root-cause finding; it looks fixable, not permanent.

PR #2211, opened 2026-07-11
Fix #2163: exercise real JS `+` string concat in counter-buttons fixture

### LINKED ISSUE #2163: Twig/Blade: JS string `+` concatenation lowers to PHP numeric `+` — compiles clean, fatals at render time
Found by the first CI run of the render-stage conformance contract (#2158, PR #2161) — a textbook instance of the "compile-clean ≠ renders-correctly" class that suite exists to catch.

## Symptom

A children-forwarding Button using JS string concatenation in an attribute:

```tsx
export function Button({ children, className = '', onClick }) {
  return <button className={'btn ' + className} onClick={onClick}>{children}</button>
}
```

compiles clean on **every** adapter, but on Twig and Blade the rendered template dies with:

```
PHP Fatal error: Uncaught TypeError: Unsupported operand types: string + string
  in .../twig/src/Environment.php(420) : eval()'d code:48
```

## Root cause

The PHP-family emitters map JS binary `+` straight to the target's `+`:

- `packages/adapter-twig/src/adapter/expr/emitters.ts:199` — `'+': '+'`
- `packages/adapter-blade/src/adapter/expr/emitters.ts:220` — `'+': '+'`

In JS, `+` is polymorphic (string concat when either operand is a string). Twig/Blade `+` is numeric-only; concatenation is `~` (Twig) / `.` (Blade/PHP). Jinja escapes this because Python's `+` also concatenates strings; Go/Perl/Ruby paths have their own handling.

Template **literals** are unaffected — both adapters already have a dedicated concat lowering (Twig `~` / Blade `.` via `bf.string(...)`, see `adapter-twig/src/adapter/expr/emitters.ts:560`) — so `` `btn ${className}` `` renders correctly. That's the workaround, and what the conformance fixture now uses (see PR #2161).

## Possible directions

1. Detect statically-string operands (string literal on either side, template-literal operand) and emit the concat operator instead of `+` — covers the overwhelmingly common `'prefix ' + expr` shape.
2. A polymorphic runtime helper (`bf.add(a, b)`) mirroring JS semantics for the residual cases where operand types aren't statically known — consistent with how other JS-semantic gaps route through `bf.*` helpers.
3. At minimum, refuse with a BF diagnostic instead of emitting a render-time fatal, so the failure moves back to the compile stage.

## Repro

Swap the template literal back to `'btn ' + className` in `renderContractFixture` (`packages/adapter-tests/src/render.contract.ts`) and run:

```
composer install --working-dir=packages/adapter-php
composer install --working-dir=packages/adapter-twig/php
cd packages/adapter-twig && bun test src/__tests__/render-contract.test.ts
```

(`adapter-blade` equivalently.) The contract fails on `no-render-error-marker` / the render call itself; with the template literal it passes 18/18 assertions on both adapters.

kfly8:
Fixes #2163. Second in the #2204 sub-issue PR stack (stacked on #2210 / the #2205 fix).

## Finding

The compiler-side fix for #2163 (JS binary `+` with a string-typed operand lowering to Twig's `~` / Blade's `.` concat operator instead of the numeric `+` that fatals at PHP render time) **already landed on `main`** via PR #2176 (`isStringConcatBinary` / `isStringTypedOperand` in `packages/jsx/src/adapters/parsed-expr-emitter.ts`, consumed by both `packages/adapter-twig/src/adapter/expr/emitters.ts` and `packages/adapter-blade/src/adapter/expr/emitters.ts`).

However, the `counter-buttons` conformance fixture (`packages/adapter-tests/fixtures/counter-buttons.ts`) — the exact fixture the issue's own repro section names — still used the **template-literal workaround** (`` className={`btn ${className}`} ``) instead of the shape the issue actually reports (`className={'btn ' + className}`). So the fix, while present in the compiler, was never exercised end-to-end by this fixture.

## Change

Swap the workaround back to the real JS `+` shape the issue describes, per its own repro instructions ("Swap the template literal back to `'btn ' + className`... run the tests").

## Testing

- `bun test packages/adapter-tests/src/__tests__/render-contract.test.ts` — 6 pass.
- `bun test packages/adapter-twig/src/__tests__` (real PHP + Twig via `composer install`) — 518 pass, 0 fail.
- `bun test packages/adapter-blade/src/__tests__` (real PHP + Blade via `composer install`) — 518 pass, 0 fail.
- `bun test packages/{adapter-erb,adapter-jinja,adapter-rust}/src/__tests__ -t counter-buttons` — pass on real ruby/python/rust toolchains.
- go-template / mojolicious / xslate skip real rendering in this sandbox (missing `go`/`perl` toolchains) but compile clean.
- `bun run compat:lock` — no diff (fixture isn't part of the `ui/` component matrix).
- No changeset needed — `@barefootjs/adapter-tests` is a private, changeset-ignored package.

## Review plan

---

### DIFF
--- a/packages/adapter-tests/fixtures/counter-buttons.ts
+++ b/packages/adapter-tests/fixtures/counter-buttons.ts
@@ -45,15 +45,15 @@ export function Counter() {
 }
 `,
   components: {
-    // Class composition uses a template literal, NOT string `+`: every
-    // adapter has a dedicated template-literal concat lowering, whereas
-    // JS binary `+` currently lowers to the target language's numeric
-    // `+` — a PHP fatal on Twig/Blade at render time despite compiling
-    // clean (tracked separately, #2163). This fixture deliberately
-    // stays inside every adapter's supported envelope.
+    // Class composition uses JS binary `+` (not a template literal) —
+    // exercising #2163's fix: `isStringConcatBinary` detects the
+    // string-literal `'btn '` operand and lowers `+` to each target's
+    // string-concat operator (Twig `~`, Blade/PHP `.`) instead of its
+    // numeric `+`, which used to compile clean but fatal at PHP render
+    // time (`Unsupported operand types: string + string`).
     './button.tsx': `
 export function Button({ children, className = '', onClick }: { children?: unknown; className?: string; onClick?: () => void }) {
-  return <button className={\`btn \${className}\`} onClick={onClick}>{children}</button>
+  return <button className={'btn ' + className} onClick={onClick}>{children}</button>
 }
 `,
   },


kfly8:
Fable review: confirmed the diagnosis (compiler fix from #2176 correctly detects the `'btn '` string-literal operand and lowers `+` to `~`/`.` for this exact fixture — one string-typed side suffices, `className`'s type never needs resolving), and no regression risk (template-literal className lowering stays heavily exercised elsewhere).

One real, pre-existing, uniform-across-adapters gap flagged: `isStringTypedOperand` has no bare-identifier arm, so `a + b` where *both* operands are non-literal string values (no literal, no getter, no `props.x`) still falls through to numeric `+` and would fatal identically. This is out of scope for this fixture-only PR and predates it — filed as #2212 (referencing #2163's own "Possible directions" #2, the `bf.add`-style polymorphic helper).

Ready for Copilot review.

---

PR #2213, opened 2026-07-11
Fix #2196 (Level 1): Perl .length counts Unicode codepoints, not UTF-8 bytes

### LINKED ISSUE #2196: Evaluator/helper string `.length` is host-native (bytes on Perl, codepoints elsewhere) — diverges from JS UTF-16 code units on non-ASCII
Refs #2069 (known-limitations catalog). Surfaced by the generative divergence probe added in #2193 (follow-up to the `Number()` coercion fixes in #2192), which flags these inputs `known` so the probe stays green while tracking the divergence here.

## Limitation

The ParsedExpr evaluator subset supports `.length` on a string (spec/compiler.md evaluator node table: "`.length` on a string/array"). Each backend evaluator computes it with its host language's native string length, which disagrees with JS's **UTF-16 code-unit** count on non-ASCII input. Measured *through the evaluator* (`item.s.length`, and the same via a nested `.map(t => t.length)`):

| input | JS (UTF-16 units) | Go | Ruby | Python | PHP | Perl |
|---|---|---|---|---|---|---|
| `"café"` | 4 | 4 | 4 | 4 | 4 | **5** |
| `"日本"` | 2 | 2 | 2 | 2 | 2 | **6** |
| `"ﾊ"` | 1 | 1 | 1 | 1 | 1 | **3** |
| `"😀"` | **2** | 1 | 1 | 1 | 1 | 4 |
| `"a😀b"` | **4** | 3 | 3 | 3 | 3 | 6 |

Two distinct divergences:

1. **Perl counts UTF-8 bytes** — wrong for *any* non-ASCII (`"café"` → 5). Most severe: `.filter(x => x.name.length > N)` / `.map(x => x.tag.length)` over accented or CJK text produces a visibly wrong number on the **Mojolicious / Xslate** backends (both share `BarefootJS::Evaluator`).
2. **Go / Ruby / Python / PHP count Unicode codepoints** — correct for the BMP but wrong for astral characters (emoji, CJK-ext), where JS counts 2 UTF-16 code units (`"😀".length === 2`).

JS `String.prototype.length` is UTF-16 code units; none of the five backends reproduce it.

## Why it's currently untested

`eval-vectors.json` is JS-normative and forbids per-backend divergence, but the committed corpus keeps string cases ASCII, so this never fails. The `len` helper spec (spec/template-helpers.md:239) explicitly scopes strings to the "ASCII domain — length units differ across hosts otherwise," but the **evaluator** node table (spec/compiler.md:654) lists `.length` on a string with no such caveat — so the evaluator subset silently inherits an undocumented divergence its own "no divergence" contract claims not to have. The generative probe (#2193) flags these inputs `known` in `generate-probe.ts` (category `string-length`).

## Where it lives

The `.length` arm of each backend's member/`readProperty` evaluation, plus the `len` helper on the template path:

- Go — `packages/adapter-go-template/runtime/eval.go`
- Perl — `packages/adapter-perl/lib/BarefootJS/Evaluator.pm` (shared by Mojolicious + Xslate)
- Ruby — `packages/adapter-erb/lib/barefoot_js/evaluator.rb`
- Python — `packages/adapter-jinja/python/barefootjs/evaluator.py`
- PHP — `packages/adapter-php/src/Evaluator.php`
- Helper `len` / string-length lowering — spec/template-helpers.md "len" (the same host-native length underlies it).

## Planned fix (directions)

Two independently-shippable levels:

1. **Floor — Perl to codepoints.** Make Perl agree with the other four (operate on the decoded/UTF-8-flagged string so `length` counts characters, not bytes). Removes the worst case (all BMP text) even without full UTF-16 semantics.
2. **Full JS parity — UTF-16 code units on all backends.** Count code units: for each Unicode scalar add 1 if ≤ U+FFFF, else 2 (surrogate pair). Well-defined and testable per language:
   - Go: `len(utf16.Encode([]rune(s)))`, or sum per-rune (`r > 0xFFFF ? 2 : 1`).
   - Python: `len(s.encode('utf-16-le')) // 2`.
   - Ruby: `s.encode('UTF-16LE').bytesize / 2`.
   - PHP: sum per-codepoint (`mb_str_split` + `> 0xFFFF ? 2 : 1`).
   - Perl: sum per-codepoint over the decoded string.

   Then add `eval-cases.ts` vectors (`"café"`, `"日本"`, `"😀"`, `"a😀b"`) so every backend is pinned to JS, and drop the `known` flags for `string-length` in `generate-probe.ts`.

If full UTF-16 parity is judged not worth the cost (the spec deliberately keeps strings ASCII-domain), the minimum coherent outcome is **level 1 alone** (so the five backends at least agree with each other) plus stating the astral-plane divergence in spec/compiler.md's evaluator section — it currently lives only in the helper spec.

Scope note: `.length` is the flagship, but the same ASCII-domain caveat covers `.startsWith(pfx, position)`, `.padStart(len)`, `.slice`, `.at`, and index math on astral strings; a full fix would extend there. This issue tracks `.length` (evaluator + `len` helper) as the concrete, highest-impact instance.

## Workaround

Keep string operands ASCII in evaluator/template contexts; precompute a length upstream in JS; or `/* @client */` the expression so the browser's UTF-16 `.length` runs.

kfly8:
Refs #2196 — resolves Level 1 (Perl-to-codepoints). Level 2 (full JS UTF-16 code-unit parity on all five backends) remains open by design, per the issue's own two-level proposal, so I'm not closing #2196 with this PR — leaving that decision to a maintainer. Third in the #2204 sub-issue PR stack (stacked on #2211 / #2210).

## Problem

`BarefootJS::Evaluator`'s `.length` arm (shared by the Mojolicious and Xslate adapters) used Perl's builtin `length()`, which counts **UTF-8 bytes** unless the scalar already carries Perl's internal UTF8 flag. This diverges from JS's `.length` (UTF-16 code units) on **any** non-ASCII input — `"café".length` was `5` instead of `4`, `"日本".length` was `6` instead of `2`.

## Fix

- New `_codepoint_length()` in `Evaluator.pm`: counts Unicode codepoints, decoding defensively via `utf8::decode` (a no-op on already-decoded or invalid-UTF-8 input) rather than assuming the caller already did. This means the fix is correct regardless of how the string arrived — no changes needed to any production caller.
- Now agrees with JS for every BMP-only string, and with Go/Ruby/Python/PHP (which already counted codepoints).
- Astral characters (emoji, CJK extension ideographs) still diverge from JS's 2-code-unit count — documented, deliberately out of scope (Level 2).

## Test-harness fixes (needed to actually exercise the fix)

- `eval_vectors.t` and `probe.pl`'s `JSON::PP` decode calls were missing `->utf8`, so non-ASCII vector content would arrive as mojibake regardless of the `Evaluator.pm` fix — mirrored `helper_vectors.t`'s existing `->utf8` pattern.
- Added a `binmode ... :encoding(UTF-8)` TAP/STDOUT layer to both so wide-character test names/notes don't warn.

## New pinned vectors

- `eval-cases.ts`: café/日本/ﾊ `.length` cases — these flow through the **strict, JS-normative** `eval-vectors.json` corpus (all backends must match exactly), so only BMP-safe cases were added here. Astral cases were deliberately NOT added to this corpus (they'd fail on every backend, not just Perl — that's Level 2).
- `generate-probe.ts`: unflagged `known` for café/日本/ﾊ in the `STRS` corpus (Perl no longer diverges), kept `known` for 😀/a😀b (still Level 2).
- `spec/compiler.md`: added the astral-character caveat to the evaluator's `.length` documentation (previously only the `len` helper spec carried it).

## Testing

- `bun test packages/adapter-tests/src/__tests__/eval-vectors.test.ts` — 3 pass (freshness).
- `prove -I lib t/eval_vectors.t t/evaluator.t t/helper_vectors.t` (real Perl 5.38) — 403 tests, all pass.
- Generative divergence probe (`bun vectors/probe/generate-probe.ts` + real `perl vectors/probe/runners/probe.pl`) — 910 cases, 0 `NEW`, 0 `ERROR`; café/日本/ﾊ no longer appear in the `KNOWN` output (previously did), only 😀/a😀b remain.
- `bun test packages/adapter-mojolicious/src/__tests__` / `adapter-xslate/src/__tests__` — no regressions (real Perl/Mojolicious/Text::Xslate runtimes not installed in this sandbox, so real-render assertions skip, but codegen/unit tests are unaffected either way).
- Changeset added for `@barefootjs/perl`.

## Review plan

---


kfly8:
Fable review: no blocking findings. Verified `_codepoint_length`'s defensive-decode idiom directly on Perl 5.38 (flagged/unflagged UTF-8, invalid-UTF-8 fallback, no in-place mutation), confirmed café/日本/ﾊ are genuinely BMP-only (so safe in the strict eval-vectors.json corpus) while 😀 is correctly kept astral/`known`, and confirmed the `->utf8` JSON decode fixes were genuine bugs (would've masked string-*value* comparisons with mojibake even though `.length` itself is self-correcting). Also confirmed `eval_json`'s un-touched `->utf8`-less decode is correct as-is — its callers already hand it decoded character strings, so adding `->utf8` there would double-decode.

Ready for Copilot review.

---

Issue #2090, opened 2026-07-04
Function-reference sort comparators (arr.sort(myCmp)) are refused with BF021 — needs scope resolution


kfly8:
Refs #2069 (known-limitations catalog v2 — split-out entry).

## Limitation

A `.sort()` / `.toSorted()` comparator passed **by reference** is refused with **BF021** even when the referenced function is a perfectly on-catalogue comparator declared in the same file:

```tsx
const byPrice = (a: Item, b: Item) => a.price - b.price
// or: function byPrice(a: Item, b: Item) { return a.price - b.price }

{items().toSorted(byPrice).map((i) => <li>{i.name}</li>)}   // BF021 today
```

Only a syntactically inline arrow / function expression is accepted. The comparator *shape* catalogue (subtraction, zero-arg `localeCompare`, relational-ternary, `||`-chained multi-key) is not the problem — the same body inlined at the call site lowers fine on every adapter.

## Where the refusal lives

- `extractSortComparator` (`packages/jsx/src/jsx-to-ir.ts`) rejects any callback that is not `ts.isArrowFunction` / `ts.isFunctionExpression` — a `ts.Identifier` is refused before any resolution attempt (`'Sort comparator must be an arrow function or function expression'`).
- The analyzer already collects everything needed to resolve the name: `ctx.localConstants` (with `isModule` + shadowing-aware `findLocalConst` lookup, `parseConstInitializer` re-parse) and `ctx.localFunctions` for `function` declarations. `tryResolveTemplateSpanFromConst` is the live precedent for exactly this resolve-then-structurally-consume pattern.
- Pinned by `unsupported-expression.test.ts` (`describe('Unsupported Sort Comparator (BF021)')`) and the `unsupportedSort` tables in the Go / Mojolicious adapter tests.

## Planned fix

1. In `extractSortComparator`, when the callback is an identifier, resolve it through `findLocalConst` (const-declared arrow / function expression) or `ctx.analyzer.localFunctions` (function declaration), parse the initializer, and feed the resulting arrow through the **existing** `sortComparatorFromArrow` catalogue gate — no new comparator machinery, no `IRLoopSort` schema change.
2. Carry the resolved arrow's `raw` / params so the client-JS `.toSorted((a, b) => …)` rebuild stays correct.
3. Resolution failures (out-of-scope name, imported symbol, off-catalogue body) keep BF021 with a message that names the resolved (or unresolved) comparator.
[...]

PR #2214, opened 2026-07-11
Fix #2206: resolve named function-reference .map(cb) callbacks

### LINKED ISSUE #2206: Named function-reference `.map(cb)` callback refuses on all template adapters (BF101)
Refs #2204 (parent tracker).

## Symptom

```tsx
const format = (t: string) => '#' + t
function TagLine({ tags }: { tags: string[] }) {
  return <p>{tags.map(format).join(' ')}</p>
}
```

`tags.map(format)` — a named function reference rather than an inline arrow — refuses with `BF101` on all 8 non-Hono template adapters (blade, erb, go-template, jinja, minijinja, mojolicious, twig, xslate). Hono/CSR adapters evaluate real JS at runtime and render this like any other `.map`.

Fixture: `packages/adapter-tests/fixtures/methods/array-map-function-reference.ts`. Pin: `packages/adapter-go-template/src/conformance-pins.ts:134` (`array-map-function-reference`), and the equivalent per-adapter `conformance-pins.ts` entries.

## Root cause

There is no arrow body to serialize into the runtime evaluator's `ParsedExpr` JSON — `format` is a bare identifier, not an inline callback — so the `UNSUPPORTED_METHODS` gate refuses the shape loudly (correctly, per the "no silent lossy codegen" contract) rather than dropping the callback.

## Related precedent

This is the same class of gap as #2090 ("Function-reference sort comparators (`arr.sort(myCmp)`) are refused with BF021 — needs scope resolution"), which was already split out of the known-limitations catalog as its own tracked issue for `.sort()`. This issue is the `.map()` analogue — resolving a top-level (or otherwise in-scope) named function reference to its parsed body so it can be treated exactly like an inline arrow callback for evaluator serialization purposes.

Also referenced in the fixture's own comment as a "#2073 follow-up" — but #2073 ("Value-producing `.map(cb).join(sep)` can't lower (BF101)") is a **distinct**, already-resolved gap about a `.map()` whose callback is an inline arrow with an expression body; that one lowers today via the structured evaluator catalogue. This issue is specifically about the named-function-reference callee, which #2073's fix did not touch.

## Proposed fix

Resolve the named function reference (module- or component-scope) to its declaration, and if its body is itself serializable to the evaluator subset (a single-expression arrow, same shape #2073 already handles for inline callbacks), treat the resolved body as if it had been written inline. Falls back to the current `BF101` refusal if the reference can't be resolved (imported from elsewhere, reassigned, etc.) — same escape hatch pattern (`/* @client */`) as every other unsupported shape.

kfly8:
Fixes #2206. Fourth in the #2204 sub-issue PR stack (stacked on #2213/#2211/#2210).

## Problem

```tsx
const format = (t: string) => '#' + t
function TagLine({ tags }: { tags: string[] }) {
  return <p>{tags.map(format).join(' ')}</p>
}
```

`tags.map(format)` — a named function reference rather than an inline arrow — refuses with `BF101` on all 8 non-Hono template adapters: there's no arrow body to serialize into the runtime evaluator's `ParsedExpr` JSON.

## Fix

Precedent: #2090 solved the identical problem for `.sort(fnref)` comparators, entirely in the shared compiler layer (`packages/jsx/src/jsx-to-ir.ts`), resolving a bare-identifier callback one hop through same-file scope (`findLocalConst`/`findLocalFunction`) and re-parsing the analyzer's stored declaration text into a fresh `ts.Expression`.

`.sort()`'s resolution runs *during* the AST walk (inside `transformMapCall`, with a live `ts.Node`/`TransformContext`). But value-returning `.map(cb).join(...)` is architecturally separate — it's flattened to a plain string early and only re-parsed generically, with no scope context, by `attachParsedExpressions`/`parseExpression` after the whole IR tree is built.

Rather than threading a resolver through `parseExpression`'s ~10-deep recursive call graph, this adds `resolveCallbackMethodFunctionReferences`: a **post-parse pass** over the already-built `ParsedExpr` tree in `attachParsedExpressions`. `convertNode` already happily produces a generic `call` node with an `identifier` arg for `tags.map(format)` — no special-casing needed there. The pass walks the tree; for every `<recv>.<CALLBACK_METHODS method>(<bare identifier>)` call, it resolves the identifier via the same `findLocalConst`/`findLocalFunction`/`parseConstInitializer`/`parseFunctionInfoAsExpr` machinery #2090 built (via a new sibling resolver, `resolveCallbackMethodFunctionReferenceIdentifier`, to keep each call site's refusal semantics independently reviewable — mirrors the existing `resolveInterleaveTagIdentifier` precedent) and splices in the resolved arrow if the body folds to a single expression.

- Same semantics as #2090: one hop only (no alias chains), component-scope beats module-scope, cross-kind ambiguity (const *and* function decl for the same name) refuses rather than guesses, `/* @client */` always suppresses regardless of resolvability.
- Generalizes for free to every `CALLBACK_METHODS` name in value position (`filter`, `sort`, `toSorted`, `reduce`, `reduceRight`, `every`, `some`, `find`, `findIndex`, `findLast`, `findLastIndex`, `flatMap`), not just `.map` — same shared code path.
- `findLocalConst`/`findLocalFunction`'s parameter narrowed from the full `TransformContext` to just the `AnalyzerContext` they actually read (a type-only refactor; all 7 existing call sites updated) since `attachParsedExpressions`'s call site only has the analyzer, not a live source-bound context.

## Fallout: a latent bug this surfaced in 5 adapters

Testing against real PHP/Python/Ruby/Rust toolchains, Blade/Twig/Jinja/minijinja/ERB's `renderExpression` (text-position rendering) wasn't threading `expr.parsed` through to `convertExpressionTo*` — it was silently re-parsing the raw string fresh instead, discarding my resolution (and any other IR-attached optimization for that position). go-template, Mojolicious, and Xslate already did this correctly (mirrors `convertExpressionToGo(expr.expr, classify, expr.parsed)`). Fixed all 5 to thread `expr.parsed` through, matching the existing correct pattern.

## Testing

- New `packages/jsx/src/__tests__/map-function-reference.test.ts` (8 tests): const-arrow resolution, function-declaration resolution, component-scope shadowing a module-scope const, unresolvable import stays refused, non-function const stays refused, `/* @client */` suppression, inline-arrow shape unaffected, generalization to value-position `.sort(fnref)`.
- `bun test packages/jsx/src/__tests__` — 2309 pass, 0 fail (full suite, no regressions).
- `bun test packages/adapter-{blade,twig,jinja,rust,erb}/src/__tests__` (real PHP/Python/Rust/Ruby toolchains) — all green, `array-map-function-reference` fixture renders byte-correct (`#perl #go`) end-to-end.
- `bun test packages/adapter-{go-template,mojolicious,xslate}/src/__tests__` — compile clean (real Go/Perl+Mojolicious/Perl+Text::Xslate toolchains unavailable in this sandbox for full render verification).
- `bun test packages/adapter-hono` / `csr-conformance.test.ts` — no regressions (this fixture has no `'use client'` marker, so client JS generation is untouched by this change).
- Removed the `array-map-function-reference` `BF101` pin from all 8 adapters.
- Changeset added for `@barefootjs/jsx` + all 8 adapter packages.
- `ui/compat.lock.json`: no diff from this PR's own changes (no `ui/` component currently uses this shape) — left untouched on this branch rather than re-committing an unrelated stale-state catch-up diff from an earlier PR in the stack.

## Review plan

---


kfly8:
## Fable review summary

Requested a fable review of this PR's design (post-parse `ParsedExpr` tree resolution for bare-identifier `.map`/`.filter`/… callbacks) and the follow-up dist-staleness fix. Full write-up: [fable's review](https://github.com/piconic-ai/barefootjs/pull/2214#issuecomment-fable-review) (see commits below for the fix).

**Real bug found and fixed** (commit `b129bfb`): `resolveCallbackMethodFunctionReferences` resolved a bare identifier in callback position against the module/component const-and-function tables without checking whether the name was actually bound by an *enclosing* scope instead — an arrow's own param (`rows.map(fn => fn.tags.map(fn).join(','))`) or a loop's item/index variable (`fns.map(format => <li>{tags.map(format).join(' ')}</li>)`). Both shapes refused loudly with BF101 before this PR; after the original commit they would have silently spliced in the *wrong* function's body for SSR while CSR rendered correctly — an SSR/CSR divergence. Fixed by threading a `bound: ReadonlySet<string>` through the resolver (mirrors `expression-parser.ts`'s existing `freeVarsInBody` pattern) and through `attachParsedExpressions`'s loop recursion. 5 new regression tests added (13 total in the suite now).

**Other findings addressed in the same commit:**
- Changeset said "5 adapters" got the `.parsed`-threading fix; corrected to 7 (Mojolicious and Xslate were included too, not "already correct" as originally claimed).
- Updated the now-stale `array-map-function-reference` fixture docstring and two dangling adapter-test comments referencing a removed conformance pin.
- Documented (docstring + pinning test) the existing one-hop resolution limit: a callback nested inside another resolved declaration's body isn't re-walked. This fails safe (still refuses with BF101) and is unchanged behavior from before this PR — flagged by fable as worth a note, not a bug.

**Confirmed correct / no action needed:**
- The post-parse tree walk covers all 16 `ParsedExpr` kinds; exhaustive via the switch's declared return type.
- Leaving the resolved node's `kind` as `'call'` (not upgrading to `'array-method'`) is safe — the parser never produces `array-method` for higher-order callback methods regardless of whether the arg is a literal arrow or (now) a resolved one; `asCallbackMethodCall` is the single recognition point everything already routes through.
- The `dist/`-staleness bug found while verifying this PR locally (`packages/compat` resolves adapters via their published `dist/` entrypoints) was already the right minimal fix — CI's own `ci-compat.yml` rebuilds adapters before running `compat:lock` for exactly this reason; only the local workflow (and the committed lock file) were stale. A follow-up issue could add the same rebuild-before-lock guard to the local `compat:lock` script, but that's out of scope for this PR.
- `findLocalConst`/`findLocalFunction`'s signature narrowing (`TransformContext` → `AnalyzerContext`) is a safe type-only refactor.

All tests pass (3558 pass, 0 fail across `packages/jsx` + `packages/adapter-tests`), `ui/compat.lock.json` regenerated with no diff from the shadowing fix (expected — the shipped fixture has no shadowing), and `tsgo --noEmit` is clean.

Requesting Copilot review next.

---

kfly8:
Copilot reviewed all 22 changed files and generated no comments — CI is fully green (37/37 checks passing). Review cycle (fable → fix → Copilot) is complete for this PR.

---

Issue #2215, opened 2026-07-11
Dynamic/signal-derived dangerouslySetInnerHTML unsupported on template adapters


kfly8:
Follow-up to #2207.

#2207's fix lowers `dangerouslySetInnerHTML={{ __html: expr }}` on the 8 non-Hono template adapters (blade, erb, go-template, jinja, minijinja, mojolicious, twig, xslate) **only when `expr` is a compile-time string literal** (`{ __html: '<b>bold</b>' }`). This is a deliberate, security-motivated scope restriction: the literal is spliced directly into the adapter's own template source as trusted text (same trust domain as hand-writing the HTML), guarded per-adapter against that language's own template metacharacters (Blade `{{`/`{!!`/`<?`, Go/Jinja/Twig/minijinja `{{`/`{%`/`{#`, ERB `<%`, Mojolicious `<%`/line-code `%`, Xslate `<:`/line-code `:`).

A **dynamic or signal-derived** `__html` value (`dangerouslySetInnerHTML={{ __html: html() }}`, a prop, a template literal, a local const, etc.) still refuses loudly with `BF101` on all 8 template adapters — this is correct/expected behavior for now, not a bug. Hono and CSR already support this today (the client's `emit-reactive.ts` drives a `createEffect`-based `el.innerHTML = ...` assignment), so this is purely a template-adapter gap.

## Why this is out of scope for #2207

Embedding a dynamic, request/prop/signal-derived value directly into a template language's raw/unescaped-output sink (`|safe`, `|raw`, `{!! !!}`, `template.HTML(...)`, etc.) at SSR render time is a first-class injection vector unless the target language's runtime also HTML-escapes/sanitizes the value before marking it raw — which none of these 8 template runtimes do by default. Getting this right requires either:

1. A runtime sanitization step before the raw-output marker (adds a dependency + trust boundary per adapter), or
2. Accepting that the developer is fully responsible for the HTML's safety at render time (matches React/Hono's own trust model for `dangerouslySetInnerHTML`), which is a deliberate policy call a maintainer should make explicitly rather than one bundled quietly into #2207's literal-only fix.

## Suggested follow-up

- Decide the trust model for the dynamic case (bare pass-through, matching React/Hono, vs. some opt-in sanitization).
[...]

Issue #2208, opened 2026-07-11
Named local array-literal consts can't bind as a loop source (BF101) — cross-adapter, surfaced by #2205


kfly8:
Surfaced while fixing #2205 (conformance harness `siblingTemplatesRegistered` flag).

## Symptom

```tsx
export function StaticList() {
  const items = [{ label: 'Alpha' }, { label: 'Beta' }]
  return (
    <ul>
      {items.map(item => (
        <ListItem key={item.label} label={item.label} className="text-sm" />
      ))}
    </ul>
  )
}
```

The `static-array-children` fixture (`packages/adapter-tests/fixtures/static-array-children.ts`) refuses with `BF101` on all 8 non-Hono template adapters: *"Loop array `items` is a local computed value (...) that the Go template adapter cannot bind as a template variable — only a string-derived local resolves to a generated struct field."*

This was previously masked because the fixture always hit `BF103` first (sibling-imported child component in a loop body — see #2205) and the conformance harness's `expectedDiagnostics` path returns early on the first pinned diagnostic without exercising the render path further. Now that #2205 silences the spurious `BF103`, this second, pre-existing, orthogonal gap is reachable for the first time.

## Root cause

Each adapter's `renderLoop` (Go: `packages/adapter-go-template/src/adapter/go-template-adapter.ts:5039-5054`; mirrored per the in-code comment by Jinja/ERB and the rest of the shared-gate family) refuses ANY function-scope local `const` array used as a loop source unless it's string-derived (`isStringExpr`) — module-scope consts are handled separately (statically evaluated and seeded elsewhere). This check doesn't distinguish a **fully-static array literal** (this fixture's `[{ label: 'Alpha' }, { label: 'Beta' }]`, entirely known at compile time) from a **runtime-computed** expression (e.g. `static-array-from-props-with-component`'s `Object.entries(props.tags).filter(...)`, which #2069 already declared a known, accepted limitation since it depends on request-time prop values).

`inline-array-map` (mapping the same kind of array literal directly inline, with no named local const) does NOT hit this check — confirming the gate is keyed on "is this a named local identifier looked up in `localConstants`", not on whether the value is actually static.

## Proposed fix

[...]

Issue #2038, opened 2026-06-29
SSR: nested higher-order callback inside a predicate body degrades to its receiver in the Kolon-lambda fallback (lossy)


kfly8:
## Summary

When a higher-order **predicate** body itself contains a *nested* higher-order callback call (e.g. `arr.filter(e => !other.some(r => r.x === e))`), the SSR adapters cannot serialize the predicate to the evaluator (`serializeParsedExpr` refuses nested arrows → `null`), so they fall back to the Kolon/`$bf.*` lambda form. In that lambda, the nested callback call has no scalar form and is **degraded to just its receiver** (e.g. the inner `.some(...)` collapses to `other`), which can silently change predicate semantics.

Surfaced by Copilot on #2037:
> If `renderPredicateEval(...)` returns null, this function always falls back to a Kolon-lambda form. That fallback cannot faithfully lower predicates that contain nested higher-order callback calls (e.g. `x.items().filter(...).length > 0`) because nested callback calls have no scalar Kolon form and get degraded (via `XslateFilterEmitter.callbackMethod`) to just their receiver.

## Scope / pre-existing

This is **not** introduced by the EXPR2 collapse (#2037). The same lossy lambda fallback existed in the pre-P5 folded model (a nested `higher-order` node inside a predicate likewise poisoned serialization and fell to the identical lambda path). #2037 faithfully preserves the prior behavior (eval-first `?? legacy-lambda`), so it is **behavior-neutral** and the fix is deferred here rather than expanding that PR's scope.

The degradation is present in the Xslate filter emitter (`XslateFilterEmitter.callbackMethod` → receiver) and the analogous Mojolicious (`MojoFilterEmitter.callbackMethod` → receiver / `'1'`) and Go template predicate fallbacks.

## Repro surface

`site/ui/components/gallery/social/thread-demo.tsx`:
```tsx
reactionEmojis.filter(e => !comment.reactions.some(r => r.emoji === e))
```
The outer `.filter` predicate `!comment.reactions.some(...)` contains a nested `.some(...)`. (This component is `"use client"`, so it runs natively on the client and is not in the SSR render-conformance gate — which is why the lossy SSR path isn't caught by existing fixtures.)

## Proposed fix

[...]

PR #2217, opened 2026-07-11
Fix #2207: lower dangerouslySetInnerHTML with a compile-time string literal

### LINKED ISSUE #2207: dangerouslySetInnerHTML has no raw-output lowering on any template adapter (BF101)
Refs #2204 (parent tracker).

## Symptom

```tsx
export function DangerousInnerHtml() {
  return <div dangerouslySetInnerHTML={{ __html: '<b>bold</b> &amp; safe' }} />
}
```

Refuses with `BF101` on all 8 non-Hono template adapters (blade, erb, go-template, jinja, minijinja, mojolicious, twig, xslate). Hono renders it correctly (raw, unescaped markup — `expectedHtml`: `<div bf-s="test"><b>bold</b> &amp; safe</div>`).

Fixture: `packages/adapter-tests/fixtures/dangerous-inner-html.ts`. Pin: `packages/adapter-go-template/src/conformance-pins.ts:140` (`dangerous-inner-html`), and the equivalent per-adapter `conformance-pins.ts` entries.

## Root cause

`dangerouslySetInnerHTML` requires a deliberate raw/unescaped-output affordance in the target template language — e.g. Go `template.HTML`, Ruby `.html_safe`, Twig `|raw`, Jinja `|safe`, PHP Blade `{!! !!}`. No lowering to any of these exists yet, so each adapter refuses the shape loudly (correctly — the alternative would be silently entity-escaping the markup so tags render as visible text, which is worse than refusing).

## Proposed fix

Per adapter, lower the `dangerouslySetInnerHTML={{ __html: expr }}` prop to that adapter's raw-output affordance instead of the normal (escaped) text/attribute emission path:

- Go: wrap in `template.HTML(...)` (requires the value to flow through a typed field, since `html/template` auto-escapes `string` but not `template.HTML`).
- ERB: `.html_safe` (or `raw(...)`).
- Jinja/minijinja: `|safe` filter.
- Twig: `|raw` filter.
- Blade: `{!! !!}` echo syntax instead of `{{ }}`.
- Mojolicious/Xslate: each engine's equivalent raw-output construct.

Since this is a deliberate unescaped-output sink, consider whether the compiler should require the value to be a **static** literal (no dynamic/signal-derived HTML) for the initial cut, given the injection-safety stakes of getting this wrong — a bug here is a straight path to unescaped-output injection, so this needs careful review before landing, not a quick per-adapter string substitution.

kfly8:
Fixes #2207. Fifth and final in the #2204 sub-issue PR stack (stacked on #2214 / the #2206 fix).

## Problem

`dangerouslySetInnerHTML={{ __html: expr }}` refused with `BF101` on all 8 non-Hono template adapters (blade, erb, go-template, jinja, minijinja, mojolicious, twig, xslate). Hono/CSR already rendered it correctly.

## Design

This is the most security-sensitive issue in the stack (a raw-HTML injection sink), so before implementing I consulted fable for an architecture review — full reasoning in the PR discussion. Key decisions:

- **Scope, deliberately narrow for v1**: only a compile-time string-literal `__html` value (`dangerouslySetInnerHTML={{ __html: '<b>bold</b>' }}`) is lowered. A dynamic value (signal, prop, template literal with substitutions, local `const`) still refuses with `BF101` — tracked separately as #2215, not folded into this fix, since embedding a non-compile-time-known string directly into a template language's raw-output sink at SSR time is a first-class injection vector unless that runtime also sanitizes it, which none of these 8 template runtimes do by default.
- **No runtime raw-output primitive** (`|safe`, `|raw`, `{!! !!}`, `template.HTML(...)`, etc.) is used for the static case — the literal is spliced directly into the adapter's own template source as trusted text (same trust domain as hand-writing the HTML), since the value is already fully known at compile time. Using a runtime raw-output primitive here would reopen a template-source injection surface for no benefit.
- **Per-adapter template-metacharacter guard**: since the literal is spliced into another language's source, it's guarded against that language's own template syntax (`{{`/`{%`/`{#` for Go/Jinja/minijinja/Twig, `<%` for ERB/Mojolicious, `{{`/`{!!`/`<?`/`@directive` for Blade, `<:` for Xslate — plus Mojolicious/Xslate line-code) so a literal containing one of those sequences refuses loudly instead of silently being reinterpreted as a live template construct.
- **Single shared policy module**: recognition, static-literal extraction, and all 8 metachar guards live in `packages/jsx/src/adapters/dangerous-inner-html.ts` — one reviewable place for the injection-safety-relevant policy — rather than duplicated ad hoc per adapter. Each adapter only wires the 3 shared functions into its existing `renderElement`/`renderAttributes`.

## Fixtures

- `dangerous-inner-html` (existing): now renders on every adapter, not just Hono — docstring/description updated accordingly.
- `dangerous-inner-html-dynamic` (new): pins the dynamic-value `BF101` refusal on all 8 template adapters (attributed to #2215 in `conformance-pins.ts`) and proves Hono/CSR still render it correctly.

## Testing

- New unit test suite `packages/jsx/src/__tests__/dangerous-inner-html-resolver.test.ts` (28 tests): static/dynamic recognition (string-key form, shorthand, local const, extra property, no-substitution vs. substituted template literals), all 8 adapters' metachar guards (positive and negative cases), the diagnostic shape.
- `bun test packages/jsx packages/adapter-tests packages/adapter-{blade,erb,go-template,jinja,rust,mojolicious,twig,xslate}` — 8114 pass, 0 fail (real PHP/Ruby/Python/Rust rendering for blade/erb/jinja/rust/twig; go/mojolicious/xslate compile-verified only, missing toolchains in this sandbox — real toolchains covered in CI).
- `bun run build && bun run compat:lock` — clean; `dangerous-inner-html` drops out of the divergence table, `dangerous-inner-html-dynamic` appears pinned to #2215 on all 8 adapters.
- `bunx biome check` on all touched files — clean.
- `tsgo --noEmit` on `packages/jsx` and all 8 adapter packages — clean.

## Review plan

Per the tracking issue's review workflow: Fable review first (already consulted on design; will re-review the diff), then GitHub Copilot review, then merge readiness.

---


kfly8:
## Fable review summary

Requested a fable design review before implementing (given the injection-safety stakes), then a full review of the diff afterward. Fable verified the implementation against the real files, ran the new/existing tests, and probed all 8 adapters empirically with static/dynamic/metachar/empty-string inputs.

**Three issues found and fixed (commit `30d8272`):**

1. **Regression**: `resolveDangerousInnerHtml` never checked `attr.clientOnly`, so `dangerouslySetInnerHTML={/* @client */ { __html: html() }}` — a working escape hatch on `main` — started refusing with BF101. Fixed: returns `null` (render normally, defer to hydrate) when the attr is `clientOnly`.
2. **Security gap**: the Blade metachar guard didn't cover `<x-…>` component-tag syntax. This project's `BladeBackend` enables Laravel's component-tag compiler by default, so a literal containing `<x-alert>` would resolve and render a **live Blade component** — full component-class/view resolution, sharper than a bare `{{ }}` interpolation. Fixed: added `</?\s*x[-:]` to the guard.
3. **Fail-open default**: an unrecognized `adapterId` (e.g. a future 9th template adapter) returned `null` (safe to splice) instead of refusing. Fixed: fails **closed** — an unknown adapter always refuses.

Also applied two nice-to-haves: widened Mojolicious/Xslate's line-code whitespace class from `[ \t]` to `\s`, and reworded the diagnostic's suggestion text to name the actual working remediation (`/* @client */`) instead of a "move into a `'use client'` component" suggestion fable verified doesn't work.

**Confirmed correct** (no changes needed): the go-template adapter's structurally-different wiring (the `errors` getter, the `ParsedExprEmitter`/`objectLiteral` dispatch interaction), all 8 adapters' `renderAttributes` double-report avoidance, ERB/Xslate/Jinja/Twig/minijinja guard completeness for their default syntax, escape-bypass impossibility (the parser decodes JS string escapes before both the guard and the splice, so they can't diverge), and Go's `html/template` context-aware-escaper edge case (documented as a caveat, not guarded — output corruption at worst, never injection, same as hand-writing malformed HTML into the template).

10 new regression tests added (38 total in the resolver suite): the `clientOnly` fix, empty-string literal, Blade `<x-...>` tags (open and close), leading-whitespace and string-start line-code forms, a metachar-only literal, and the fail-open test flipped to assert fail-closed.

Full suite green: 8124 pass, 0 fail across `packages/jsx` + `packages/adapter-tests` + all 8 adapter packages. `tsgo --noEmit` and `bunx biome check` clean. `bun run build && bun run compat:lock` — no diff (no `ui/` component uses this shape).

Requesting Copilot review next.

---

kfly8:
Copilot's 3 comments were all the same underlying doc-accuracy issue (comments implying template literals are always dynamic, when a no-substitution template literal actually resolves as static — matches the parser and existing tests). Fixed in `8e02687` and all 3 threads resolved.

CI is fully green (36/36 checks passing). Review cycle (fable design consult → implementation → fable review with 3 real fixes → Copilot review with 3 doc fixes) is complete for this PR.

---

kfly8:
Resolved the merge conflict with `main` (now that #2214 merged, retargeting this PR's base) — one real conflict in `packages/compat/src/__tests__/compat-engine.test.ts` where main's own #2208 issue-URL fix and this PR's #2215 fix both touched the same expected `issues` array; combined into the union of all three (#2038, #2208, #2215) that go-template's BF101 pins now carry. Rebuilt all 8 adapter dists (stale after the merge, same class of issue as the earlier lockfile CI failure) and regenerated `ui/compat.lock.json` (no drift). Full suite green: 8211 pass, 0 fail. `mergeable_state` is now `clean`.

---

Issue #987, opened 2026-04-23
RFC: Type-based reactive primitive detection (one-shot resolution pass)


kfly8:
## Background

The compiler currently identifies reactive primitives (\`createSignal\`, \`createMemo\`, \`createEffect\`, \`onMount\`, \`onCleanup\`) by matching exact identifier names at specific AST shapes. This has known blind spots:

- \`import { createSignal as sig }\` — rename imports
- \`Solid.createSignal(0)\` — namespace-qualified calls
- \`const dispose = createEffect(...)\` — disposer capture (analogous to the \`createSignal()[0]\` issue solved in #985)
- Re-exports or factory wrappers

PR #985 patched the signal side with extra AST shape recognition. The same class of issue will recur for \`createMemo\` and especially \`createEffect\` as users adopt SolidJS-style patterns. A structural fix is better than per-pattern patches.

## Measured reality (#986)

Baseline on \`site/ui/components\` (196 files):

| Metric | Current |
|---|---|
| Overall wall time | ~1.4 s |
| Avg / p95 / p99 per file | 6.55 / 13.17 / 30.32 ms |
| Files triggering \`ts.createProgram\` | 1 / 196 (0.5%) |
| Measured per-Program overhead | **~545 ms** (documented comment says ~220 ms — stale) |

**A naive migration that ran Program for every file importing \`@barefootjs/client\` would push builds from ~1.4 s to ~80 s** (~60x regression). The refactor has to be carefully architected to be affordable.

## Goals

1. Recognize reactive primitives regardless of import alias, namespace, or AST shape variation.
2. Keep the per-file wall time within a strict budget (proposal: ≤ 2x current avg, ≤ 1.5x p95).
3. Eliminate structural risk of "checker queried outside the one-shot pass" (the multi-invocation problem previously encountered).

## Design principle — one-shot resolution pass

Do the type work **once per file at the start**, not per-query.

\`\`\`
analyzeComponent(source)
  │
  ├─ (1x) import prescan: detect @barefootjs/client usage
  ├─ (1x) if needed: create ts.Program + checker
  ├─ (1x) RESOLUTION PASS:
  │        walk all CallExpressions
  │        build Map<ts.Node, PrimitiveKind>
  │
  └─ later phases (jsx-to-ir, ir-to-client-js, resolver)
        → lookup only; checker reference is unreachable from here
\`\`\`

[...]

PR #2329, opened 2026-07-19
feat(compiler): cross-file + object-return reactive factory inlining, BF111/BF112 diagnostics (#2325)

### LINKED ISSUE #2325: Support user-authored reactive-factory helpers: cross-file resolution + object-return shapes (#931 round 2)
_Draft — the deferred "round 2" of #931, surfaced while decomposing Sora's (`piconic-ai/sora`) `App.tsx` state management into an importable hook. Investigated and designed against the current `0.21.4` source this session._

## Context

PR #936 gave us same-file reactive-factory inlining. A local `function foo(...) { …; return [a, b] as const }` that wraps a reactive primitive is source-text-inlined into any `'use client'` component that calls it as `const [x, y] = foo(...)`, so every downstream phase sees ordinary `createSignal(...)` as if hand-written. The whole feature is a **pre-pass** at `packages/jsx/src/analyzer.ts:175-185`, running before `createAnalyzerContext`:

- `prescanReactiveFactoriesInSource` (`analyzer.ts:3879`) → `detectReactiveFactory` (`analyzer.ts:3906`): scans **the current file's module-scope function declarations only** (explicit comment at `3897`).
- `rewriteFactoryCallsInSource` (`analyzer.ts:4022`) → `maybeRewriteDecl` (`analyzer.ts:4050`): pure source-text splice with param substitution, per-call-site suffix renaming of local bindings (`_bf{N}`), and return-tuple→caller-name renaming.
- `validateReactiveFactoryCalls` (`analyzer.ts:4148`, code `BF110`): emits a diagnostic for tuple-destructures of unrecognized callees, replacing the old silent break.

Two shapes were deliberately left out (#931 "Out of scope" §1 and the tuple-only detection rule):

1. **Cross-file helpers** — `detectReactiveFactory` only ever runs over the entry file's own AST (`prescanReactiveFactoriesInSource` calls `ts.forEachChild(sourceFile, visitTop)` on the single parsed source at `3901`). An imported factory is invisible; the call site falls through to `BF110`.
2. **Object-return helpers** — `detectReactiveFactory` hard-requires `ts.isArrayLiteralExpression` on the return argument (`analyzer.ts:3926`) and `maybeRewriteDecl` bails unless the binding is an `ArrayBindingPattern` (`analyzer.ts:4051`). `return { a, b }` destructured as `const { a, b } = f()` is rejected.

### The concrete driver

Sora's `components/App.tsx` (736 lines) is dominated by a ~20-member list-CRUD + debounced-autosave state machine (signals `lists`, `activeIndex`; ~15 handlers: `navigate`, `createNewList`, `selectList`, `deleteListById`, `deleteCurrentList`, `commitActiveEdits`, `handleClearAllLists`, `scheduleSave`, `flushSave`, …). Extracting it into a `useListStore()` hook is blocked by **both** gaps at once:

- The hook wants to live in its own file → needs **(A)**.
- Returning ~20 members positionally (`return [lists, setLists, navigate, createNewList, …] as const`) and destructuring by position is unreadable and fragile; the ergonomic shape is `return { lists, setLists, navigate, … }` destructured as `const { lists, navigate, createNewList } = useListStore()` → needs **(B)**.

### A silent-failure hole that exists *today* (found while tracing)

`validateReactiveFactoryCalls` only inspects `ArrayBindingPattern` decls (`analyzer.ts:4158`). An **object**-destructured factory call — `const { lists } = useListStore()` — matches neither the inliner nor the `BF110` check, so it reproduces the exact original #931 catastrophe: broken client JS, `ReferenceError` at SSR, **zero diagnostics**. So (B) is not merely an ergonomic nicety; it closes a live regression surface.

## Why this matters

Custom hooks / composables that live in their own file and return a named record are the *default* idiom in React (`useSomething()` returning an object), SolidJS (custom `createX`), and Vue (composables). BarefootJS lists "SolidJS: Reactivity model" as a design reference. Right now the compiler supports only the least-idiomatic slice: same-file, positional-tuple. The natural shape a user reaches for first — imported hook returning an object — is precisely the one that silently breaks.

### (A) Cross-file resolution

There is already a **shipped precedent** for "read another file's source cheaply, no type checker, recognize a reactive shape": `scanImportedClientSignals` (`analyzer.ts:3597`) + `resolveRelativeImportToFile` (`analyzer.ts:3636`). It resolves relative imports via `fs.statSync` extension-probing (`.tsx/.ts/.jsx/.js` + `index.*`), `fs.readFileSync`s the raw text, and regex-scans for `/* @client */ export const [a,b] = createSignal(...)`. It is **relative-import-only** (bails on bare/aliased specifiers at `3600`) and swallows read errors (`3608`). This respects #987's hard rule: **no `ts.Program`/type-checker in the common path** (measured ~545 ms/file; a 196-file bench would go ~1.4 s → ~80 s).

The difference for factory inlining: `scanImportedClientSignals` only needs the exported *name* (it marks `ctx.importedClientSignalNames`). Factory inlining needs the **full function shape** — parameter names + body text — to splice in. But `detectReactiveFactory` already extracts exactly that (`ReactiveFactoryInfo`: `params`, `bodySource`, `returnTupleIdentifiers`, `localBindings`) from an AST node. So the mechanism is: reuse `resolveRelativeImportToFile` to find imported helper files, parse each with `ts.createSourceFile` (cheap — no Program), run the *existing* `detectReactiveFactory` over their module-scope functions, and merge results into the same `factories` map `rewriteFactoryCallsInSource` already consumes.

| Option for (A) | Verdict | Why |
|---|---|---|
| **A1. Resolve + parse imported helper files, reuse `detectReactiveFactory`, merge into the prescan map** | ✅ recommended | Reuses the entire existing rewrite path unchanged. Mirrors `scanImportedClientSignals`'s trust model exactly. No new `ts.Program`. |
| A2. Regex-scan imported files (like `scanImportedClientSignals` does) instead of AST-parsing | ❌ | Regex can extract a *name* but not a param list + multi-statement body reliably. Factory bodies are arbitrary code. Must parse. |
| A3. `ts.Program`-based module resolution + type checker | ❌ | Violates #987. ~60x build regression. Buys nothing the cheap path lacks here. |

**Recommendation: A1.** Keep the same restrictions as `scanImportedClientSignals`: relative imports only; non-relative/aliased-package imports out of scope (documented, not an error). Parsing an imported file is one `ts.createSourceFile` per *distinct imported helper file that actually exports a matching factory* — bounded and lazy (skip files with no reactive-primitive text, mirroring the `prescan.factories.size > 0` gate at `analyzer.ts:176`).

### (B) Object-return shapes

| Option for (B) | Verdict | Why |
|---|---|---|
| **B1. Extend the same factory concept: accept `return { a, b }` + object-destructure call sites, reuse suffix-rename + splice** | ✅ recommended | Same detection/rewrite pipeline; the return-shape and binding-shape are the only two branches that change. Property **shorthand only** (`{ a, b }`). |
| B2. Support property renames (`return { lists: myLists }`, `const { lists: l } = f()`) | ⚠️ partial / defer | Doable but doubles the rename-map bookkeeping and the caller-name mapping is no longer 1:1 positional. Recommend **out of scope** for this round; shorthand covers the idiom. |
| B3. Separate mechanism (a distinct "hook" concept) | ❌ | Pure duplication. The tuple vs object distinction is one AST branch, not a new feature. |

**Recommendation: B1**, shorthand-only. Object property order gives a deterministic index for the existing positional rename machinery; the caller destructures by *key*, so the rename map becomes `returnPropertyName → callerLocalName` keyed by property name rather than by position. Renames (`{ lists: myLists }`) are out of scope (report a diagnostic and decline — see BF111 below).

## In scope

**(B) Object-return:**
1. `return { a, b, … }` (property **shorthand** only) as a valid factory return, alongside the existing tuple form.
2. Object-destructure call sites: `const { a, b } = factory(...)`, matched against the factory's returned property names.
3. `BF110` extended to fire on object-destructures of unrecognized callees (closes the silent hole).

**(A) Cross-file:**
4. Factories declared in a **relative-imported** file (`./`, `../`), resolved via `resolveRelativeImportToFile`, parsed with `ts.createSourceFile`, detected via the existing `detectReactiveFactory`.
5. Function-declaration factories at module scope of the imported file (same shape restriction as same-file).
6. Auto-provisioning the inlined body's required **reactive-primitive imports** into the component file when absent (see risk table — this is mandatory, not optional).

## Out of scope
1. **Non-relative / package-aliased imports** (`import { useX } from '@/hooks'` or from a published package). Same restriction `scanImportedClientSignals` already lives with (`analyzer.ts:3600`). Bare-specifier factory call → `BF110`.
2. **Property renames in object returns/destructures** (`{ lists: myLists }`). → new `BF111` (below).
3. **Free non-reactive identifiers captured from the helper file's own scope** (helper-file-local `const`/imports the factory body references, e.g. `readStored`/`writeStored`). Detect and decline (→ `BF112`) rather than silently emit a dangling reference. Auto-importing arbitrary helper-file symbols is a strictly bigger problem; defer.
4. **Transitive / recursive factories** (imported factory that itself calls another factory). `rewriteFactoryCallsInSource` already refuses to recurse into factory bodies (`analyzer.ts:4045`); keep that.
5. **Arrow-function-const factories** — still deferred (same-file already only handles `FunctionDeclaration` at `3893`); not regressing, not extending here.

### (B) object-return

1. **`detectReactiveFactory` (`analyzer.ts:3906-3992`)** — after the tuple branch, add an object-literal branch. Where it currently requires `ts.isArrayLiteralExpression(expr)` (`3926`), also accept `ts.isObjectLiteralExpression(expr)`; require every property to be a `ShorthandPropertyAssignment` (reject `PropertyAssignment` renames → let it fall to null / BF111). Store the property names in `returnTupleIdentifiers` (or add a `returnKind: 'tuple' | 'object'` discriminant to `ReactiveFactoryInfo` in `types.ts:1469`).
2. **`maybeRewriteDecl` (`analyzer.ts:4050-4109`)** — where it requires `ts.isArrayBindingPattern(decl.name)` (`4051`), branch on `ObjectBindingPattern`. For object patterns, build `callerNames` by matching each binding element's **property name** to the factory's return binding names (not positional), and reject element renames/defaults/rest (`el.propertyName` present, `el.dotDotDotToken`) → decline so BF110/BF111 can fire. Reuse the existing suffix-rename (`4080-4085`), param-substitution (`4091-4096`), and return-name→caller-name rename (`4098-4102`) unchanged.
3. **`validateReactiveFactoryCalls` (`analyzer.ts:4148-4194`)** — where it does `if (!ts.isArrayBindingPattern(decl.name)) continue` (`4158`), also handle `ObjectBindingPattern`, so unrecognized object-destructure callees now emit `BF110` instead of silently breaking.
4. **`BF111`** (new, in the reserved `BF110-BF119` band at `errors.ts:88`): "Reactive factory object-return/destructure uses a property rename; only shorthand `{ a, b }` is supported."

### (A) cross-file

5. **New `prescanImportedReactiveFactories(source, filePath)`** near `prescanReactiveFactoriesInSource` (`analyzer.ts:3879`), modeled on `scanImportedClientSignals` (`3597`):
   - Parse the entry `source` for relative imports.
   - For each relative import (`startsWith('./')||'../'`), `resolveRelativeImportToFile` (`3636`), `fs.readFileSync` with the same silent-skip `try/catch` as `3606-3610`.
   - Gate: skip files whose text contains no `REACTIVE_PRIMITIVES` (`3863`) member — cheap `String.includes` guard, mirrors the `factories.size > 0` no-op discipline.
   - `ts.createSourceFile` the helper text, run the existing `detectReactiveFactory` over its module-scope `FunctionDeclaration`s, keyed by the **imported local name** (respecting `import { useX as useY }` aliasing, like `3630`).
6. **Merge** the imported factory map into the local one before `rewriteFactoryCallsInSource` at `analyzer.ts:175-178`. Local factories win on name collision.
7. **Import provisioning (mandatory).** After a cross-file inline, the spliced body references `createSignal`/etc. that the component file may not import. Detect which `REACTIVE_PRIMITIVES` the inlined body uses and ensure the component's `@barefootjs/client` import includes them (inject/augment the import in the rewritten source, or record them for the emit phase). Without this, cross-file inlining compiles-but-fails-to-bundle.
8. **Free-identifier guard (`BF112`).** Before inlining a *cross-file* factory, scan its body's free identifiers (not in `params`, not `localBindings`, not `REACTIVE_PRIMITIVES`, not globals like `localStorage`/`window`/`document`). If any resolve to a binding in the helper file's own scope, decline to inline and emit `BF112` ("factory `useX` references `readStored` from its own module; move it into the component file or inline the call"). This converts a would-be silent runtime break into a build error.

### Watch-mode cache invalidation

**Finding: already handled for the direct-import case — confirm and lock it with a test.**

The incremental build cache (`packages/cli/src/lib/build-cache.ts`) records, per source entry, `deps: Record<string, string>` (abs path → content hash), with `isEntryFresh` (re-hashes every recorded dep). Deps are harvested in `compileEntry` at `build.ts:2015-2020` via `collectRelativeImportDeps` (`build.ts:1926`), a shallow lexical scan of the entry source's direct relative imports that stats and hashes each resolved file.

Consequence: a component that imports a factory helper via a relative path already has that helper file recorded as a dep → editing the helper bumps its hash → the component recompiles. This rides existing rails; no new invalidation machinery needed.

Caveat: **not transitive** — if a factory helper file itself imports a second file the factory body depends on, edits to that second file won't invalidate the component. This is sidestepped because `BF112` already declines to inline factories with helper-file-local free identifiers, so the only cross-file content that affects output is the directly-imported helper, which *is* tracked. Add a test: edit an imported helper's factory body → dependent component recompiles with the change.

### Types
- `ReactiveFactoryInfo` (`types.ts:1469`): add `returnKind: 'tuple' | 'object'`.

## Difficulty / risk assessment

| Item | Surface | Honesty |
|---|---|---|
| (B) object detection in `detectReactiveFactory` | +1 AST branch at `3926` | **Easier than it looks.** Object-literal-of-shorthand is a near-mirror of array-of-identifiers. |
| (B) object-destructure rewrite in `maybeRewriteDecl` | +1 branch at `4051` | **Moderate.** The one real change: caller-name mapping goes from positional to key-matched. Everything downstream of the map is unchanged. |
| (B) BF110 object-destructure | `4158` | **Easy + high value** — closes a live silent-break hole. |
| (A) resolve + parse imported helpers | new fn, reuses `resolveRelativeImportToFile` + `detectReactiveFactory` | **Easier than it looks.** Both halves already exist and are trusted. Mostly glue. |
| (A) **import provisioning** for inlined primitives | new emit/rewrite step | **Harder than it looks.** The component often doesn't import `createSignal`; the inlined body needs it. Must augment the import or the bundle fails. Easy to forget, hard to catch without an end-to-end bundle test. |
| (A) **free-identifier capture** (helper-file-local refs) | `BF112` guard | **Harder than it looks — the real footgun.** Cross-file bodies routinely reference module-local helpers (the #930 `createPersistentSignal` reads `readStored`/`writeStored`). Same-file inlining never hits this because the refs stay in scope. Must detect and decline, or this resurrects the silent-`ReferenceError` failure #931 set out to kill. |
| Watch invalidation | none (rides `collectRelativeImportDeps`) | **Already solved for direct imports.** Only needs a confirming test; transitive is dodged by BF112. |
| `ts.Program` cost | none | **Zero risk** — design uses only `ts.createSourceFile`, per #987. |

## Test matrix

Extend `packages/jsx/src/__tests__/reactive-factory-inlining.test.ts` (same `describe`/`test` + `analyzeComponent`/`compileJSX`/`TestAdapter` conventions).

**(B) object-return:**
1. Object-return factory, `return { count, setCount }`, destructured `const { count, setCount } = f()` → 1 signal, getter/setter named correctly.
2. Object-return, subset destructure `const { count } = f()` (setter unused) → inlines, no error.
3. Object-return with custom-setter wrapper (`createPersistentSignal` object shape).
4. Object-return factory called twice → hygiene (`_bf0`/`_bf1` suffix) holds.
5. Object-destructure of an **unknown** callee → now emits `BF110` (regression-locks the closed silent hole).
6. Object-return **rename** `return { count: c }` or `const { count: c } = f()` → `BF111`, no inline.
7. Mixed: object-return factory whose body also has `createMemo` → both collected.

**(A) cross-file:**
8. Factory in `./hooks.tsx`, imported + tuple-destructured → inlines; component that does **not** import `createSignal` still produces bundleable client JS (asserts import provisioning).
9. Cross-file + object-return combination (the `useListStore()` marquee case) → works.
10. Aliased import `import { useCounter as useC }` → resolved under the alias.
11. Non-relative import (`from '@/hooks'`) factory call → `BF110` (out of scope, diagnosed not silent).
12. Cross-file factory whose body references a helper-file-local `const`/import → `BF112`, no dangling reference emitted.
13. Cross-file factory file with no reactive primitive → skipped by the cheap gate, no parse cost, `BF110` at call site.

**Watch/cache:**
14. Build component importing a factory helper; edit the helper's factory body; rebuild → component recompiles and output reflects the change.

---

### Critical files for implementation
- `packages/jsx/src/analyzer.ts` (prescan/detect/rewrite: 175-185, 3597-3665, 3863-4194)
- `packages/jsx/src/types.ts` (`ReactiveFactoryInfo`: 1469-1486)
- `packages/jsx/src/errors.ts` (`BF110`-`BF119` band: 88-89)
- `packages/cli/src/lib/build.ts` (dep harvesting: 1926-1952, 2015-2020)
- `packages/jsx/src/__tests__/reactive-factory-inlining.test.ts` (test conventions)

kfly8:
## Summary

Closes #2325. Round 2 of #931 (#936): user-authored reactive-factory helpers now work in the two shapes real code actually uses — imported from their own file, and returning a named object instead of a positional tuple.

### What this PR does

- **(B) Object-return factories** — `return { count, setCount }` (property shorthand only) is now a valid factory return, destructured as `const { count } = factory(...)`. Subset destructures are supported; undestructured returned members are suffix-renamed for per-call-site hygiene. Property renames / defaults / rest decline with the new **BF111**.
- **Closes a live silent-failure hole**: object-destructured factory calls previously bypassed both the inliner and BF110 — reproducing the original #931 catastrophe (broken client JS, SSR ReferenceError, zero diagnostics). Reactive-looking object destructures now diagnose (BF110/BF111/BF112) instead of silently breaking. The trigger set is deliberately narrower than the tuple rule — ordinary object destructures of non-reactive calls are left untouched.
- **(A) Cross-file factories** — factories exported from a **relative-imported** file (`./`, `../`, aliased imports supported) are resolved via the existing `resolveRelativeImportToFile` + `ts.createSourceFile` path (no `ts.Program`; #987's build-time budget respected — gated so factory-free files pay ~nothing), detected with the same `detectReactiveFactory`, and inlined through the unchanged rewrite pipeline. Local factories win on name collision.
- **BF112** — an imported factory whose body captures bindings from its own module scope (e.g. reading a helper-file-local constant) declines with a build error instead of emitting a dangling reference.
- **CLI**: `collectRelativeImportDeps` now probes `.jsx`/`.js` (+ index variants), matching the analyzer's own resolution, so helper edits invalidate the importer's cache entry in watch/incremental builds.

### Not needed (verified during implementation)

- No import-provisioning machinery: the client-JS `@barefootjs/client/runtime` import line is usage-derived, so inlined `createSignal` calls are provisioned automatically — pinned by test.
- No new cache invalidation machinery: relative-imported helpers were already dep-tracked by the build cache; transitive helper deps are moot because BF112 declines factories that reach outside their own body.

### Out of scope (unchanged from the issue)

Non-relative/package imports (diagnosed via BF110, not silent), property renames (BF111), transitive factories, arrow-function-const factories.

## Test plan

- [x] `bun run --filter '@barefootjs/jsx' test` — new object-return cases in `reactive-factory-inlining.test.ts` and a new `reactive-factory-cross-file.test.ts` covering the #2325 matrix, including the cross-file + object-return combination.
- [x] `bun run --filter '@barefootjs/cli' test` — dep-scanner `.js`/`.jsx` rows + new `build-factory-invalidation.test.ts` (edit helper → dependent component recompiles, untouched sibling stays cached).
- [x] `tsc --noEmit -p packages/jsx/tsconfig.json` — clean.
- [x] `bun run build` (root) — diffed against a clean `main` build in an isolated worktree; identical pre-existing failures (composer not installed in this environment), unrelated to this change.
- [x] 5 pre-existing `@barefootjs/cli` test failures (macOS case-insensitive-filesystem artifacts in `resolve-source.test.ts`/`meta-loader.test.ts`) confirmed present on clean `main` via `git stash` before/after comparison — unrelated to this branch.

### DIFF
--- a/.changeset/reactive-factory-cross-file-object-return.md
+++ b/.changeset/reactive-factory-cross-file-object-return.md
@@ -1,0 +1,12 @@
+---
+'@barefootjs/jsx': minor
+'@barefootjs/cli': patch
+---
+
+Reactive-factory helpers (#931 round 2, #2325): factories may now return a
+shorthand object literal (`return { count, setCount }`) destructured as
+`const { count } = factory(...)`, and may be imported from a relative path.
+New diagnostics: BF111 (property renames in factory object returns/destructures),
+BF112 (imported factory capturing helper-module bindings); BF110 now also covers
+reactive-looking object-destructure call sites that previously failed silently.
+CLI: dependency scanner tracks `.jsx`/`.js` relative imports for cache invalidation.
--- a/packages/adapter-tests/coverage-map.json
+++ b/packages/adapter-tests/coverage-map.json
@@ -2798,6 +2798,19 @@
         "text"
       ]
     },
+    "reactive-factory-object-return": {
+      "kinds": [
+        "call",
+        "identifier",
+        "literal"
+      ],
+      "axes": [
+        "literal:number"
+      ],
+      "contexts": [
+        "text"
+      ]
+    },
     "reactive-prop-binding": {
       "kinds": [
         "call",
@@ -4132,11 +4145,11 @@
     "array-method": 62,
     "arrow": 55,
     "binary": 67,
-    "call": 159,
+    "call": 160,
     "conditional": 18,
-    "identifier": 239,
+    "identifier": 240,
     "index-access": 12,
-    "literal": 197,
+    "literal": 198,
     "logical": 41,
     "member": 120,
     "object-literal": 43,
@@ -4181,7 +4194,7 @@
     "binary:>=": 1,
     "literal:boolean": 37,
     "literal:null": 22,
-    "literal:number": 85,
+    "literal:number": 86,
     "literal:string": 141,
     "logical:&&": 7,
     "logical:??": 37,
--- a/packages/adapter-tests/fixtures/index.ts
+++ b/packages/adapter-tests/fixtures/index.ts
@@ -1,4 +1,13 @@
 import { fixture as counter } from './counter'
+// Same-file reactive-factory helpers (#931, #2325): a helper wrapping
+// createSignal in a shorthand-object return inlines pre-analysis, so the
+// compiled output — and thus every adapter's marked template — must be
+// byte-identical to counter.ts's hand-written createSignal call. This is
+// the cross-adapter proof that #2325's object-return inlining generalizes
+// (the compiler unit tests already cover it against a single test
+// adapter; this fixture is what actually exercises Hono / Go / PHP /
+// Django / Rust / Ruby / Perl / ... through the shared harness).
+import { fixture as reactiveFactoryObjectReturn } from './reactive-factory-object-return'
 // Shared-component corpus (#1466): fixture-hydrate fixtures lifted
 // from `integrations/shared/components/` also participate in
 // cross-adapter HTML conformance so the same .tsx is guaranteed to
@@ -399,6 +408,7 @@ import type { JSXFixture } from '../src/types'
 
 export const jsxFixtures: JSXFixture[] = [
   counter,
+  reactiveFactoryObjectReturn,
   counterShared,
   toggleShared,
   conditionalReturnButton,
--- a/packages/adapter-tests/fixtures/reactive-factory-object-return.ts
+++ b/packages/adapter-tests/fixtures/reactive-factory-object-return.ts
@@ -1,0 +1,24 @@
+import { createFixture } from '../src/types'
+
+export const fixture = createFixture({
+  id: 'reactive-factory-object-return',
+  description:
+    'Same-file reactive-factory helper returning a shorthand object (#2325) inlines to a plain signal, identically to counter.ts',
+  source: `
+'use client'
+import { createSignal } from '@barefootjs/client'
+
+function createCounter(initial: number) {
+  const [count, setCount] = createSignal(initial)
+  return { count, setCount }
+}
+
+export function Counter() {
+  const { count, setCount } = createCounter(0)
+  return <button onClick={() => setCount(n => n + 1)}>Count: {count()}</button>
+}
+`,
+  expectedHtml: `
+    <button bf-s="test" bf="s1">Count: <!--bf:s0-->0<!--/--></button>
+  `,
+})
--- a/packages/cli/src/__tests__/build-factory-invalidation.test.ts
+++ b/packages/cli/src/__tests__/build-factory-invalidation.test.ts
@@ -1,0 +1,138 @@
+/**
+ * Build-cache invalidation for reactive-factory helper files (#2325).
+ *
+ * `collectRelativeImportDeps` records every relative import a component
+ * source references so an edit to an imported helper invalidates the
+ * importer's cache entry, not just edits to the component file itself. That
+ * scanner's extension probe list historically stopped at `.tsx`/`.ts` —
+ * missing `.jsx`/`.js` — while the analyzer's own `resolveRelativeImportToFile`
+ * (used to resolve cross-file reactive factories, #2325) already covers all
+ * four. This file pins the end-to-end consequence: a factory helper authored
+ * as plain `.ts` must be tracked as a dependency, so editing it triggers a
+ * real recompile of the importing component (not a stale cache hit) on the
+ * next build, while an untouched sibling component is still served from cache.
+ *
+ * Harness modeled on `build-manifest-components.test.ts` (`makeTmpDir`,
+ * `linkClientRuntime`, try/finally cleanup).
+ */
+
+import { describe, test, expect } from 'bun:test'
+import { build } from '../lib/build'
+import { loadCache } from '../lib/build-cache'
+import { TestAdapter } from '../../../jsx/src/adapters/test-adapter'
+import { mkdirSync, writeFileSync, rmSync, readFileSync, realpathSync, symlinkSync } from 'fs'
+import { resolve } from 'path'
+import { tmpdir } from 'os'
+
+function makeTmpDir(label: string) {
+  const dir = resolve(tmpdir(), `bf-test-${label}-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`)
+  mkdirSync(dir, { recursive: true })
+  return realpathSync(dir)
+}
+
+/**
+ * Link the workspace `@barefootjs/client` into the throwaway project so the
+ * build can copy `barefoot.js` — without it the compiled client bundles'
+ * `../../barefoot.js` imports don't resolve on disk and the rebuild pass
+ * raises BF053, which is noise unrelated to what this file pins.
+ */
+function linkClientRuntime(projectDir: string) {
+  const scope = resolve(projectDir, 'node_modules/@barefootjs')
+  mkdirSync(scope, { recursive: true })
+  symlinkSync(resolve(import.meta.dir, '../../../client'), resolve(scope, 'client'))
+}
+
+// The factory helper, deliberately `.ts` — not `.tsx` — so `discoverComponentFiles`
+// (which only picks up `.tsx`) never treats it as its own component entry,
+// while the relative-import resolution (both the analyzer's cross-file
+// factory resolution and this file's `collectRelativeImportDeps` dep
+// scanner) still finds it.
+function hooksSource(initial: number): string {
+  return `import { createSignal } from '@barefootjs/client'
+
+export function createCounter() {
+  const [count, setCount] = createSignal(${initial})
+  return { count, setCount }
+}
+`
+}
+
+const COUNTER_SOURCE = `'use client'
+import { createCounter } from './hooks'
+
+export function Counter() {
+  const { count, setCount } = createCounter()
+  return <button onClick={() => setCount(count() + 1)}>{count()}</button>
+}
+`
+
+// Untouched sibling — the cache-hit control.
+const SIBLING_SOURCE = `'use client'
+import { createSignal } from '@barefootjs/client'
+
+export function Sibling() {
+  const [value, setValue] = createSignal('sibling')
+  return <p onClick={() => setValue('clicked')}>{value()}</p>
+}
+`
+
+async function runBuild(projectDir: string, outDir: string) {
+  return build({
+    projectDir,
+    adapter: new TestAdapter(),
+    componentDirs: [resolve(projectDir, 'components')],
+    outDir,
+    minify: false,
+    contentHash: false,
+    clientOnly: true,
+  })
+}
+
+describe('build-cache invalidation for reactive-factory helper files (#2325)', () => {
+  test('editing a .ts factory helper invalidates only the importing component', async () => {
+    const projectDir = makeTmpDir('factory-invalidation-src')
+    const outDir = makeTmpDir('factory-invalidation-out')
+    try {
+      const componentsDir = resolve(projectDir, 'components')
+      mkdirSync(componentsDir, { recursive: true })
+      writeFileSync(resolve(componentsDir, 'hooks.ts'), hooksSource(0))
+      writeFileSync(resolve(componentsDir, 'Counter.tsx'), COUNTER_SOURCE)
+      writeFileSync(resolve(componentsDir, 'Sibling.tsx'), SIBLING_SOURCE)
+      linkClientRuntime(projectDir)
+
+      const first = await runBuild(projectDir, outDir)
+      expect(first.errorCount).toBe(0)
+      expect(first.compiledCount).toBe(2)
+
+      const counterClientJsPath = resolve(outDir, 'components/Counter.client.js')
+      const firstCounterJs = readFileSync(counterClientJsPath, 'utf8')
+      expect(firstCounterJs).toContain('createSignal(0)')
+
+      // The build-cache entry for Counter must record the helper file as a
+      // dependency — this is the #2325 fix under test (EXT_CANDIDATES now
+      // covers `.ts` via the same list the analyzer itself probes).
+      const cacheAfterFirst = await loadCache(outDir)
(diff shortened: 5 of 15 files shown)


kfly8:
Added a cross-adapter conformance fixture (`reactive-factory-object-return`) for the same-file object-return case, in response to a question about whether this feature is adapter-agnostic and conformance-tested.

**Answer**: architecturally yes (the factory-inlining pre-pass runs as pure source-text rewriting before `analyzeComponent` parses the file, so every adapter's emitter sees plain `createSignal(...)` with no trace of the factory), but this PR's own tests only proved it against a single synthetic `TestAdapter`. The new fixture proves it empirically: generated `expectedHtml` is byte-identical to `counter.ts`'s hand-written equivalent, and it passes against the real Hono and Go template adapters (Blade/Jinja skip in this dev environment for lack of php/python3, same as every other fixture).

The cross-file half (#2325 (A)) remains outside conformance-fixture reach — `JSXFixture.components` compiles in-memory, never touching the real filesystem the analyzer's cross-file resolution reads from. Same pre-existing gap as `scanImportedClientSignals`, not something new here.

PR #2370, opened 2026-07-23
Add Tetris and Markdown editor integration examples

kfly8:
## Summary

Two new onboarding showcases for the shared integration examples, aimed at making BarefootJS's capabilities easier to grasp for newcomers. Both live in `integrations/shared/components` (so every adapter compiles them) and are wired into the Hono example with routes, styles, and e2e coverage.

### 🎮 Tetris — a game loop on signals
`integrations/shared/components/Tetris.tsx`

Demonstrates that signal-based reactivity can drive a real-time game:
- A **self-scheduling `setTimeout` drop loop** that re-arms at the current level's speed.
- **Document-level keyboard controls** wired up in `onMount` (← → move · ↑ rotate · ↓ soft drop · Space hard drop · P pause).
- A derived **`display` memo** that overlays the falling piece on the settled board, rendered as a reactive **10×20 grid** (nested `.map`).
- All game rules (collision, rotation, wall-kicks, line clears, scoring, levels) live in **plain functions** — only the JSX reads signals, so the reactive graph stays small. SSR pre-renders the full board (no hydration layout shift).

### 📝 Markdown editor — one signal, many derived views
`integrations/shared/components/MarkdownEditor.tsx`

Demonstrates fine-grained derivation from a single source of truth:
- A single `text` signal drives a **live block-level preview** plus **word / character / line / read-time** counters, all through `createMemo`.
- The preview is one reactive list of **uniform blocks**, so the map lowers to a single JSX expression; inline **bold** segments render from a nested map over data **precomputed in the parser** (rather than a prop-driven child component, which severs client reactivity).
- Toolbar buttons (Bold / H1 / H2 / List / Quote / Code / Clear) do imperative textarea manipulation, showing heavy logic living outside the reactive graph.

## Testing

- Added shared e2e suites (`integrations/shared/e2e/{tetris,markdown-editor}.spec.ts`) and Hono wiring specs — covering hydration, keyboard input, the game loop / piece movement, live preview, counters, and inline formatting.
- Full Hono e2e suite green (**119/119**, including the 15 new tests).
- Verified the two shared components compile cleanly under the CSR, Go-template (echo), and Perl (mojolicious) adapters in addition to Hono.

## Generated files

Regenerated the four Go integrations' committed `components.go` (`chi`, `echo`, `gin`, `nethttp`) for the two new shared components, to satisfy the Go-template CI drift gate.

## Notes

- No changeset: changes are confined to `integrations/`, not published `packages/*/src`.
- While building the editor I hit a compiler gap worth a follow-up: a `.map()` callback with a **statement body containing multiple `if`-returns of JSX** is not lowered — it emits raw JSX into the client bundle (a runtime `SyntaxError`) instead of erroring at compile time. The examples here use the supported single-JSX-expression form; flagging separately in case it's worth a diagnostic.

---
_Generated by [Claude Code](https://claude.ai/code/session_01FG9KVo7k7nmVkvpuDXKCaC)_

### DIFF
--- a/integrations/chi/components.go
+++ b/integrations/chi/components.go
@@ -110,6 +110,42 @@ type FormProps struct {
 	Accepted bool `json:"accepted"`
 }
 
+// Seg represents a seg.
+type Seg struct {
+	Bold bool `json:"bold"`
+	Text string `json:"text"`
+}
+
+// Block represents a block.
+type Block struct {
+	Cls string `json:"cls"`
+	Segs []Seg `json:"segs"`
+}
+
+// MarkdownEditorInput is the user-facing input type.
+type MarkdownEditorInput struct {
+	ScopeID string // Optional: if empty, random ID is generated
+	BfParent string // Optional: parent scope id
+	BfMount string // Optional: slot id in parent
+}
+
+// MarkdownEditorProps is the props type for the MarkdownEditor component.
+type MarkdownEditorProps struct {
+	ScopeID string `json:"scopeID"`
+	BfIsRoot bool `json:"-"`
+	BfIsChild bool `json:"-"`
+	BfParent string `json:"-"`
+	BfMount string `json:"-"`
+	BfDataKey string `json:"-"`
+	Scripts *bf.ScriptCollector `json:"-"`
+	Text interface{} `json:"text"`
+	Blocks []Block `json:"blocks"`
+	CharCount int `json:"charCount"`
+	WordCount int `json:"wordCount"`
+	LineCount int `json:"lineCount"`
+	ReadMinutes int `json:"readMinutes"`
+}
+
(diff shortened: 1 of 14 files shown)


kfly8:
Closing in favor of addressing the underlying onboarding friction these examples surfaced, rather than shipping the examples with workarounds.

The most severe issue: a `.map()` callback with a **statement body containing `if`-returns of JSX** compiles with "0 errors" but emits **raw JSX into the client bundle** → runtime `Uncaught SyntaxError: Unexpected token '<'` → the whole component fails to hydrate, silently. SSR looks perfect and the build reports success, so there's no signal anything is wrong.

Follow-up work will fix the compiler (proper lowering and/or a compile-time diagnostic) and then reland these examples using the natural form as coverage. Superseded by that PR.

---

PR #2386, opened 2026-07-25
fix(jsx): route JSX-returning flatMap expression bodies through the segments carrier

kfly8:
Fixes **BUG-1** from the tagged-todo SSR/hydration audit (#2385) — first PR in a stack of three (BUG-1 → BUG-2 → BUG-3/4).

## Problem

A `.flatMap()` arrow with an **expression** body returning JSX:

```tsx
<ul>{todos().flatMap(t => t.tags.map(tag => <li key={`${t.id}:${tag}`}>{tag}</li>))}</ul>
```

fell through every `transformMapCall` dispatch arm to the `IRExpression` scalar path with **zero diagnostics**. `ctx.getJS` strips types, not JSX, so the raw callback — JSX included — spliced verbatim into the client bundle. The module is invalid JS (`Unexpected token '<'`), so **the whole component silently fails to hydrate** while SSR looks correct. The braced twin (`{ return t.tags.map(...) }`) already rode the `FlatMapCallback` structured-segments carrier; the unbraced form is the same shape minus the braces.

This is exactly the "raw JSX in the client bundle" leak class the Stage-3 root cure (`spec/callback-fidelity.md`) exists to prevent — it survived because the expression-body shape was outside both the dispatch and the `map-body-no-silent-divergence` sweep.

## Fix (three structural changes, sound-or-loud)

1. **Route the shape through the existing carrier.** `buildFlatMapCallback` now accepts a block *or* expression body (segment reconstruction and the SSR `rawBody` are position-based, so both flow through identically), and `transformMapCall` routes any JSX-bearing non-block flatMap body through it, parenthesized or not. SSR (Hono) and the hydrate template render it correctly.
2. **Structural net at the scalar fallthrough.** A map-like callback whose body carries an **inline JSX literal** but produced no loop lowering now refuses loudly (BF021) instead of leaking — any future recognizer gap in this class becomes a diagnostic, not a silent SyntaxError. (A JSX-helper *call* — `map(t => renderItem(t))` — carries no inline JSX literal and legitimately stays on the reactive-text path.)
3. **Adapter gate for segment-carried flatMap bodies.** A DSL template runtime can't execute the body at SSR — pre-gate, the Go adapter emitted `{{range …}}{{end}}` with an **empty body** (silent divergence, block bodies included). DSL targets now refuse with BF021 + the `/* @client */` escape, mirroring the const-preamble and array-builder gates. JS runtimes are unaffected.

## Tests / fixtures (change-time coupling)

- `map-body-no-silent-divergence.test.ts`: four new shapes (expression body, parenthesized, signal-array, JSX-inside-unrecognized-call) across JS + DSL tiers.
- `flatmap-segments.test.ts`: expression-body pins (segments carrier, `escapeText`, `_p.` rewrite in hydrate templates).
- New conformance fixture `flatmap-expression-body` (leaf data carries `<`, `&`, `"`, `'`) + BF021 `conformancePins` entries on all eight DSL adapters + regenerated `coverage-map.json`.
- The fixture is CSR-skip-listed with a pointer to the pre-existing CSR/SSR flatMap-leaf `data-key` asymmetry; the stacked client-loop rewiring PR removes the skip.

## Scope boundary

The client-side `mapArray` wiring for flatMap loops (un-flattened source, `null` keyFn, empty item template → item loss at hydration, `cloneNode(null)` crash on add) is a separate pre-existing defect shared by block and expression bodies alike — **fixed by the next PR in this stack** (BUG-2).

## Verification

- `bun test packages/jsx` — 2694 pass / 0 fail
- `bun test packages/adapter-tests` — 1425 pass / 0 fail
- `bun test packages/adapter-hono` — 972 pass / 0 fail
- `bun test packages/adapter-go-template` — 1177 pass / 0 fail

---

### DIFF
--- a/.changeset/flatmap-expression-body-segments.md
+++ b/.changeset/flatmap-expression-body-segments.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Route JSX-returning `.flatMap()` expression bodies through the structured-segments carrier. Previously `items.flatMap(t => t.tags.map(tag => <li key={...}>{tag}</li>))` (the unbraced twin of the supported block-body form) fell through to the scalar expression path and spliced raw JSX verbatim into the client bundle — invalid JS, so the whole component silently failed to hydrate. Also adds a structural net (any map-like callback with an inline JSX literal that produces no loop lowering now refuses loudly instead of leaking) and an adapter gate for segment-carried flatMap bodies on DSL backends (BF021 + `/* @client */` instead of an empty SSR loop body).
--- a/packages/adapter-blade/src/conformance-pins.ts
+++ b/packages/adapter-blade/src/conformance-pins.ts
@@ -39,6 +39,12 @@ export const conformancePins: ConformancePins = {
   'reduce-typeof-body': [{ code: 'BF101', severity: 'error' }],
   'reduce-right-typeof-body': [{ code: 'BF101', severity: 'error' }],
   'flatmap-typeof-projection': [{ code: 'BF101', severity: 'error' }],
+  // JSX-returning `.flatMap()` body carried as structured segments
+  // (expression body / block body): a JS runtime executes it verbatim; a
+  // DSL template runtime can't, so it refuses with BF021 + `/* @client */`
+  // (pre-gate this emitted an empty loop body — silent divergence).
+  // See spec/callback-fidelity.md.
+  'flatmap-expression-body': [{ code: 'BF021', severity: 'error' }],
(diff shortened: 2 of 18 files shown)


kfly8:
Closing as superseded by the restacked series **#2391 → #2388 → #2390**. This PR's expression-body routing (segments carrier + blanket DSL gate) was the part #2390 later reversed for pure projections — rather than merge an approach-then-correction sequence, the still-valid pieces were re-cut: the loud-refusal enforcement (structural net, DSL gate for statement bodies, leaf-wiring refusal) now lands in #2391, and the expression-body handling lands once, in its final neutral-IR shape, in #2390.

---

PR #2387, opened 2026-07-25
fix(jsx): reconcile flatMap loops over flattened leaf descriptors on the client

kfly8:
Fixes **BUG-2** from the tagged-todo SSR/hydration audit (#2385) — second PR in the stack, **based on #2386** (BUG-1). Merge #2386 first; this diff shows only the client-loop rewiring.

## Problem (browser-verified)

The client half of a JSX-returning `.flatMap()` loop was wired as if it were a plain `.map()`:

```js
mapArray(() => todos(), _s19, null, (t, __idx, __existing) => {
  if (t().tags.length > maxTags) return [];;   // statements-before-return leaked in as a bogus preamble
  if (__existing) return __existing;
  const __tpl = document.createElement('template');
  __tpl.innerHTML = ``;                        // EMPTY — the body lives in flatMapCallback segments
  return __tpl.content.firstElementChild.cloneNode(true)
}, 'l1')
```

- source = **un-flattened** items, keyFn = **null** → SSR leaves reconciled against source count with index keys: **leaves vanished at hydration with zero interaction** (4 → 3 in the audit; 4 → 2 in the minimal repro)
- empty item template → **`cloneNode(null)` crash** the first time an item is added
- no per-leaf wiring → the list **never reacted** to any data change

## Fix — the flattened leaves become the reconciliation domain

- **`renderFlatMapClientBody`** renders the existing segments carrier so each JSX leaf becomes a `({ k, h })` descriptor — `k` from the leaf's own `key` (template-literal keys included), `h` its HTML string. The mapArray accessor becomes `(<chained>).flatMap(params => body)`; keyFn reads `d.k` with an index fallback.
- The plain/branch loop plans gain a **`flatMapLeafItem`** renderItem: adopts the SSR element at hydration (trust-SSR first effect run), builds new leaves from `d.h`, and patches an existing leaf in place via the new `@barefootjs/client` **`patchLeaf`** helper when its HTML changes under a stable key (identity preserved — mapArray owns the node; `data-key` never touched).
- Leaf `data-key` moves off the string templates entirely (Hono SSR never emitted it), closing the documented CSR/SSR data-key asymmetry — including its **unescaped interpolation**; identity is stamped by mapArray via `setAttribute`. The csr-conformance skip from #2386 is removed.
- The shadow preamble is dropped for `flatMapCallback` loops (segments carrier = the single door), and the references graph now traces flatMap bodies so init-scope extractions (`const maxTags = _p.maxTags`) survive.
- **Sound-or-loud:** a flatMap leaf carrying an event handler, component, nested loop, or spread refuses with BF021 (whole-leaf patching can't wire them); reactive expressions inside leaves are allowed by design.

## Tests

- `flatmap-segments.test.ts`: pins the descriptor mapArray emission, preamble suppression, leaf-wiring refusal.
- `patch-leaf.test.ts`: runtime unit tests (identity preservation, attr sync, `data-key` exclusion, tag-mismatch warning).
- **New shared fixture `tag-cloud`** (flatMap block body: early return + `const` preamble + keyed leaves + `<`, `&`, `"` data), exercised end-to-end in a real Chromium by the fixture-hydrate runner: hydration adoption, in-place patch, client-side leaf creation, removal.
- BF021 `conformancePins` on all eight DSL adapters + regenerated coverage/compat/support-matrix artifacts.

## Verification

- `packages/jsx` 2704 pass · `adapter-tests` 1431 pass · `adapter-hono`+`adapter-go-template` 2155 pass · `packages/client` 537 pass · `compat` 191 pass · fixture-hydrate **33/33 in Chromium**
- The audit harness (#2385) re-run against this branch: every flatMap scenario now matches fresh SSR, including add (`cloneNode` crash gone, new leaves render with correct special-char keys). Remaining audit diffs are BUG-3 (delegated row handlers), next in the stack.

---


kfly8:
Closing as superseded by **#2391**, which carries this PR's content unchanged (descriptor `mapArray` reconciliation, `patchLeaf`, `tag-cloud` fixture + fixture-only source root) but re-cut directly onto `main` — absorbing the still-valid loud-refusal pieces of #2386 so the restacked series (**#2391 → #2388 → #2390**) has no add-then-remove churn in pins/skips.

---

Issue #1128, opened 2026-04-30
compiler: client template references module-level imports (e.g. nodeTypes) it can't reach


kfly8:
## Bug

The standalone client `template: _p => \`…\`` function generated by the JSX compiler can reference module-level imports it doesn't have access to, throwing `ReferenceError` at hydration / `render()` time.

## Repro

\`\`\`tsx
'use client'

import { nodeTypes } from './nodes'  // module-level import: a Record<string, function>
import { Flow } from '../ui/xyflow'

interface Props { /* … */ }

export function DeskCanvas(props: Props) {
  return (
    <Flow nodeTypes={nodeTypes} nodes={[]} edges={[]} />
  )
}
\`\`\`

Generated output:

\`\`\`js
hydrate('DeskCanvas', {
  init: initDeskCanvas,
  template: (_p) => \`…\${renderChild('Flow', { nodes: [], edges: [], nodeTypes /* ❌ */, … })}…\`
})
\`\`\`

The `nodeTypes` reference inside the template tag's interpolation has no closure access. When `render(container, 'DeskCanvas', props)` runs the template at CSR mount, the browser throws `ReferenceError: nodeTypes is not defined`.

(For DeskCanvas, `nodeTypes` is a `Record<string, (this: HTMLElement, props) => void>` — a non-serializable object. So even if the reference resolved, materialising it in an HTML string would be wrong; the static template branch shouldn't have been taken at all.)

## Expected

Either:
- `canGenerateStaticTemplate` rejects the JSX subtree when an attribute value references a module-level import that isn't a JSON-serializable constant — falling back to CSR via `renderChild`.
- Or the template captures imports it needs (closure / module scope), e.g. by emitting `const __template_nodeTypes = nodeTypes; ` before the `hydrate(...)` call and rewriting the template to use `__template_nodeTypes`.

## Context

Found while debugging desk page hydration after migrating `DeskCanvas` to JSX-native xyflow (`piconic-ai/desk#42`). PR #1127 fixes the destructured-from-props-object class of bug; this issue tracks the broader template-closure problem.

PR #2419, opened 2026-07-29
Investigation: conditional-branch text is interpolated unescaped (no fix yet — naive fix breaks composition)

kfly8:
**This PR ships no change.** `git diff origin/main` is empty. It exists to record a real bug, a located cause, and — the part worth keeping — measured evidence that the obvious fix is wrong.

Found while investigating whether text-only ternaries could stop being lowered as conditionals (a lazy-row-gate question). Kept separate from that perf work on purpose.

## The bug (confirmed)

A reactive conditional whose branch is TEXT interpolates the value into the branch template **unescaped**. The same value written so that it lowers to a text slot **is** escaped. Both shapes in one CSR render, with `row.a = 'A&<b>'`:

```html
<span class="tern"><!--bf-cond-start:s0-->A&<b><!--bf-cond-end:s0--></span>
<span class="text"><!--bf:s2-->A&amp;&lt;b&gt;<!--/--></span>
```

The first is `{sel() === row.id ? row.a : row.b}`, the second the same expression wrapped in `String(...)`. The branch template becomes markup via `innerHTML` / `parseHTML`, so `<b>` there is a real element.

## The cause (located)

One asymmetry in all four template renderers' `'expression'` case (`packages/jsx/src/ir-to-client-js/html-template.ts` — `irToHtmlTemplate`, `irToPlaceholderTemplate`, `irToComponentTemplateWithOpts`, `generateCsrTemplateWithOpts`):

```ts
if (node.slotId) {
  return `<!--bf:${node.slotId}-->${escapeTextSlotExpr(...)}<!--/-->`   // escaped
}
return `${value}`                                                       // NOT escaped
```

A conditional **branch's** inner text expression carries no `slotId` — the conditional owns the id, not the expression — so it always lands on the un-slotted path.

## Why the obvious fix is wrong — measured

I applied the slotted form's escaping decision to the un-slotted path in all four renderers, on the premise that a conditional branch's text was the only thing reaching it. `bun test packages/jsx` (2803 tests) against that change: **3 failures plus 15 failing snapshots (26 snapshot lines).** Every one is the same mistake, and together they enumerate what the path actually carries:

| What reached the path | Evidence | Correct handling |
|---|---|---|
| `children` passthrough | `escapeText(_p.children)` ×8 in `doc-examples.test.ts.snap` | **Never escape** — it is the rendered markup of a child component. Escaping breaks composition outright |
| flatMap / array-builder leaf HTML | `escapeText(escapeText((c)))`, `map-arbitrary-body.test.ts` — a test whose name is *"leaf text interpolations escape like the SSR JSX runtime (parity)"* | Already escaped upstream; a second layer is a parity break |
| `#1128` empty placeholder | `${''}` → `${escapeText('')}`, `template-closure.test.ts` | Must stay a bare empty placeholder |
| const-inlined literal | `${('x')}` → `${escapeText(('x'))}`, `csr-template-loop-shadowing.test.ts` | Inlining must survive verbatim |
| conditional branch inner text | the bug above | **This** is the one that needs escaping |

So **the un-slotted path is not an escaping decision point.** Five different kinds of content arrive there and four of them must not be touched. Only a caller that knows which kind it holds can decide.

Reverted rather than iterated in place, so the branch never carries a composition regression.

## What a correct fix needs

Escape where a conditional branch renders a **text** child, not for every un-slotted expression. Either thread the branch context into the four renderers' expression case, or render the branch's text child explicitly in the `'conditional'` case. Not a one-line symmetry — and the golden blast radius should then be limited to components with a text-branch conditional, instead of the 26 snapshot lines above.

## Still unverified

- **What each of the nine SSR adapters does with the same branch.** Hono passes the ternary to its JSX runtime, which escapes text children, so SSR looks safe *there* — which would make this an SSR/CSR **parity** violation on top of the injection. The other eight are unverified.
- A concrete end-to-end exploit path in a real app. What is demonstrated is the emitted template and the rendered DOM, not a full chain.

## State of this branch

Two revert commits plus CI's two auto-snapshot commits, which cancel out. Content is identical to `main`. Verified on the reverted branch:

- the three named test files — 34 pass, 0 fail
- `doc-examples` — 219 pass, 0 fail

Close this and open an issue, or keep it as the base for the narrow fix — either is fine. The reproduction and the enumerated negative result are the deliverable.


kfly8:
Closing — the real fix is in #2421.

The 0-file diff here is not a glitch, but it is confusing to leave open, which is fair. This branch is what remains after reverting a fix that was wrong: two revert commits plus two auto-snapshot commits from CI, which cancel out exactly. Its value was the negative result, and that is now folded into #2421's description and commit messages.

For the record, what this PR established:

- the bug is real (a text-branch conditional interpolates unescaped; `A&<b>` becomes a real `<b>` element that also swallows the `bf-cond-end` marker)
- the cause is one asymmetry in four renderers — the slotted form escapes, the un-slotted form did not, and a conditional branch's text owns no slot id
- **the obvious fix is wrong**: five kinds of content reach the un-slotted path, and `children` (a child component's rendered HTML) must not be escaped. Escaping it broke composition in 8 places

#2421 puts the decision in the IR instead, which is what that negative result pointed at.

---

Issue #1694, opened 2026-05-31
Client template does not HTML-escape interpolated text content (parallel to #1692 attribute fix)


kfly8:
## Summary

#1692 fixed HTML-escaping of interpolated **attribute values** in the client-render template (via `escapeAttr`). The parallel gap remains for **text content**: `${expr}` text interpolations are emitted raw into the template string, so a string child containing `<` / `&` / `>` diverges from the SSR-escaped bytes and is a markup-injection vector when the template is inserted via `innerHTML`.

## Repro

```tsx
'use client'
export function Label({ text }: { text: string }) {
  return <span>{text}</span>
}
```

Emitted client template (current):

```js
hydrate('Label', { init: initLabel, template: (_p) => `<span bf="s1"><!--bf:s0-->${_p.text}<!--/--></span>` })
```

With `text = 'a<b&c'`:
- **SSR (Hono)** renders `a&lt;b&amp;c`
- **CSR template** renders raw `a<b&c` → diverges, and `<b…>` is parsed as markup when inserted via `innerHTML`.

(The reactive update path is unaffected — it writes via `textContent`/`__bfText`, which is safe. This is specifically the template **string** path used for client/child render.)

## Why it wasn't fixed in #1692

#1692 was scoped to attribute values, where escaping is unconditional (an attribute value is always text). Text content is trickier: the same `${...}` slot may carry either a **string** (must escape `& < >`) or **already-assembled HTML** — e.g. `renderChild(...)` output or hoisted-children markup (`bf-s="__BF_PARENT_SCOPE__"…`) — which must **not** be escaped. So the fix needs the codegen to distinguish text-valued interpolations from HTML-valued ones and only escape the former (text context: `& < >`, not `"` / `'`).

## Scope / acceptance

- [ ] Escape string/text interpolations in the client template with a text-context escape (`& < >`), matching the SSR adapters.
- [ ] Do **not** escape `renderChild(...)` / hoisted-children / marker HTML interpolations.
- [ ] CSR conformance: a fixture with metacharacter-bearing text content matches SSR.
- [ ] Runtime/unit coverage for the text-escape helper (mirror `escape-attr.test.ts` from #1692).

## References

- #1692 — the attribute-value counterpart (`escapeAttr` runtime helper + `templateAttrExpr` codegen)
- `packages/jsx/src/ir-to-client-js/html-template.ts` — client template codegen (text-node emission)
[...]

PR #2421, opened 2026-07-29
fix(jsx): escape by default — decide escaping from IR (BLOCKED: classifier misses const chains)

kfly8:
> ## ⛔ Do not merge — two blockers
>
> **1. The `Slot` / asChild composition regresses.** The classifier misses a `children` value that reaches JSX through local consts, so a grandchild's markup gets escaped:
>
> ```jsx
> const childProps = children.props
> const childChildren = childProps.children
> return <Tag {...childProps} {...props} className={...}>{childChildren}</Tag>
> ```
>
> The child expression is the identifier `childChildren`, not `children`, so `childExpressionContentKind` returns text. After const inlining the template emits `${escapeText((_p.children.props).children)}` — the grandchild's rendered HTML, escaped. Fixing this needs Phase 1 to follow the local const chain (`childChildren` ← `childProps.children` ← `children.props`) back to the `children` prop; the analyzer already tracks constants for the lazy-row gate, so the material is probably there.
>
> **2. CI's fixture auto-regeneration silently absorbed that regression.** `update-fixtures` pushed `d2772cd6 Update fixture expectedHtml (auto-generated)`, which wrote `escapeText((_p.children.props).children)` into `slot.client.js` as the new expected output. **The suites can therefore go green on a real composition break.** I only caught it by reading the auto-generated diff. That hazard is worth addressing on its own terms, independently of this branch — a golden that regenerates itself cannot also be the thing that catches an escaping regression.
>
> What I did verify in that diff: `escapeText(renderChild` appears **0 times**, so `renderChild()` results are not escaped, and the ternary's `: ${_p.children}` arm stayed raw. Everything else newly wrapped is genuinely text — string literals, numbers, `.length`, `t.name`, `tag`, `prefix`, `stateLabel`.

The design below is unchanged and I believe it is right; the classifier is what is incomplete.

---

## The bug

With `row.a = 'A&<b>'`, a text-branch conditional emitted the value raw:

```jsx
<td>{sel() === row.id ? row.a : row.b}</td>
```
```html
<td bf="s1"><!--bf-cond-start:s0-->A&<b><!--bf-cond-end:s0--></td>
```

The HTML parser turns that into a real element that also **swallows the range marker** (verified with happy-dom):

```html
<td bf="s1"><!--bf-cond-start:s0-->A&amp;<b><!--bf-cond-end:s0--></b></td>
```

- `textContent` is `"A&"`, not `"A&<b>"` — the intended text is lost
- `bf-cond-end` is now inside the injected element, so **data can relocate the boundary `insert()` swaps branches against**

The same expression wrapped in `String(...)` was escaped, because that shape lowers to a text slot. Escaping depended on how you wrote it.

## The cause

One asymmetry repeated in all four template renderers: the **slotted** form escaped, the un-slotted form did not. A conditional branch's inner text carries no `slotId` — the conditional owns it — so it always landed on the un-slotted path. **Position and content kind were conflated.**

## The design

`IRExpression.contentKind?: 'text' | 'markup'`, decided in Phase 1; the renderers branch on it. `slotId` now decides only *where* a value goes.

**Absent means text — escape.** An unclassified expression fails toward a visibly wrong render (`&lt;b&gt;` on screen) rather than toward injection. It also puts the client emission in agreement with every SSR backend this project targets, each escaping by default with an explicit raw opt-out:

| Adapter | Escape default | Raw opt-out |
|---|---|---|
| hono | JSX runtime | `raw(...)` |
| jinja / rust | `autoescape` | `\| safe` |
| twig | `autoescape` | `\| raw` |
| blade | `{{ }}` | `{!! !!}` |
| go-template | auto-escaper | `template.HTML(...)` |
| xslate | `autoescape` | `\| mark_raw` |
| mojolicious | auto-escape | `Mojo::ByteStream` |
| erb | adapter emits it explicitly (stdlib ERB does not auto-escape) | `raw(...)` |

Blocker 1 is precisely where "absent means text" fails *unsafely* in the other direction — not injection, but a broken render — because the value really is markup and the compiler did not know.

`'markup'` is set where the value is HTML: a `children` passthrough (`{props.children}`, `{props.children ?? ''}`, `{props.children || fallback}`, `{cond ? props.children : null}` — closed traversal set), and `joinArrayChild`. `String(props.children)` and `'x' + props.children` are deliberately not markup: they stringify it.

## Two omissions the suites caught

- **`escapeText(escapeText((c)))`.** `escapeLeafTextExpressions` pre-wrapped un-slotted preamble leaves precisely because the emitters left them alone. With the emitters deciding from IR it was a second wrapper, so the pass is **deleted**. Four tests pinned its `escapeText((x))` spelling and now pin `escapeText(x)` plus "`escapeText(escapeText(` never appears".
- **`escapeText(__bfSlot('OFF', __slots))`.** `branchSlotsVar` gated only the slotted arm; `__bfSlot` returns raw `<!--bf-slot:N-->` markers for live `Node` values (#1694). Now gates both.

## Verification status

Stated only where measured.

| | Result |
|---|---|
| `bun test packages/adapter-tests` (17 files, incl. CSR conformance) | 1451 pass, 0 fail |
| CSR conformance alone | 231 pass, 0 fail |
| `flatmap-segments` alone | 11 pass, 0 fail |
| `bun run lint` | clean |
| `bun test packages/jsx` | re-running — an earlier claim of "0 fail" was a number I did not measure (that run reported 5 failures; output was tail-truncated and CI's jsx-2 shard found one of them) |
| site/ui e2e | running |
| root `test:e2e` | not runnable in this container — `chrome-headless-shell` is absent while only `chromium` is installed; the uvicorn server starts fine, so all 104 tests fail at browser launch. Unrelated to this change |

Changeset is `minor` — this changes rendered output.


kfly8:
## ⛔ Blocker 3 — site/ui e2e regresses, 23 tests

Ran `site/ui`'s own e2e (the suite with a known baseline in this container; the root `test:e2e` cannot run here, see the PR body).

| | Baseline | This branch |
|---|---|---|
| Known reds | 3 — `form-builder` ×2, `studio` ×1 | — |
| Failing here | — | **23** — `xyflow.spec.ts` ×22, `validation.spec.ts` ×1 |
| Skipped | 20 | 20 |

These are **different tests** from the baseline's three known reds, so this is not the pre-existing set.

Why I read it as a real regression rather than noise: `xyflow` is the `renderNode` JSX-callback page, and among the failures is *"pill bodies hydrate without [object HTMLDivElement]"* — that is the preamble / flatMap-leaf machinery whose escaping this branch changed, including the `escapeLeafTextExpressions` removal. A JSX callback returning an element that now gets escaped instead of spliced would produce exactly this shape of failure.

Not yet done: confirming against `main` in the same container, and reading the actual assertion errors (my capture was tail-truncated to the failure list). Both are needed before saying anything more definite.

Running total of what needs resolving before this can merge:

1. `Slot` / asChild — the classifier misses `children` reaching JSX through local consts
2. CI's fixture auto-regeneration wrote that regression in as the expected output, so suites can go green on a composition break
3. This — 23 site/ui e2e failures concentrated in the JSX-callback path

(1) and (3) may well share a cause: both are markup values the classifier does not recognise as markup. If so, following const chains fixes both. That is a hypothesis, not a finding.

---

Issue #1211, opened 2026-05-08
compiler: JSX inside arrow-function callbacks emitted unchanged, crashes client bundle


kfly8:
## Symptom

JSX nested inside an arrow-function callback expression survives compilation as raw JSX in the emitted client bundle, so the `.js` file fails to parse with `SyntaxError: Unexpected token '<'`. This kills hydration for **every** other component in the same bundle — not just the one with the callback.

## Reproduction

```tsx
"use client"
import { Flow } from "@/components/ui/xyflow"

export function MyDemo() {
  return (
    <Flow
      nodes={...}
      edges={...}
      renderNode={(n) => (
        <div>{n.data.label}</div>   // <-- this JSX is left raw in the bundle
      )}
    />
  )
}
```

The compiled client JS contains literally:

```js
get renderNode() { return (n) => (
  <div>{(n.data).label}</div>
) }
```

## Why it matters

- `<Flow renderNode>` is the documented escape hatch for custom node bodies in `@barefootjs/xyflow`, so users following the docs hit this immediately.
- Workaround we shipped in `site/ui` is to return an HTML *string* from the callback instead of JSX (see `site/ui/components/xyflow-demo.tsx`'s `pillBodyHTML`). That avoids the parse error but loses signal-driven reactivity inside the body.
- Acknowledged in `site/ui/components/xyflow-intro-demo.tsx`'s top comment as a known limitation, but no tracking issue.

## Suggested fix

Recursively walk callback bodies during the JSX→IR pass and compile JSX expressions there too, the same way it does for the top-level return.

## Workarounds in tree

- Return an HTML string from `renderNode` (see `pillBodyHTML` / `fanBodyHTML` in `site/ui/components/xyflow-demo.tsx`).
- Keep callback-returning-JSX components in their own file so the parse failure only takes that bundle down.

Issue #1250, opened 2026-05-12
Adapter contract: shared ParsedExpr emit, module sections, cross-adapter conformance


kfly8:
## Problem

Three interconnected adapter-layer issues weaken the compiler contract:

1. **`ParsedExpr` conversion is duplicated per adapter.** Every adapter re-implements traversal of the same expression AST (`identifier`, `member`, `call`, `binary`, `logical`, `higher-order`, `template-literal`, …) to its target language.
2. **Module-structure responsibilities leak from adapters back into the compiler.** `applyExportKeyword` post-processes adapter output with regex; `rewriteTemplateImports` patches imports the adapter was supposed to have emitted correctly.
3. **No cross-adapter conformance test** verifies that the same IR produces equivalent hydration-marker shape and identifiers across adapters.

Together, these create silent divergence: a new expression kind missed by one adapter, a module-structure regex that breaks under unrelated output drift, and marker drift that is only caught when a user compiles the same code through a second adapter.

### `ParsedExpr` duplication
- `packages/jsx/src/adapters/interface.ts:141` — `renderExpression` interface entry; each adapter must implement.
- `packages/adapter-go-template/src/adapter/go-template-adapter.ts:1465–1650` — full `ParsedExpr` switch (`identifier`, `literal`, `call`, `member`, `binary`, `unary`, `logical`, `higher-order`, `template-literal`), ~185 lines.
- `packages/adapter-mojolicious/src/adapter/mojo-adapter.ts:539–920` — `renderPerlFilterExpr` + `renderParsedExprToPerl` + `convertExpressionToPerl`, the same switch structure spread across ~380 lines.

### Module-structure leakage
- `packages/jsx/src/compiler.ts:93–97` — `applyExportKeyword` runs `replace(/^function /, 'export function ')` on adapter output. Assumes the first `function ` token is the component.
- `packages/jsx/src/compiler.ts:44–83` — `rewriteTemplateImports` filters/rewrites adapter-supplied `ImportInfo[]`. Adapters were supposed to provide the right set; the compiler patches what's missing.

### Conformance gap
- `packages/adapter-tests/src/run-adapter-conformance.ts` — wires per-adapter suites; no multi-adapter comparison.
- No test compiles the same IR through Hono + Go + Mojolicious and asserts equivalent marker IDs / structure.

## Why this produces defects

[...]

PR #2425, opened 2026-07-29
fix(jsx): keyed loop row's bare-expression conditional branch never updates on value change

kfly8:
## The bug

A keyed `.map()` row whose reactive conditional's branch is a **bare expression** (no wrapping element — e.g. `row.done ? row.label : 'pending'`) never updates when the item's value changes but the branch's own condition does not flip.

Measured in a real DOM (happy-dom), same key `id`, changing only the displayed value:

| row content | initial | value changed | condition flipped |
|---|---|---|---|
| `{row.done ? row.label : 'pending'}` | AAA | **AAA (stale — the bug)** | pending |
| `{String(row.done ? row.label : 'pending')}` | AAA | BBB | pending |
| `{row.label}` | AAA | BBB | BBB |

## Pinned cause (two halves)

1. `insert()` (`packages/client/src/runtime/insert.ts:303`) correctly no-ops when its condition is unchanged — branch-internal updates are deliberately the effect system's job, not DOM replacement's. **Untouched by this PR.**
2. `summarizeLoopChildBranch` (`packages/jsx/src/ir-to-client-js/collect-elements.ts`) collected **no** `reactiveTexts` for *any* branch whose entire content was a bare expression, so the effect `insert()` needed was never emitted — the value was baked into the branch's initial template string once, at row-creation time, and never touched again.

## The stale rationale

The skip's own comment blamed "the loop-child arm's `$t()`-based anchor lookup" for why an extra effect could go wrong. `$t()` no longer exists — it was one of four content mechanisms removed by the slot-unification work. The current claim door (`stringifyLoopChildArm` → `lazySlots(..., kind: 'markup')` → `writeMarkup`, `packages/client/src/runtime/claim-slots.ts`) clears the claimed range and splices a Node in by identity, so "a second instance lands beside the first" cannot happen through it anymore.

**The skip is narrowed, not removed**, because the *other* half of the rationale still holds: a branch that is a `CallExpression` may return a live DOM Node (a hoisted `renderNode={(n) => <Pill/>}` callback lowered to a component call, #1211/#1213), and re-invoking a non-idempotent call inside an *additional* nested effect discards the previous element's listeners/state on every unrelated tick.

## The fix

- `summarizeLoopChildBranch` now skips only when `node.hasFunctionCalls` is set — an AST-computed flag (`exprHasFunctionCalls`) that recursively walks the whole expression, catching a call nested inside a template literal or a sub-ternary, not just a top-level one. A property access, identifier, literal, template literal, string concatenation, or nested ternary of those cannot itself construct a DOM node, and is now collected.
- `transformConditionalBranch` (`packages/jsx/src/jsx-to-ir.ts`) now also allocates a `slotId` when the branch's free references include a `render-item` binding — read from the `freeRefs` the function already computes, not a new parse or a regex re-scan of the expression text (unlike the sibling `transformConditional`/`transformLogicalAnd` condition-side `referencesLoopParam` helper, which token-matches expression *text* and can false-match inside an unrelated string-literal branch). Without this, a bare loop-item-reading branch had no `slotId` at all, so there was nothing for the new effect — or the `<!--bf:sN-->` marker `irToHtmlTemplate` needs to emit for it — to attach to.

## SSR/CSR marker parity

`irToHtmlTemplate` renders both the marked-template SSR output (Go/ERB/Blade/Jinja/Mojolicious/Rust-minijinja/Twig) and the CSR/hydration template from the same node, so the two cannot disagree on shape. Confirmed with a new adapter conformance fixture (`loop-item-ternary-bare-branch`), passing against ERB/Jinja/Mojolicious/Rust-minijinja (Go/Blade/Twig skip actual rendering in this environment — toolchains unavailable — but compile identically).

The **Hono** (JSX-runtime) adapter renders conditional branches through a separate path (`renderNodeRawCtx`/`wrapWithCondMarker` in `hono-adapter.ts`) that omits a per-branch `slotId` marker regardless of this fix — confirmed this is **pre-existing** (already true before this change for a call-bearing branch like `_p.renderCell(id)`), and Hono is intentionally excluded from the marker-based `runAdapterConformanceTests` suite. A Hono-server-rendered (not client-created) row hitting this exact shape stays stale until its first condition flip, at which point `insert()`'s branch swap replaces the DOM with the correctly-marked template fragment and self-heals — not worsened by this change, but not fixed either. Documented in the changeset as a known follow-up.

## Tests

- `packages/client/__tests__/runtime/loop-branch-stale-text.test.ts` — compiles the real shape and drives the real runtime: same-key value change now updates; condition flip still works; the call-bearing shape keeps today's (unfixed) stale behavior.
- `packages/jsx/src/__tests__/loop-branch-bare-expression-reactive-text.test.ts` — emission pin: the safe shape emits the branch-text effect, the call-bearing shape (including a call nested inside a template literal) doesn't.
- `packages/adapter-tests/fixtures/loop-item-ternary-bare-branch.ts` — HTML output fixture pinning the new inner `<!--bf:sN-->` marker.

Hunted for breaking shapes: both arms plain strings, a nested-element arm, a nested-conditional arm, a call-bearing arm, a branch mixing an outer signal call with an item read, same-key replace/reorder/append — all behave as expected.

Reverted the fix and confirmed all three new tests fail for the right reason, then restored and confirmed they pass.

## Verification

| Suite | Result |
|---|---|
| `bun test packages/client` | 605 pass, 0 fail |
| `bun test packages/jsx` | 2780 pass, 0 fail |
| `bun test packages/adapter-tests` | 1456 pass, 0 fail |
| `tsc --noEmit -p packages/client` | clean |
| `tsc --noEmit -p packages/jsx` | clean |

`packages/adapter-tests/coverage-map.json` changed only additively (the new fixture's entry plus matching +1 increments in the summary counts for its construct kinds) — read line by line, nothing else moved.

---

### DIFF
--- a/.changeset/hono-cond-branch-slot-marker.md
+++ b/.changeset/hono-cond-branch-slot-marker.md
@@ -1,0 +1,46 @@
+---
+"@barefootjs/hono": patch
+---
+
+Fix the Hono adapter dropping a conditional branch's own `bf:sN` slot
+marker, the only renderer among erb/jinja/go-template/mojolicious/hono to do
+so.
+
+`loop-branch-stale-text` (previous changeset) gave a keyed `.map()` row's
+bare-expression conditional branch (`task.done ? task.label : 'pending'`)
+its own `slotId` so a per-item update effect can rewrite it without the
+outer conditional re-evaluating. `irToHtmlTemplate` renders that marker
+identically into the CSR/hydration template and every marked-template
+adapter (erb: `bf.text_start("s1")`…`bf.text_end`; jinja/go-template:
+`bfTextStart "s1"`…`bfTextEnd`; mojolicious: `bf->text_start("s1")`…
+`bf->text_end`). The Hono adapter renders conditional branches through a
+separate path (`renderNodeRawCtx`/`wrapWithCondMarker` in
+`hono-adapter.ts`) that bypassed `renderExpression` entirely for expression
+nodes, so it never looked at the branch's `slotId` and emitted no marker at
+all — just `{bfComment("cond-start:s0")}{task.label}{bfComment("cond-end:s0")}`.
+
+Measured consequence: claiming that Hono-shaped branch HTML with
+`lazySlots(branchScope, [{ id: 's1', kind: 'markup', path: [] }])` and
+writing to it warned twice (`slot s1 marker not found; skipping`, `no
+claimed slot for id s1; write ignored`) and left the DOM unchanged. A
+Hono-server-rendered row hitting this exact shape stayed stale until its
+first condition flip.
+
+Fix: `wrapWithCondMarker`'s expression-node branch now checks the node's
+own `slotId` and, when present, wraps the content in `{bfText(id)}`…
+`{bfTextEnd()}` INSIDE the outer `cond-start`/`cond-end` pair — matching
+every other adapter's marker structure byte-for-byte. A branch with no
+`slotId` (a plain string literal, e.g. `conditional-wrapping-loop`'s
+`'[x]'`/`'[ ]'`) gets no inner marker, unchanged. The existing
+`bfText`/`bfComment`/`bfTextEnd` utility-import detection already scans the
+generated code by identifier, so no import-list change was needed.
+
+After the fix, claiming the same branch HTML updates the DOM
+(`<!--bf:s1-->CCC<!--/-->`) with no warnings.
+
+`loop-item-ternary-bare-branch`'s `expectedHtml` (adapter conformance
+fixture) gains the inner `<!--bf:s1-->Write it<!--/-->` marker inside the
+outer `bf-cond-start:s0`/`bf-cond-end:s0` pair — the only byte change
+anywhere in the SSR/CSR conformance suites, since the earlier PR's
+CI-generated regeneration had (wrongly) captured Hono's marker-omitting
+output as the expected shape.
--- a/.changeset/loop-branch-stale-text.md
+++ b/.changeset/loop-branch-stale-text.md
@@ -1,0 +1,74 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix a keyed `.map()` row whose reactive conditional branch is a bare
+expression (e.g. `row.done ? row.label : 'pending'`) never updating the
+branch text when the item's value changes without its condition flipping.
+
+Two gates used to conspire to freeze the value forever: `insert()`
+(`@barefootjs/client` runtime, unchanged by this fix) correctly no-ops when
+its condition is unchanged — branch-internal updates are the effect
+system's job, not DOM replacement's — but `summarizeLoopChildBranch`
+(`packages/jsx/src/ir-to-client-js/collect-elements.ts`) collected NO
+`reactiveTexts` for ANY bare-expression branch, so the effect `insert()`
+was relying on was never emitted. The value was baked into the branch's
+initial template string once, at row-creation time, and never touched
+again.
+
+The skip existed to protect a different, still-real shape: a branch that is
+a `CallExpression` which may return a live DOM Node (a hoisted
+`renderNode={(n) => <Pill/>}` callback lowered to a component call,
+#1211/#1213). Re-invoking such a call inside an *additional* nested effect
+calls it again on every unrelated tick, discarding the previous element's
+listeners/state. That risk is real, but the skip's own rationale comment
+attributed the failure mode to "the loop-child arm's `$t()`-based anchor
+lookup" — `$t()` no longer exists; it was one of four content mechanisms
+removed by the slot-unification work. The current claim door
+(`stringifyLoopChildArm` → `lazySlots(..., kind: 'markup')` → `writeMarkup`,
+`packages/client/src/runtime/claim-slots.ts`) clears the claimed range and
+splices a Node in by identity, so "a second instance lands beside the
+first" cannot happen through it — only the call's non-idempotence still
+applies, which is why the skip is narrowed rather than removed.
+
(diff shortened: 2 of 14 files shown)


kfly8:
## Blocked — SSR/CSR marker divergence, and it is Hono that diverges

Not pushing a fix for this one: it changes Hono's SSR output bytes for every conditional branch carrying a slot, which is a product decision rather than a cleanup. Recording the measurements so the decision can be made from facts.

### What the change does today

Allocating a `slotId` for a bare loop-item branch is correct for the client — that slot is what the new update effect claims. But only some renderers emit the marker for it:

| renderer | emits `<!--bf:s1-->` … `<!--/-->` inside the branch |
|---|---|
| CSR / hydration template | **yes** |
| ERB · Jinja · Mojolicious · Rust-minijinja | **yes** (4 pass) |
| **Hono (JSX runtime)** | **no** |

Hono renders branches through `renderNodeRawCtx` / `wrapWithCondMarker` (`hono-adapter.ts:775-845`), which emits only the `cond-start` / `cond-end` pair.

### Consequence for a Hono-server-rendered row

The claim fails, so the row is *not* fixed — and it now warns. Measured directly against Hono-shaped branch HTML:

```
DOM after writes: <!--bf-cond-start:s1-->AAA<!--bf-cond-end:s1-->   ← unchanged
  [barefootjs] slot s2 marker not found; skipping
  [barefootjs] no claimed slot for id s2; write ignored
  [barefootjs] no claimed slot for id s2; write ignored
```

Before this change there was no effect and no claim, so no warning. So for the SSR path this is a regression in noise with no gain — one claim warning plus one per write per tick, per row. The PR body's "pre-existing, unworsened" framing does not hold, and `[hono] marker conformance (#1250 phase 3)` is a real failing test, not an excluded suite.

### Why the auto-regenerated expectation hid it

`expectedHtml` is generated from Hono, so the auto-fix commit (`e812a870`) rewrote the fixture that was added to pin the marker into one that asserts its **absence**:

```diff
- <li data-key="write"><!--bf-cond-start:s0--><!--bf:s1-->Write it<!--/--><!--bf-cond-end:s0--></li>
+ <li data-key="write"><!--bf-cond-start:s0-->Write it<!--bf-cond-end:s0--></li>
```

That propagated Hono's shape into the shared expectation and broke `CSR Conformance Tests` for the same fixture. A golden that regenerates itself cannot also be the detector for the thing it covers.

### The option space has collapsed to one

Making the client tolerant of a missing inner marker does **not** work: the CSR template still emits the marker, so CSR conformance stays red against a Hono-shaped expectation. The only shape where all four renderers agree is Hono emitting the marker like everyone else. The primitives already exist — `bfText(slotId)` → `<!--bf:sN-->` and `bfTextEnd()` → `<!--/-->` in `packages/adapter-hono/src/utils.ts`.

Scope to weigh before doing it: that touches every conditional branch with a slot, not only this shape, so Hono SSR bytes move more broadly and fixtures regenerate. Worth doing deliberately rather than as a side effect of this PR.

### Unrelated, already handled

`lockfile` was failing because the new fixture landed without regenerating `ui/support-matrix.lock.json`. Fixed in `03c9ed4a`; the regenerated diff is 152 counter pairs, every one exactly +1, nothing else moved — read rather than trusted.

---

kfly8:
## Unblocked — Hono now emits the branch marker

The divergence recorded above is fixed in `bb32b16b`. `wrapWithCondMarker` had a dedicated `node.type === 'expression'` branch that never consulted the node's own `slotId`; it now nests `{bfText(id)}` … `{bfTextEnd()}` inside the outer cond pair when the branch carries one, and emits unchanged when it does not.

Verified independently of the implementation, not taken on report:

**SSR and CSR now agree byte-for-byte in marker structure** — compiled the same component through `HonoAdapter` and read the output:

```
Hono SSR : <!--bf-cond-start:s1--><!--bf:s2-->AAA<!--/--><!--bf-cond-end:s1-->
CSR row  : <!--bf-cond-start:s1--><!--bf:s2-->AAA<!--/--><!--bf-cond-end:s1-->
```

The `else` branch (`'pending'`, a string literal with no `slotId`) correctly gains no marker on either side. The emitted Hono template's import list picks up the helpers: `import { bfComment, bfText, bfTextEnd } from '@barefootjs/hono/utils'`.

**The SSR-path claim now succeeds** — same probe as the blocker comment above, before vs after the adapter change:

| branch HTML | DOM after write | warnings |
|---|---|---|
| `<!--bf-cond-start:s1-->AAA<!--bf-cond-end:s1-->` | unchanged | 2 |
| `<!--bf-cond-start:s1--><!--bf:s2-->AAA<!--/--><!--bf-cond-end:s1-->` | `…<!--bf:s2-->BBB<!--/-->…` | **0** |

**Suites** (re-run locally, not relayed): `packages/adapter-hono` **1269 pass / 0 fail** (was 1268 / 1 — the marker conformance failure); the `loop-item-ternary-bare-branch` fixture across adapters **5 pass / 0 fail** (was 4 / 1, the CSR conformance failure). `ui/support-matrix.lock.json` needs no further regeneration — `committed lock matches a fresh computeSupportMatrix()` passes standalone.

### The one remaining red is not from this PR

`lockfile` fails on `determinism > two engine runs over the same inputs produce deep-equal and byte-equal reports`. That test exceeds bun's 5s per-test timeout under load. Measured:

| ref | isolated runtime |
|---|---|
| `origin/main` | **6.02 s** |
| this branch | 7.75 s |

It is already over the limit on the base branch, so it flakes independently of this change — nothing here touches the compat engine. Not fixing it inside this PR to avoid mixing an unrelated test-timeout change into a review about marker parity; a re-run should clear it, and giving that test an explicit timeout is worth doing separately.

---

PR #2430, opened 2026-07-30
perf(jsx): build a lazy row's content door on first use, not at claim time

kfly8:
Closes the second half of `spec/slot-unification.md` §8's "claim-mechanism allocation shape" follow-up. The first half (hoisting the per-row `ClaimPlan` literal) landed in #2429.

## The waste

A lazy-eligible loop's adopted-row claim resolved the element refs **and** constructed the content door in one tuple:

```js
const __lzc_l0 = (__e) => {
  const __el = __e.primaryEl
  return [qsa(__el, '[bf="s3"]'), lazySlots(__el, __lzs_l0)]
}                                 // ^ one closure per row, often never used
```

That tuple is built for an entry the first time either apply path touches the row, and `applyOuter`'s seed pass touches **all** of them. So a loop whose only outer-involving binding is an attribute — the common `className={selected() === row.id ? 'danger' : ''}` shape, and the one the SSR bench renders — built a door for every row of the list at hydration and used none of them. Nothing writes content until that row's item actually changes.

## The change

The claim leaves the door's slot `null`; `applyItem`/`applyOuter` materialize it on demand, once per apply body and only when that body has content bindings to write:

```js
const __d = __r[1] ?? (__r[1] = lazySlots(__e.primaryEl, __lzs_l0))
```

`createRow` is deliberately untouched — it writes every text on the tick it clones the row, so its door is used immediately and deferring it would only add a branch. Its emitted output is byte-identical. The deferral is confined to server-rendered rows, which is where the allocation was wasted.

Reachability holds because a row whose refs `createRow` seeded always finds a door already in the slot and never evaluates the `??` right-hand side; only adopted rows see `null`, and the adopted claim plan is the only plan they can use.

**§2's claim-once rule is preserved.** That rule is about one door resolving the whole plan on first access, and each door already enforces it internally — `lazySlots`/`lazyClaimSlots` call `claimRefs` once and cache. Deferring *construction* moves when that single resolution happens rather than adding a second resolution path. A loop with an outer-involving TEXT still claims every row at seed, because read-compare-write cannot compare what it has not resolved; the deferral is a no-op there by design.

## Measurement

SSR post-hydration heap (`benchmarks/ssr/bench-ssr-memory.ts` — 1,000 rows, item texts plus an outer-signal class):

| | before | after (run 1) | after (run 2) |
|---|---|---|---|
| barefoot | 1630.6KB | **1573.2KB** | **1572.9KB** |
| react (control) | 3474.8KB | 3474.8KB | 3475.8KB |
| solid (control) | 2578.2KB | 2578.0KB | 2578.0KB |

**-57.4KB (-3.5%)**, reproduced twice, against a per-run stdev of 0.1–0.6KB. SSR output does not change; no `expectedHtml` fixture moved.

### The update path does not regress

`applyItem` gains one nullish check per row per update, so `update10th` was checked directly. Quick mode (3 iterations) showed 12.10ms → 14.10ms with non-overlapping ranges, which looked like a regression; at full iteration counts (warmup 5, measure 10 — same machine, same session) it is not one:

| | median | mean | stdev | range | n |
|---|---|---|---|---|---|
| before | 13.15ms | 14.21ms | 2.12 | 11.9–18.3 | 10 |
| after | **12.65ms** | 12.80ms | 0.88 | 11.8–14.8 | 10 |

Both quick-mode figures fall inside the after-condition's own n=10 spread, so the apparent 2ms was three-iteration noise. That matches the mechanism: rows the DOM suite creates come from `createRow`, so their door is already in the slot and the `??` short-circuits.

The DOM suite's `memory` column is also unmoved — 971.9KB → 972.1KB locally, and 969.8KB on this PR's CI run against 970.5KB from #2429's. That is the expected result rather than a null one: the column creates its rows client-side through the path this change leaves eager.

## Two corrections to the §8 estimate

The spec entry is updated rather than deleted, because the wrong estimate is the reusable part.

- **The ~230KB/1k rows figure was too high.** It predated #2429's removal of the per-row plan objects, and `lazySlots` never scanned eagerly to begin with (`claim-slots.ts:617` defers to the first write) — so what remained was a closure per row, not a resolved slot map.
- **The DOM suite's memory column cannot show this at all.** That is where #2429 was measured, but it creates its 1,000 rows client-side through `createRow`, the path left eager here. The right bench is SSR post-hydration heap.

## Verification

New behavioural coverage: `packages/client/__tests__/runtime/lazy-row-adopted-door.test.ts` renders real Hono SSR markup, hydrates it, then changes an item. That sequence is the only one that executes the deferred branch — the emission tests read the generated text, the `createComponent` tests take the eager path, and `csr-conformance` / `ssr-hydration-contract` never run a post-hydration update at all. Verified to fail when the door is built against the wrong root, so it pins behaviour and not just spelling.

Green locally: `packages/jsx` in full (2742 pass, 0 fail) plus `typecheck:tests`, `adapter-tests` (1456), `packages/client` (605).

Regenerated expectations, both deliberate rather than auto-committed (#2428): the 6 `doc-examples` snapshots (all three lines of the same shape change) and `ai-chat.client.js`, the one shared fixture with a lazy loop carrying texts. The three drift-detection jobs `update-doc-snapshots` / `update-expected-html` / `update-meta` are green, so nothing regenerable was missed.

### DIFF
--- a/.changeset/lazy-row-defer-content-door.md
+++ b/.changeset/lazy-row-defer-content-door.md
@@ -28,6 +28,10 @@ clones the row, so its door is used immediately and deferring it would only
 add a branch. The deferral is therefore confined to server-rendered rows,
 which is where the allocation was wasted.
 
+With the door gone from the claim, the element refs are the only parts left
+that read the row root — so a row with no reactive ATTRIBUTE now claims to a
+bare `[null]` and no longer binds `__e.primaryEl` at all.
+
 **§2's claim-once rule is preserved.** That rule is about one door resolving
 the whole plan on first access, and each door already enforces it internally —
 `lazySlots`/`lazyClaimSlots` call `claimRefs` once and cache the result.
--- a/.changeset/lazy-row-defer-content-door.md
+++ b/.changeset/lazy-row-defer-content-door.md
@@ -1,0 +1,58 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Build a lazy loop row's content door on first use instead of at claim time,
+so an `applyOuter` that only drives attributes stops allocating one closure
+per row of the list.
+
+A lazy-eligible loop's adopted-row claim (`__lzc_<mid>`) resolved the element
+refs *and* constructed the content door — `lazySlots(__el, __lzs_<mid>)` — in
+one tuple. That tuple is built for every entry the first time either apply
+path touches the row, and `applyOuter`'s seed pass touches **all** of them.
+So a loop whose only outer-involving binding is an attribute (the common
+`className={selected() === row.id ? … : …}` shape) built a door per row at
+hydration and used none of them: nothing writes content until that row's item
+actually changes.
+
+The claim now leaves the door's slot `null` and `applyItem`/`applyOuter`
+materialize it on demand, once per apply body and only when that body has
+content bindings to write:
+
+```js
+const __d = __r[1] ?? (__r[1] = lazySlots(__e.primaryEl, __lzs_l0))
+```
+
+`createRow` is deliberately unchanged: it writes every text on the tick it
+clones the row, so its door is used immediately and deferring it would only
+add a branch. The deferral is therefore confined to server-rendered rows,
+which is where the allocation was wasted.
+
+**§2's claim-once rule is preserved.** That rule is about one door resolving
+the whole plan on first access, and each door already enforces it internally —
+`lazySlots`/`lazyClaimSlots` call `claimRefs` once and cache the result.
+Deferring *construction* moves when that single resolution happens; it does not
+add a second resolution path. A loop with an outer-involving TEXT still claims
+every row at seed, because read-compare-write cannot compare what it has not
+resolved, so the deferral is a no-op there by design.
+
+Measured on the SSR post-hydration heap bench
+(`benchmarks/ssr/bench-ssr-memory.ts` — 1,000 rows, item texts plus an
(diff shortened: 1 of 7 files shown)


kfly8:
Both review points were real and are fixed in bce34b5a.

**Dead `const __el = __e.primaryEl`** — correct, and it was reachable rather than hypothetical: the six `doc-examples` snapshots regenerated in the previous commit are exactly that shape (text-only rows, no reactive attribute, claiming to `[null]`). Deferring the door removed the last part that read the row root in that case.

The emit is now driven by `refParts`' output rather than a second predicate, so a future part that reads `elVar` keeps the binding without anyone remembering to update a condition:

```ts
const parts = refParts(lazyRow, '__el', null, adoptedPlanVar, true)
lines.push(`${indent}const ${claimVar} = (__e) => {`)
if (parts.some(p => p.includes('__el'))) lines.push(`${indent}  const __el = __e.primaryEl`)
```

`ai-chat.client.js` keeps its binding — that row has an attr ref — which is what makes the condition discriminating rather than always-off. Pinned from both sides: `a text-only adopted claim binds no row root` for the `[null]` case, and the existing `the adopted-row claim leaves the content door slot empty` for the attr-bearing one.

**Unremoved temp module** — also correct. `setup()` now removes its `mkdtempSync` directory in a `finally` once the import resolves, matching `lazy-row-node-content.test.ts`. The module is loaded by then, so deleting the file is safe.

`packages/jsx` 2743 pass / 0 fail, `packages/client` 608 pass / 0 fail, biome clean.

---

PR #2442, opened 2026-07-31
feat(router): reconcile `<head>` page metadata; document that head resources are not managed

### LINKED ISSUE #2438: router docs: say that <head> is not managed — route-scoped stylesheets belong inside the region
> **Reframed after review.** I originally filed this as a defect, arguing that reconciling `document.title` but nothing else in `<head>` looked like an oversight. That argument doesn't hold: the title is read for the route announcement (`announceNavigation` + the live region), so it's navigation correctness, not the start of a head-reconciliation feature. And a region-swapping router doing exactly what its name says is not a bug. **The bug was in my app** — I put route-scoped CSS in `<head>`, outside the region the router manages. What's left here is a feature request and a discoverability complaint, and I've kept the reproduction because the failure mode is worth knowing about.

## What happened

A route in my app ships its own stylesheet (a full-screen editor whose class names — `.head`, `.chip`, `.group`, `.drawer` — would collide with the rest of the site, so it must not be loaded globally). I emitted that `<link rel="stylesheet">` conditionally in `<head>`.

Soft-navigating **into** that route renders it completely unstyled; a reload fixes it. Soft-navigating **out** leaves the stylesheet linked, so it then applies to every route visited afterwards.

Both are the correct consequence of `startRouter` swapping `[bf-region]` and nothing else. My own renderer even carried the comment "head and scripts survive navigation" — I wrote down the contract and then violated it.

## Fix, for anyone who lands here

Put the route-scoped stylesheet **inside** the swapped region:

```tsx
<div bf-region="page">
  {isEditor ? <link rel="stylesheet" href="/editor.css" /> : null}
  {children}
</div>
```

`rel="stylesheet"` is body-ok per HTML, so this is valid. The sheet now enters and leaves with the region — no head handling, no per-link opt-out, and the route keeps soft navigation. This is the right placement under a region-swap contract, not a workaround.

## The request

**Documentation.** One line in the router docs — "head is not managed; put route-scoped stylesheets inside the region" — would have saved this entirely. The symptom (an unstyled page that a reload fixes) points at caching or the build, not at navigation, so it costs more to diagnose than it should.

**Optionally, an opt-in.** For apps that can't move the element (a sheet emitted by a framework layout, `<link rel="canonical">`, `<meta name="description">`), something like `startRouter({ head: 'stylesheets' })` or honouring a `bf-head` marker would help. Not asking for it to be the default.

If an opt-in ever happens, the ordering matters in both directions: add incoming sheets **before** the swap (or the new page paints unstyled), and remove outgoing ones **after** (or the page still on screen loses its styles for a frame). I built both variants before settling on the in-region placement, and hit each of those in turn.

## For reference: what does and doesn't carry over

Measured on a soft navigation between two routes (Chromium, router 0.28.1):

| head state | after navigating |
|---|---|
| `document.title` | updated |
| `<meta name="description">` | previous route's |
| `<meta property="og:*">` | previous route's |
| `<link rel="stylesheet">` | previous route's |

The `<meta>` rows are harmless in practice (crawlers don't soft-navigate), and are only listed so the scope is clear.

## Versions

`@barefootjs/router` 0.28.1 (same on 0.26.x / 0.27.0).

kfly8:
Closes #2438.

## What

`<head>` splits in two, and the split is the whole contract.

**Page metadata is now reconciled on every swap.** It always runs — there is no flag to disable it. Before, a swap wrote `document.title` and nothing else, and only because the route announcement reads it. So a soft navigation kept the previous route's `<meta name="description">`, `og:`/`twitter:` cards, and `<link rel="canonical">`. That is *invisible* wrongness: unlike the tab title you cannot see it in development, which is exactly the class the spec's "correct by default" rule exists for. It also isn't only a crawler concern — share sheets, in-app webviews, browser/extension UI, and RUM tools all read the live DOM.

**Head resources stay untouched, and that is a contract, not a gap.** A resource's lifetime isn't derivable from the incoming document: the shell, a `[data-bf-permanent]` node, a portal, or an island outliving the region may still need a sheet the next page's head omits, so "absent downstream" is no evidence of "no longer needed". Route-scoped CSS belongs *inside* the region, where both orderings are right by construction and no load is awaited in the navigation path.

```tsx
<Region>
  {isEditor ? <link rel="stylesheet" href="/editor.css" /> : null}
  {children}
</Region>
```

## The reconciled set

A **closed allowlist**, keyed by identity:

| head node | key |
| --- | --- |
| `<title>` | — (unchanged; also feeds the announcement) |
| `<meta name="description \| keywords \| robots \| author \| theme-color">` | `name` |
| `<meta property="og:*">` / `<meta name="twitter:*">` | `name` or `property` |
| `<link rel="canonical \| alternate \| prev \| next">` | `rel` + `hreflang`/`type`/`media` |

Present in both → replaced (skipped when the nodes are already equal, so metadata shared across routes causes no DOM churn); only incoming → added; only current → **removed**, so nothing leaks forward into later routes. Key attributes are folded (trim + lowercase) since `rel`, `hreflang`, `type` and `media` are all case-insensitive.

Anything outside the table is never read, replaced, or removed. That closedness is the deliberate divergence from Turbo's `provisionalElements`, which removes every untracked head element and can sweep away runtime-injected analytics or CSP nodes. A multi-token `rel="alternate stylesheet"` falls outside by construction — it's a resource. `data-bf-head="false"` opts a node out.

## Why this is cheap here

The incoming document is already parsed for the region diff, so this is a bounded node diff over data in hand: no route manifest, no payload, no fetch, no await, no layout effect. Next's App Router routes the same job through the RSC payload and has carried soft-navigation metadata bugs across many releases ([discussion #67663](https://github.com/vercel/next.js/discussions/67663)); Turbo does it document-diff style with no manifest at all. The document-diff position is the easy one.

## Why not head stylesheets

Both precedents converge on the removal half needing an author signal:

- **Turbo** adds new stylesheets and blocks the body swap on their load, but never removes one unless the author writes `data-turbo-track="dynamic"` — so it still leaks sheets forward by default.
- **Remix / React Router** removes them automatically, and has open issues where the outgoing route goes unstyled mid-transition ([#10160](https://github.com/remix-run/remix/issues/10160)) and where a `React.lazy` component loses its styles ([#9630](https://github.com/remix-run/remix/issues/9630)) — precisely the ordering hazard #2438 predicted.

In-region placement already *is* the author signal, and a better one: it needs no new API and gets both orderings right by construction. The spec records that if an opt-in ever ships it is a **per-element `data-bf-head` marker**, never a global `startRouter({ head: … })` option — scope is a property of the element, not of the app.

## Changes

- **`packages/router/src/head.ts`** (new) — `reconcileHead(incomingDoc)`: the allowlist, keying, and the replace/add/remove pass.
- **`packages/router/src/router.ts`** — call it beside the `document.title` write. Ordering-free, so it sits with the synchronous swaps rather than the awaited tail.
- **`packages/router/README.md`** — the section reframed as the two tiers, with the allowlist table and the in-region recipe.
- **`spec/router.md`** — a shipped **v2.5** phase for metadata (including why it is unconditional and why the allowlist is closed), and the resource half rewritten as a scoped limitation with the constrained opt-in shape.
- **`packages/router/__tests__/router.test.ts`** — 7 tests: replace/add/remove, allowlist non-interference (analytics, CSP, preconnect, alternate-stylesheet), `data-bf-head="false"`, key normalization, no-churn on equal nodes, plus the two resource-tier pins.
- **`integrations/hono/blog.tsx` + `e2e/blog.spec.ts`** — the router showcase now serves per-route `description` / `og:title` / `canonical`, with a Chromium e2e test walking index → post → index.
- **`.github/workflows/ci.yml`** — `packages/router` was in no workflow's path filter and no test group, so its tests had never run in CI. Added to both. Happy to split this out if you'd rather it landed separately.

## Testing

`bun test packages/router` — 53 pass, biome and `tsgo --noEmit` clean. `bunx playwright test e2e/blog.spec.ts` in `integrations/hono` — 6 pass in Chromium.

Both new layers verified non-vacuous by commenting out the `reconcileHead` call: the two unit reconciliation tests fail (the two non-interference tests keep passing, as expected), and the e2e test fails with the description still on the index's value. The key-normalization test fails against the un-normalized key and passes after.

Observed in the browser across a soft navigation, with the shell node marked to prove it wasn't a reload:

| | index | after → post | after ← index |
| --- | --- | --- | --- |
| `description` | The latest 10 posts… | Where this goes next — post 1 of 10… | The latest 10 posts… |
| `canonical` | `/blog` | `/blog/posts/where-this-goes` | `/blog` |
| `viewport` (outside allowlist) | unchanged | unchanged | unchanged |
| `meta[name=description]` count | 1 | 1 | 1 |

### DIFF
--- a/.changeset/router-head-metadata.md
+++ b/.changeset/router-head-metadata.md
@@ -11,7 +11,8 @@ route's `<meta name="description">`, `og:`/`twitter:` cards, and
 unlike the tab title.
 
 A swap now brings a **closed allowlist** of page metadata in line with the
-incoming document, with no option to turn on. Present in both → replaced
+incoming document. It always runs — there is no flag to disable it. Present in
+both → replaced
 (skipped when the nodes are already equal, so metadata shared across routes
 causes no DOM churn); only incoming → added; only current → removed, so nothing
 leaks forward into later routes.
--- a/.changeset/router-head-metadata.md
+++ b/.changeset/router-head-metadata.md
@@ -1,0 +1,33 @@
+---
+"@barefootjs/router": minor
+---
+
+Reconcile `<head>` page metadata across a soft navigation
+
+A swap used to write `document.title` and nothing else in `<head>` — and only
+because the route announcement reads it. So a soft navigation left the previous
+route's `<meta name="description">`, `og:`/`twitter:` cards, and
+`<link rel="canonical">` in place: wrongness you cannot see in development,
+unlike the tab title.
+
+A swap now brings a **closed allowlist** of page metadata in line with the
+incoming document, with no option to turn on. Present in both → replaced
+(skipped when the nodes are already equal, so metadata shared across routes
+causes no DOM churn); only incoming → added; only current → removed, so nothing
+leaks forward into later routes.
+
+| head node | key |
+| --- | --- |
+| `<meta name="description \| keywords \| robots \| author \| theme-color">` | `name` |
+| `<meta property="og:*">` / `<meta name="twitter:*">` | `name` or `property` |
+| `<link rel="canonical \| alternate \| prev \| next">` | `rel` + `hreflang`/`type`/`media` |
+
+Anything outside that table is never read, replaced, or removed — runtime-
+injected analytics tags, CSP `<meta http-equiv>`, `<link rel=preconnect>`, and a
+multi-token `rel="alternate stylesheet"` are all safe by construction. Opt a
+node out with `data-bf-head="false"`.
+
+Head **resources** (`<link rel="stylesheet">`, `<script>`, `<style>`) remain
+untouched: a resource's lifetime isn't derivable from the incoming document, so
+route-scoped CSS belongs inside the region, where it enters and leaves with the
+swap.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ on:
       - 'packages/chart/**'
       - 'packages/xyflow/**'
       - 'packages/adapter-tests/**'
+      - 'packages/router/**'
       # site/ui is the Hono adapter's end-to-end target, so Hono changes must
       # re-run this workflow's e2e-site-ui job (ci-hono.yml no longer duplicates it).
       - 'packages/adapter-hono/**'
@@ -33,6 +34,7 @@ on:
       - 'packages/chart/**'
       - 'packages/xyflow/**'
       - 'packages/adapter-tests/**'
+      - 'packages/router/**'
       # site/ui is the Hono adapter's end-to-end target, so Hono changes must
       # re-run this workflow's e2e-site-ui job (ci-hono.yml no longer duplicates it).
       - 'packages/adapter-hono/**'
@@ -72,7 +74,7 @@ jobs:
           - group: jsx-2
             run: bun test $(bun scripts/ci/shard-test-files.ts packages/jsx 2 2)
           - group: rest
-            run: bun test packages/client packages/test packages/adapter-tests
+            run: bun test packages/client packages/test packages/adapter-tests packages/router
 
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
--- a/integrations/hono/blog.tsx
+++ b/integrations/hono/blog.tsx
@@ -36,13 +36,23 @@ const importMap = JSON.stringify({
   },
 })
 
-const blogRenderer = jsxRenderer(({ children, title }) => {
+const blogRenderer = jsxRenderer(({ children, title, description, canonical }) => {
   return (
     <html lang="en" data-theme="dark">
       <head>
         <meta charset="UTF-8" />
         <meta name="viewport" content="width=device-width, initial-scale=1.0" />
         <title>{title ?? 'Barefoot Blog'}</title>
+        {/*
+          Per-route metadata, so a soft navigation has something to reconcile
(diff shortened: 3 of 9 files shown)


kfly8:
Addressed the review in 1eb3848.

**Wording (×3 — `head.ts`, README, changeset).** Fair — "no option to turn on" does read as "not available yet". All three now say it always runs and there is no flag to disable it.

**`headKey` normalization.** Taking the fix, correcting the stated consequence. `hreflang`/`type`/`media` are now folded (trim + lowercase) alongside `rel`, so `hreflang="en-US"` and `en-us` are one slot.

But duplicates do **not** accumulate without it, and I checked rather than assumed: a test written to the reported failure mode passes against the unfixed code. A missed key means the incoming node is appended *and* the stale one is removed by the `current`-not-in-`incoming` pass, so the count stays at one either way. What the missed key actually costs is churn and position — the slot is torn down and rebuilt at the end of `<head>` instead of replaced where it stood, which also defeats the `isEqualNode` no-churn skip. That is what the regression test pins (it fails against the unfixed code, passes after).

**Also in this commit**, from a separate request: the `integrations/hono` blog now serves per-route `description` / `og:title` / `canonical`, with an e2e test walking index → post → index in Chromium. It asserts all three follow while the shell node stays marked — so a full reload cannot satisfy it — and that `viewport` and the stylesheet, both outside the allowlist, stay untouched. Non-vacuity verified the same way: commenting out the `reconcileHead` call leaves the description on the index's value and the test fails.

---

Issue #2447, opened 2026-07-31
No DSL adapter lowers a .map() callback preamble's value declarations — the row template reads an unassigned variable


kfly8:
## Shape

A `.map()` callback with a block body whose preamble computes a value used in the row markup:

```tsx
'use client'
export function LoopPreambleAttrValue(props: { rows: Row[] }) {
  const [rows] = createSignal<Row[]>(props.rows)
  return (
    <ul>
      {rows().map(row => {
        const cls = row.done ? 'done' : 'open'
        return <li key={row.id} class={cls}>{row.label}</li>
      })}
    </ul>
  )
}
```

## Divergence

Hono renders it correctly. Every other adapter emits a template that **references a variable the template never assigns**, so the attribute comes out empty:

```
Expected: <li class="open" data-key="1">…write it…</li>
Received: <li class=""     data-key="1">…write it…</li>
```

The preamble's `const cls = …` is simply absent from the emitted template. What the adapters emit:

| Adapter | Emitted row |
|---|---|
| blade | `<li … class="{!! e($bf->string($cls)) !!}">` — `$cls` never set |
| twig | `<li … class="{{ bf.string(cls) }}">` — `cls` never set |
| xslate | `<li … class="<: $cls :>">` — `$cls` never set |
| go | `<li … class="{{$.Cls}}">` — a **parent-level** field, not per-row, and never populated |

Confirmed failing against real runtimes: **erb**, **jinja**, **mojolicious**, **rust** (minijinja) — all render `class=""`. Read off the emitted templates for **blade**, **twig**, **xslate**, **go-template**, whose runtimes were absent from the verifying environment (local Go is `1.24.7` against a harness needing 1.25+).

Go's variant is the same class of defect as #2445: the child value is hoisted to the parent scope instead of being built per row.

## This is pre-existing

Verified by stashing the compiler changes from the §9.5 lazy-preamble widening and re-running: ERB still renders `class=""`. The widening only changes which client-JS emission a loop takes; it does not touch SSR template generation.

Related but distinct from #2444 (nested-child scope-id derivation) and #2445 (Go per-row child props) — this one is about preamble value declarations not being lowered at all.

## Reproduction fixture

Not landed, because registering it would turn 4+ adapter conformance suites red. It is ready to drop into `packages/adapter-tests/fixtures/` alongside the fix (or alongside the #2444/#2445 work, which already touches the same `render-divergences.ts` files):

```ts
import { createFixture } from '../src/types'

[...]

Issue #2445, opened 2026-07-31
Go: a child component nested inside a dynamic loop row gets one hoisted props value, so every row renders it empty


kfly8:
## Shape

Same shape as #2444 — a signal-driven `.map()` whose row **root is a plain element** and whose subtree contains a **child component**:

```tsx
<ul>
  {rows().map(row => (
    <li key={row.id}>
      <Badge text={row.label} />
    </li>
  ))}
</ul>
```

Covered by the conformance fixture `composite-row-child-component`
(`packages/adapter-tests/fixtures/composite-row-child-component.ts`).

## Divergence

The Go adapter emits the nested child's props as a **single struct field on the parent**, built once outside the loop, and the template passes that same value for every row:

```gotemplate
<ul …>{{bfComment "loop:l0"}}{{range $_, $row := .Rows}}<li data-key="{{.ID}}">{{template "Badge" $.BadgeSlot0}}</li>{{end}}{{bfComment "/loop:l0"}}</ul>
```

```go
BadgeSlot0: NewBadgeProps(BadgeInput{
	ScopeID:  scopeID + "_s0",
	BfParent: scopeID,
	BfMount:  "s0",
}),
```

`Text` is never populated — there is no per-row props construction — so every row renders `<span class="badge"></span>`. Unlike #2444 this is wrong *content*, not just a divergent scope id (the scope id `<parent>_s0` is actually correct here).

Not executed end-to-end: the verifying environment has `go1.24.7` and the harness requires ≥1.25, so this is read off the emitted `component.tmpl` + `component.types` rather than from a failing render. The Go conformance run therefore currently passes locally only because `onRenderError` skips when the runtime is unavailable.

## Why the sibling shape works

When the row root **is** the child component, the adapter already emits a per-row slice — see the `todo-app-ssr` note in `packages/adapter-go-template/src/render-divergences.ts`, which describes `.TodoItems []TodoItemProps` as "the loop-body CHILD COMPONENT slice the template actually ranges over". The gap is specifically the composite case: row root is an element, child sits inside it, and the per-row slice is never built.

## Where it is declared

`packages/adapter-go-template/src/render-divergences.ts` carries a `composite-row-child-component` entry (which drives that package's `skipJsx`, and is published to `ui/compat.lock.json` / the docs compatibility-matrix page). Graduating means deleting it.

Issue #2444, opened 2026-07-31
A child component nested inside a dynamic loop row gets a random scope id on every adapter except Hono


kfly8:
## Shape

A signal-driven `.map()` whose row **root is a plain element** and whose subtree contains a **child component** — the "composite row + nested child" loop plan (`useElementReconciliation` with `nestedComponents > 0`).

```tsx
'use client'
function Badge(props: { text: string }) {
  return <span class="badge">{props.text}</span>
}
export function CompositeRowChildComponent(props: { items: Item[] }) {
  const [rows] = createSignal<Item[]>(props.items)
  return (
    <ul>
      {rows().map(row => (
        <li key={row.id}>
          <Badge text={row.label} />
        </li>
      ))}
    </ul>
  )
}
```

Covered by the conformance fixture `composite-row-child-component`
(`packages/adapter-tests/fixtures/composite-row-child-component.ts`).

## Divergence

Hono (the reference) derives the nested child's scope id from the parent scope plus the mount slot — `bf-s="test_s0"`. Every other adapter, and the CSR template lambda, emit a freshly randomized `Badge_<random>` instead:

```
Expected: <li data-key="1"><span bf-s="test_s0"    bf="s1" class="badge">…one…</span></li>
Received: <li data-key="1"><span bf-s="Badge_9fk2q" bf="s1" class="badge">…one…</span></li>
```

The child's *content* is correct on all of them — only the scope id diverges. In the DSL adapters the id comes from the runtime's `render_child` helper, which generates its own id rather than deriving one:

```blade
<li data-key="{!! e($bf->string(data_get($row, 'id'))) !!}">{!! $bf->render_child('badge', ['text' => data_get($row, 'label')]) !!}</li>
```

Confirmed failing against real runtimes: **erb**, **jinja**, **mojolicious**, **rust** (minijinja), plus **CSR conformance**. Confirmed from the emitted template (runtime not installed in the verifying environment, so not executed): **blade**, **twig**, **xslate**. Go diverges too but for a different reason — tracked separately.

Consequences: SSR output is nondeterministic per request, and does not match the reference adapter. Whether hydration still matches the node correctly (the client reads `bf-s` off the existing DOM rather than regenerating it) was **not verified** as part of this write-up.

## Related

Same class as the existing `grandchild-composition` CSR skip — the third composition level reuses `test_s0` in CSR instead of deriving `test_s0_s0`. A fix for scope-id derivation of nested children should cover both.

## Where it is declared

[...]

Issue #1065, opened 2026-04-26
compiler: plain branch loop emits mapPreamble without signal-accessor wrap


kfly8:
## Summary

In the **plain** branch-loop emission path (a \`.map()\` directly inside a conditional branch with no nested inner loops), the inner \`.map()\` callback's \`mapPreamble\` is emitted **without** rewriting loop-param references to signal-accessor form. Because \`mapArray\`'s \`renderItem\` receives the item as a signal accessor (\`item: () => T\`), bare references like \`cell.flag\` evaluate to \`(function).flag === undefined\`, silently producing wrong values.

## Repro

\`\`\`tsx
'use client'
export function CondList() {
  const [show] = createSignal(true)
  const [items, setItems] = createSignal<Cell[]>([{ id: 1, value: 'a', flag: true }])
  return (
    <div onClick={() => setItems(prev => [...prev])}>
      {show() ? (
        <ul>
          {items().map((cell) => {
            const cls = cell.flag ? 'on' : 'off'
            return <li key={cell.id} className={cls}>{cell.value}</li>
          })}
        </ul>
      ) : null}
    </div>
  )
}
\`\`\`

## Compiled (broken) output

\`\`\`js
mapArray(() => items(), __loop_s2, (cell) => String(cell.id), (cell, __idx, __existing) => {
  const cls = cell.flag ? 'on' : 'off';    // ← bare cell.flag ⇒ undefined ⇒ cls always 'off'
  const __el = __existing ?? (() => {
    ...
    __tpl.innerHTML = \`<li data-key=\"\${cell().id}\" \${(cls) != null ? 'class=\"' + (cls) + '\"' : ''}>...\`
    //                                  ^^^^^^^ template references DO get cell() wrapping
  })()
  ...
})
\`\`\`

The template-literal references (e.g. \`\${cell().id}\`) are correctly wrapped, but the preamble line is not — that asymmetry causes the bug.

## Root Cause

\`packages/jsx/src/ir-to-client-js/control-flow/plan/build-branch-loop.ts:54\`:
\`\`\`ts
mapPreambleRaw: loop.mapPreamble || '',   // ← \"Raw\" — no wrap
\`\`\`
And \`stringify/branch-loop.ts:78-80\` emits \`mapPreambleRaw\` verbatim into the \`renderItem\` body.

The composite-loop builder does it correctly (\`build-composite-loop.ts:90\`):
\`\`\`ts
[...]

PR #2450, opened 2026-07-31
Widen the lazy row gate: value-only preamble, and claim element refs per slot

kfly8:
Two of the three §9.5 lazy-eligibility items. The third (`row contains a reactive conditional`) is designed but not implemented — see the last section.

## G1 — a value-only `.map()` preamble is now lazy-eligible

A row whose callback had a preamble was refused on sight, keeping the eager emission: a root, a signal and an effect per row. The rule was "any preamble at all", over-broad by construction — what matters is whether the preamble declares row-local *reactivity*.

`analyzeLazyPreamble` replaces it with a structural proof that re-executing the statements is observationally free: a sequence of `const` declarations whose initializers contain no `new`, assignment, `++`/`--`, `await`, `yield`, `delete`, function/arrow/class expression — and no call, **with one exception: a zero-argument signal or memo read**.

That exception is the point. `const cls = selected() === row.id ? 'danger' : ''` is the shape the krausest bench writes, and it is sound in all three plan bodies because of where they run: `mapArrayLazy` wraps `createRow` and `applyItem` in `untrack()`, and `applyOuter` **is** the loop-level effect that is supposed to subscribe. Everything else stays refused with a reason naming the exact node — `createSignal` is a call, `arr.push(x)` is a call, and `Math.random()` is a call whose value would differ between `createRow` and `applyItem`.

The preamble is emitted into **`createRow` only, before the clone**: the per-row template literal interpolates the declared local, because an attribute reading a preamble local is not classified as reactive and so never becomes a wired binding. The apply bodies never reference one.

A binding that *reads* a declared local refuses the loop. That case is currently unreachable from either direction — a child-position read becomes a `preambleRegions` entry, which the gate already refuses, and an attribute-position read lands in the template as above — so the refusal is the fail-safe that keeps this sound if either fact changes, not dead weight. Modelling it instead would have meant re-running the preamble in the apply bodies and threading the preamble's own dependencies into every binding that reads a local, for a shape nothing can currently produce.

Two `#1065` regression tests moved with the shape rather than being weakened: their sources are now lazy-eligible, so the assertion that the preamble goes through `wrapLoopParamAsAccessor` (`cell.flag` → `cell().flag`) is made against `createRow`. Same `mapPreambleWrapped` computation either way.

## W2 — claim element refs per slot, on first use

An adopted row resolved its refs through one whole-row closure, `__e.refs ?? (__e.refs = __lzc_<loop>(__e))`, which ran a `qsa` scan for **every** reactive-attr slot in the row. So an `applyOuter` driving one attribute still scanned for every other slot, on every row, at hydration: a three-attr-slot row ran three scans to write one attribute — 3,000 scans across a 1,000-row list.

Each binding now claims its own slot on first use, the same deferral the content door already had. `applyOuter` on that three-slot row contains one `qsa`, and the whole-row closure is gone with it.

The cache test is `2 in __r`, not `??`. A slot whose scan finds nothing records that and is not scanned again; a `??`-guarded cache would re-scan it every tick forever. `createRow` is untouched — it resolves refs from known clone paths and never scanned.

**Measured by counting, and the count is pinned**: three scans in `applyItem` (which does write all three), one in `applyOuter`, zero in `createRow`.

The SSR heap bench does **not** move — 1573.1KB, unchanged against the 1573.2KB #2430 left it at — and cannot: its row has a single reactive-attr slot, so its `applyOuter` already claimed exactly what it wrote. No current bench app has a row whose outer bindings cover only some of several attr slots, which is the shape that would move a wall-clock number. Stated rather than papered over.

## Conformance fixture, and a third pre-existing DSL gap (#2447)

`loop-preamble-attr-value` covers the newly-eligible shape. Registering it exposed a gap independent of #2444 and #2445: **no adapter lowers a `.map()` callback preamble's value declarations into the template at all**. What they emit references a variable nothing ever assigns — `$cls` (blade/xslate), `bf.string(cls)` (twig), and on Go a parent-scope `{{$.Cls}}` — so the attribute renders empty on all eight non-Hono adapters.

Confirmed pre-existing by stashing the compiler change and re-running: ERB still renders `class=""`. Filed as #2447 and declared per adapter in `render-divergences.ts`, with the issue URL in the reason strings so the `compat-issue-freshness` job covers them. The Hono render and the CSR conformance leg both pass, so only the DSL SSR legs are skipped.

Verified against real runtimes for erb, jinja, mojolicious, rust. For blade, twig, xslate, go-template the divergence is read off the emitted template — those runtimes are absent from the environment this was written in (local Go is `1.24.7` against a harness needing 1.25+).

## Tests

All green: jsx + client + adapter-tests (4923 pass after regenerating 6 doc-example client-JS snapshots, whose diff is exactly the removed closures), all 9 adapter suites, compat.

New coverage: `lazy-preamble.test.ts` (13 refusal shapes pinned to their exact reason, 4 accepted, plus emission placement), `lazy-row-preamble.test.ts` in the client runtime (compiles the component, imports the emitted module, reads the DOM — a missing preamble splice produces an empty attribute, not a compile error), and the scan-count block in `lazy-row-eligibility.test.ts`.

## G3 — not implemented; here is what it needs

`row contains a reactive conditional` is the highest-value refusal left, and it is a larger piece of work than these two combined. The blocker, now identified:

- A row conditional is addressed by a **different marker vocabulary** — `bf-c="id"` attributes and `bf-cond-start:id` / `bf-cond-end:id` comments — from the `<!--bf:id-->…<!--/-->` pairs a claimed slot resolves. A claimed `'markup'` slot cannot address a conditional slot as it stands.
- The eager path calls `insert()` per row, which creates **one `createEffect` per row** plus two branch-template evaluations for fragment-vs-element probing. Calling it from a lazy row would reintroduce exactly the per-row reactive resource the design removes.

The tractable slice, and what makes it tractable:

1. Restrict to a conditional whose **both arms are wiring-free** — `childComponents`, `innerLoops`, `conditionals`, `events`, `reactiveAttrs`, `reactiveTexts` all empty on both `LoopChildBranchSummary`s. That strips `bindEvents`, `branchCleanup`, auto-focus and `__bfSlot` live-node capture, i.e. most of `insert`'s body. This subset covers the bare ternary in child position the spec calls out.
2. Decide **fragment-vs-element at compile time** from whether `whenTrueHtml`/`whenFalseHtml` contain `<!--bf-cond-start:`, instead of probing at runtime. That removes the two per-row template evaluations outright.
3. Add a runtime helper that performs the swap with **no effect of its own**, driven from `applyItem`/`applyOuter` with the boolean deduped in `entry.last`. Seeding reuses `insert`'s existing first-run comparison (`getTemplateRootSignature` against the live `outerHTML`), so §9.3(1) read-compare-write is satisfied without a new mechanism.

So it is mostly a *reduction* of `insert.ts` rather than new machinery — but it is a new runtime helper, a new analysis module, plan and emitter wiring, and its own unit + DOM + conformance coverage. Worth its own PR.

---


kfly8:
Three CI failures on `94d51b3a`. Two were mine and are fixed in `dfb2e858`; one is not, with evidence below.

**`test (rest)` — mine, fixed.** `lazy-row-adopted-door.test.ts`'s *precondition* asserted the emitted shape of the whole-row claim closure this PR removes. The test's two behavioural assertions were fine; only the shape pin needed retargeting to the invariant that replaced it (an adopted row starts with an empty `refs` array and fills each slot on demand). This was in my local run too and I miscounted it as one of the doc-example snapshots — my verification was too loose there.

**`update-expected-html` — mine, fixed.** `fixtures/__snapshots__/ai-chat.client.js` is a frozen client-JS bundle written by `generate-expected-html.ts`, not a bun snapshot, so `bun test -u` never touched it. Regenerated; the diff is exactly the removed closure.

**`e2e-hono` — not from this PR.** `blog.spec.ts:36` clicks `.island.like` and expects the counter to read `1`, got `0`.

- `LikeButton` is a plain island — `createSignal`, a click handler, one text slot. No loop, no `.map()` preamble, no `mapArrayLazy`. Neither change in this PR can reach it: G1 only affects loops whose callback has a preamble, W2 only the ref-claiming inside a lazy row plan.
- It does not reproduce. `bunx playwright test e2e/blog.spec.ts --repeat-each=3` on this branch: **18/18 passed**, including that test.
- `main` is green on the same spec (runs `30623594273` at `108834a9`, this PR's base, and `30631235047`).
- The assertion is `expect(await page.locator('.island.like .v').innerText()).toBe('1')`. `innerText()` + `toBe` samples once with no retry, unlike `toHaveText`, and it runs immediately after a click that triggers a router region swap plus re-hydration. With `retries: 0` and `timeout: 5000` in that config, that is a race.

The one-word fix (`toBe` → `toHaveText`) belongs with the router/blog work that just added this spec, not in a lazy-row PR, so I have not touched it here. Happy to open it separately if that is wanted.

---

kfly8:
`e2e-hono` failed the same assertion again on `dfb2e858` (2/2 on this PR). Correcting my earlier comment: I called it a non-polling matcher, which was incomplete. The real cause is a **hydration-timing gap in the test**, and swapping `toBe` for `toHaveText` would not have fixed it.

**The sequence** (`blog.spec.ts:31-36`):

```js
await page.click('.sortable-list li:first-child .item-link')   // router region swap
await page.waitForSelector('.island.like', { timeout: 3000 })  // waits for PRESENCE only
...
await page.click('.island.like')
expect(await page.locator('.island.like .v').innerText()).toBe('1')
```

`waitForSelector` resolves when the swapped-in server markup lands. Re-hydration happens afterwards, and in this app it goes through a **dynamic import**: `defaultRehydrate` (`packages/router/src/seams.ts:70-77`) falls back to `await import('@barefootjs/client/runtime')` before calling `rehydrateScope`. The build log for this very job confirms that path is the one taken:

```
Warning: runtime tree-shake — falling back to full runtime copy
(dynamic import("@barefootjs/client/runtime") in components/router-entry.js)
```

So there is a window where the button exists but its click handler is not attached. A click in that window is simply lost — which is exactly why the counter reads `"0"` (the server-rendered value) rather than being absent. And because the click is *lost*, a polling `toHaveText('1')` would poll forever and still fail; the matcher was never the problem.

That also explains CI-vs-local: the window widens under load, and this job's log shows the worker being hammered right before the failure — repeated `workerd` `disconnected: ::write(...): Connection reset by peer` / `Broken pipe`.

**Not attributable to this PR:**

- Mechanically out of reach. G1 only changes loops whose `.map()` callback has a preamble; W2 only changes ref-claiming inside a `mapArrayLazy` plan. `LikeButton` is `createSignal` + a click handler + one text slot — no loop, no preamble, no lazy row. Nothing in the router, the seams, or the dynamic import is touched.
- Does not reproduce. Full suite under CI's conditions locally (`CI=1`, single worker, all 105 tests): **105 passed**. `blog.spec.ts` alone with `--repeat-each=3`: 18/18.
- `main` passes it on CI (`30623594273` at `108834a9`, this PR's base).
- Shipped JS is byte-identical before and after this PR (27.6KB raw / 9.7KB gzip in both benchmark runs), so there is no "bigger bundle, slower hydration" story either.

**This needs a decision, so I have not pushed a fix.** Two options, and they differ in kind:

1. **Test-side only** — wait for hydration before clicking instead of for presence. There is no signal to wait on today (`[bf-s]` is already on the server-rendered markup, so the `toggle.spec.ts` idiom does not apply here), so this ends up as a retrying click: `await expect(async () => { await page.click(...); await expect(v).toHaveText('1') }).toPass()`. Contained, but it encodes "clicks may be lost" as expected behaviour.

2. **Product-side** — have the router signal when a region has finished re-hydrating (an event or an attribute), and have the test await it. This is the more truthful fix and would also give app authors something to hook, but it is a new router API and `spec/router.md` would need to say so.

My recommendation is (2), with (1) as a stopgap if you want CI green first. Either way it belongs with the router/blog work rather than in a lazy-row PR — say the word and I will open it separately.

---

Issue #2470, opened 2026-08-02
Hono: a nested ternary under a non-reactive outer condition emits a double-braced JSX expression (syntax error)


kfly8:
Found while reviewing an over-broad narrowing attempt on #2463. The narrowing is not being shipped, but the bug it exposed is **independently reachable on `main` today** — verified with a plain module-level constant as the outer condition. Silent: zero diagnostics, and the generated file simply doesn't parse.

## Repro

```tsx
const MODE = 'b'
export function Chain() {
  return <div>{MODE === 'a' ? <span>A</span> : MODE === 'b' ? <span>B</span> : <span>C</span>}</div>
}
```

## What the Hono adapter emits

```tsx
<div …>{MODE === 'a' ? <span>A</span> : {MODE === 'b' ? <span>B</span> : <span>C</span>}}</div>
```

The nested conditional is re-wrapped in its own `{…}` while sitting in the **alternate position of an enclosing ternary**, where only a plain JS expression is legal. The emitted `.tsx` fails to parse (`Expected "}" but found "==="`).

## Cause

`renderConditional` always wraps its output in `{…}` — correct for the JSX-*child* position it was written for. `renderNodeRawCtx` falls through to `renderNode` for a nested conditional, so the inner render adds a second pair of braces in a position that must stay brace-free.

Reachable only when the **outer** conditional's own `cond.slotId` is null — i.e. the non-reactive path, which never goes through `wrapWithCondMarker`'s fragment-based embedding. A signal- or prop-conditioned chain takes the reactive path and is fine, which is why the corpus never hit it: every fixture with a nested ternary chain has a reactive outer condition. A module-level constant (as above) is enough to reach it.

## Fix

Split the bare ternary text out of the braced form: `renderConditional` returns `{<body>}`, and embedding sites that splice into another ternary's branch use the bare body directly. This is the same reasoning `renderNodeRawCtx`'s existing `'expression'` special case already applies to a `null`/`undefined` literal branch.

A working implementation exists — it was written as part of the dropped #2463 narrowing commit and can be lifted as-is (`renderConditionalBody` in `packages/adapter-hono/src/adapter/hono-adapter.ts`).

## Coverage

Needs a conformance fixture with a nested ternary chain under a non-reactive outer condition. No existing fixture covers that combination, which is precisely why this survived.

Issue #968, opened 2026-04-22
compiler: top-level conditional return in "use client" component drops the conditional (SSR always renders truthy branch)


kfly8:
## Bug

When a `"use client"` component's **top-level return** is a ternary conditional expression, the compiler drops the conditional in the SSR template and always renders the truthy branch. The client JS also doesn't emit any conditional DOM reconciliation logic, so the falsy branch is never shown even after hydration.

This works correctly when the conditional is **inside** an element (not the root of the return).

## Steps to Reproduce

```tsx
"use client"
import { createSignal } from '@barefootjs/client'

export function BadgeDemo() {
  const [count, setCount] = createSignal(0)

  // BUG: conditional at top level — SSR always renders <span>
  return count() > 0 ? (
    <span>{count()}</span>
  ) : null
}
```

Compiled SSR output (actual):

```tsx
// conditional is dropped — <span> always rendered
<span bf-s={__scopeId} bf="s1">{bfText("s0")}{count()}{bfTextEnd()}</span>
```

Expected SSR output (matching the working pattern):

```tsx
// conditional preserved with bfComment markers
{count() > 0
  ? <span bf-c="s0" bf="s2">{bfText("s1")}{count()}{bfTextEnd()}</span>
  : <>{bfComment("cond-start:s0")}{bfComment("cond-end:s0")}</>}
```

## Working Pattern (for reference)

When the conditional is **inside** an always-rendered container element, the compiler handles it correctly:

```tsx
"use client"
export function BadgeDemo() {
  const [count, setCount] = createSignal(0)

  // WORKS: conditional is a child expression, not the root return
  return (
    <a href="/somewhere">
      {count() > 0 ? <span>{count()}</span> : null}
    </a>
  )
}
```

This is how `AdminUnreadBadge` works in `components/gallery/admin/admin-unread-badge.tsx`.

## Workaround

Wrap the conditional in a `<span className="contents">` container so the conditional becomes a child expression:

```tsx
return (
  <span className="contents" aria-live="polite">
    {count() > 0 ? <span>{count()}</span> : null}
  </span>
)
```

`display: contents` removes the wrapper from the layout flow, making it visually transparent.

## Impact

- Any `"use client"` component that returns `condition ? <JSX> : null` at the top level will SSR-render the JSX unconditionally.
- On hydration, the client JS only sets attributes/text on the always-present element — it never toggles its visibility.
- Result: the element is permanently visible even when the condition is false.

## Found During

[...]

PR #2472, opened 2026-08-02
Lower a component-level if/else early return to a branch switch (#2463)

### LINKED ISSUE #2463: Signal-conditioned early return: no branch-switch effect emitted; CSR template references the signal out of scope
Found during the onboarding TSX-fidelity exploration (PR #2461). Silent — zero diagnostics, zero console errors.

## Repro

```tsx
"use client"

export function LoadGate() {
  const [loading, setLoading] = createSignal(true)
  if (loading()) {
    return <button onClick={() => setLoading(false)}>Loading...</button>
  }
  return <p>Ready</p>
}
```

## What is emitted

SSR renders the initial branch correctly. The compiled client JS wires the click handler but emits **no effect that switches branches** — clicking calls `setLoading(false)` and nothing subscribes, so the UI can never leave the SSR branch:

```js
export function initLoadGate(__scope, _p = {}) {
  if (!__scope) return
  const __scopeId = __scope.getAttribute('bf-s')
  const [loading, setLoading] = createSignal(true)
  const [_s0] = $(__scope, 's0')
  if (_s0) _s0.addEventListener('click', () => { setLoading(false) })
}

hydrate('LoadGate', { init: initLoadGate, template: (_p) => `${loading() ? `<button bf="s0">Loading...</button>` : `<p>Ready</p>`}` })
```

Additionally the CSR `template` lambda references `loading()`, which is declared inside `initLoadGate` — a `ReferenceError` if the template is ever evaluated (CSR mount).

## Expected

The `IRIfStatement` contract in `spec/compiler.md`: "Client JS handles all branches and switches at runtime." The semantically identical root ternary (`return loading() ? <button .../> : <p>Ready</p>`) compiles correctly to an `insert()` with a branch-switch effect — the early-return statement form should lower to the same plan.

Not covered by the closed early-return issues (#1401/#1404/#1405/#1409/#1414/#1422) — those are prop-conditioned; this is signal-conditioned.

## Conformance coverage (PR #2461)

Fixture `signal-early-return` pins the (correct) initial-branch SSR output; the CSR-conformance skip entry for it carries this issue's URL — graduating means removing that skip. A fixture-hydrate `interactions` step (click → expect `Ready`) should land together with the fix.

kfly8:
Closes #2463.

```tsx
export function LoadGate() {
  const [loading, setLoading] = createSignal(true)
  if (loading()) return <button onClick={() => setLoading(false)}>Loading...</button>
  return <p>Ready</p>
}
```

SSR renders the initial branch correctly, the click handler is wired, `setLoading(false)` runs — and nothing happens, forever. No diagnostic, no console error. The semantically identical root ternary has always worked.

## Two independent single-site defects

**1. No branch-switch effect.** `collect-elements.ts` has a `conditional:` visitor that pushes onto `ctx.conditionalElements`; there was **no `ifStatement:` visitor at all**. `walkIR` fell through to `descendDefault`, walking `consequent`/`alternate` as flat content with `insideConditional` still false, so `emitConditionalUpdates` iterated nothing and `buildInsertPlan` never ran. Both branches' elements and events were collected as ordinary top-level bindings — which is exactly why the handler *is* wired while nothing subscribes to `loading`.

**2. The CSR `template` lambda referenced an out-of-scope signal.** `canGenerateStaticTemplate`'s `if-statement` arm never inspected `node.condition`, so a signal-conditioned early return passed the static gate and routed to `irToComponentTemplate`, which does no signal substitution:

```js
template: (_p) => `${loading() ? … : …}`   // `loading` is declared inside initLoadGate
```

A `ReferenceError` on any CSR mount. Now `(true)` — the baked initial value, as the CSR builder produces for every other shape.

## One predicate, structural

Both arms now share `hasUnsafeRef(freeIdsFromRefs(node.origin?.freeRefs))` — free identifiers resolved from the TS AST at IR-build time, never re-derived from the condition string.

The `conditional` arm previously used `condition.includes('()') && !isSimplePropExpression(...)`, a string probe against JS source — the exact thing `CLAUDE.md`'s first code convention forbids. Mirroring it into the if arm would have propagated that; instead both arms now ask the real question, so **ternaries get narrower too**. A duplicated predicate is how these two drifted in the first place, so there is now only one.

## `insertRoot`, and why it isn't `insert()`

An `IRIfStatement`'s branches are each the **whole component**, so each branch's own root carries `bf-s` directly — there is no `bf-c` anchor and no `<div style="display:contents">` wrapper (#968) that a root ternary gets. There is no stable element to query within: the scope element *is* the swap target, tracked by direct reference.

It shares `runConditionalEffect` with `insert()` — the same condition-evaluation, first-run, skip-unchanged and cleanup-ordering state machine — so the two can differ in *where* a swap lands, never in *when* one happens. On a swap, `bf-s`/`bf-h`/`bf-m`/`bf-r`/`data-key` are copied onto the incoming root so `.closest('[bf-s]')` from nested children still resolves to the scope id a parent may already have recorded.

## Cost, stated plainly

Sixteen component snapshots grow: an early return whose condition isn't provably module-scope-evaluable now carries branch machinery where a static template used to suffice. None of those components was broken. This is the price of an if-statement behaving like the ternary it is equivalent to, under rules ternaries already obey (including the wrap-by-default fallback for calls the analyzer can't prove pure, #937/#941).

An attempt to trim this by reclassifying prop-derived locals as non-reactive was **written and dropped**: it broke six invariant tests, because for a *guard* (`{prev ? <a href={prev.href}>{prev.title}</a> : null}`) the branch machinery is load-bearing — it keeps the branch's effects from running while the branch is inactive. Removing it forced `const prev = _p.prev ?? {}`, which `does not default conditional-guard props to {} in client JS` pins against directly. Correctness first; that optimization needs its own design.

That attempt did surface a genuine latent bug, filed separately as **#2470** — Hono emits a double-braced JSX expression for a nested ternary under a non-reactive outer condition, producing a file that doesn't parse. Verified reachable on `main` today with a module-level constant, independent of anything here. A working fix exists in the dropped commit and is noted on the issue.

## Verified in a real browser

Not just a regenerated snapshot — the emitted client JS bundled against the real runtime, hydrating the fixture's own SSR bytes in Chromium:

```
preHydration : <button bf-s="LoadGate_test" bf-r="" bf="s0">Loading...</button>
postHydration: <button bf-s="LoadGate_test" bf-r="" bf="s0">Loading...</button>
afterClick   : <p bf-s="LoadGate_test" bf-r="">Ready</p>
```

Scope identity survives the swap; no page errors.

## Fixtures

`signal-early-return` graduates off the CSR-conformance skip list and gains a click → expect-`Ready` interaction. New `if-statement-child-swap` covers a branch swap that exchanges child components, which the runtime-identity copying above exists for.

## Tests

Serially, one suite per run: `packages/jsx` **2923 run / 0 fail**, `packages/adapter-tests` **1493 pass / 0 fail** (real `go run` for the Go adapter), `packages/client` **627 pass / 0 fail**, `packages/cli` **1554 pass / 0 fail**. Biome clean over `packages/jsx/src` and `packages/client/src`; `tsc --noEmit` clean for both. Lockfiles regenerated with every adapter dist rebuilt on this branch — the only delta is the new fixture.

---


kfly8:
**Not mergeable yet — `e2e-site-ui` caught a real regression in this PR. Diagnosed; fix in progress.**

`badge.spec.ts` and `button.spec.ts` fail: changing the playground's `variant`/`size` updates the child's attribute but not its computed class.

## Cause

Moving a root element's reactive attributes into a branch arm switched the expression from a **live props read** to a **destructured snapshot**.

`main`:
```js
createEffect(() => {
  { const __v = `… ${({…})[(_p.variant ?? 'default')]} ${(_p.className ?? '')}`
    _s0.setAttribute('class', String(__v)) }
})
```

this branch:
```js
__disposers.push(createDisposableEffect(() => {
  { const __v = `… ${({…})[variant]} ${className}`
    __ra_s0.setAttribute('class', String(__v)) }
}))
```

`_p.variant` is re-read on every effect run, so it tracks the getter the parent forwards. `variant` is `const variant = _p.variant ?? 'default'` — taken once at the top of `init`, outside any effect. It can never change, so the class freezes at its initial value while the attribute (applied by a different path) updates. Same shape in `Badge` and `Button`; both are `if (asChild) return <Slot …/>; return <span/button …/>`.

So the branch-arm path and the top-level path use **different expression transformers** for reactive attributes. That is the same "two paths that must agree" hazard this PR removes on the static-template gate, still present one layer down.

## Why no fixture caught it

Neither if-statement fixture has a **reactive attribute on the branch's own root element**: `signal-early-return`'s branches are a static `<button>` and `<p>`, and `if-statement-child-swap` exercises a child component inside a branch. The corpus had no coverage for the combination, so a green suite meant nothing here — the E2E layer is what caught it.

Fix plus a fixture that pins the root-element-reactive-attribute shape to follow.

## Also on this PR, not caused by it

`update-expected-html` fails on five fixtures added by #2471 — `controlled-checkbox-checked`, `controlled-radio-checked`, `details-open`, `dialog-open`, `progress-meter-value`. Verified pre-existing: checking out `origin/main` with every adapter dist rebuilt and running `generate-expected-html.ts` produces the same five diffs (attribute reordering and re-indentation), with nothing from this branch involved. Worth a separate fix so the gate is meaningful again.

---

kfly8:
Status after the regression fix. The `_p.x` live-read fix is in and locally verified; what is red now is **my new fixture**, not the fix.

## Where it stands

- Regression fixed: the branch-arm path now routes through `rewriteDestructuredPropsInExpr`, the same function the top-level path uses (`emit-reactive.ts:414`). Verified by instrumenting the call — `({…})[variant]` → `({…})[(_p.variant ?? 'default')]` — and by the four regenerated snapshots (`button`, `command`, `kbd`, `tooltip`), which are exactly the components whose early-return branch root carries a reactive attribute.
- Local: `packages/jsx` 2926 run / 0 fail, `packages/adapter-tests` 1830 run / 0 fail, generated artifacts all regenerated with every dist rebuilt.
- `update-expected-html` was red from #2471, unrelated to this PR; #2476 fixes it standalone and is merged in here so this PR's CI can report on its own changes.

## What is red

`branch-root-prop-attr` — the fixture I added to pin the shape that had no coverage — fails JSX conformance on the non-Hono adapters:

```
Expected: <span … class="cls-a" data-slot="tag">tag</span>
Received: <span … class=""        data-slot="tag">tag</span>
```

Twig (and the other `test` jobs) render the class empty. The reactive class on an if/else branch's own root element, driven by a prop with a default, is not resolved in those adapters' SSR.

I missed this locally because `packages/adapter-tests` does not exercise the per-adapter suites — those live in each adapter package, and I only ran the shared one. That is the same category of gap the fixture itself exists to close, one level up.

## Two ways forward

1. **Declare it.** Add `render-divergences.ts` entries for the adapters that can't render the shape, with a tracking issue. Honest, but it parks a real cross-adapter gap behind a skip.
2. **Make the fixture's SSR expectation adapter-neutral.** The fixture exists to pin *CSR* reactivity through `interactions`; if its SSR shape is what trips eight adapters, the shape is doing work the fixture doesn't need. Passing an explicit prop instead of relying on a destructure default would likely settle it.

I lean towards 2 — a coverage fixture should fail for the behaviour it targets, not for an unrelated SSR gap — but either way the divergence itself deserves its own issue, since "prop-defaulted reactive class on a branch root renders empty" is a real SSR bug in those adapters independent of this PR.

Not pushing another round without verifying across the per-adapter suites first.

---

kfly8:
**Closing as superseded by #2478**, which fixes #2463 with a better design: signal-conditioned early-return chains normalise into the existing `IRConditional`/ternary machinery behind a narrow gate, while prop-conditioned chains stay on the statement path byte-for-byte. That means one branch-switch lowering instead of the two this PR would have shipped (`insert()` + `insertRoot()`), zero snapshot churn on components that were never broken, and — structurally — the branch-arm live-props regression this PR spent a round fixing cannot occur, because prop-conditioned early returns never enter the branch machinery at all.

Accounting for what this PR contained:

- **#2463 core fix** → superseded by #2478 (merged).
- **`insertRoot` runtime primitive, `ifStatement` collector visitor, static-gate changes, `swapRootElement` hardening** → retired with the approach.
- **#2477 root-cause fix + `branch-root-prop-attr` fixture** → extracted to #2480 on top of current main, reframed as what it actually is: the `Record[key]` one-hop-const lookup gap (all nine adapters green, no divergence entries).
- **`signal-early-return` graduation / scope-gate ledger deletion** → done differently (and equivalently) by #2478 and #2474.
- **Five-fixture `expectedHtml` regeneration** (#2476) → landed on main as `148597f2`; #2476 closed as redundant.
- **`if-statement-child-swap` fixture** → exercised `insertRoot`'s identity copying, which no longer exists; not ported. If coverage for a signal-conditioned branch swap that exchanges child components is wanted under #2478's wrapper lowering, it should be written fresh against that machinery.
- **#2473** (fragment-branch scope-representation mismatch) → filed against `insertRoot`'s attribute-copy model; with that model retired the issue as written is largely moot. Noting there.
- **#2470** (Hono nested-ternary double brace) → independent of everything here, still open, still real; the working fix noted on the issue references this branch's dropped commit and remains liftable.

---

Issue #2492, opened 2026-08-02
Go: JS-computed signal/memo initializers don't seed SSR — renders `[]` / empty where every other adapter renders the computed value


kfly8:
Found while landing the `callback-param-shadows-prop` fixture (#2482 audit follow-up). The compiler-side invalid-rewrite bug that fixture primarily pins is fixed; Go then fails it for a reason of its own.

## Repro

```tsx
'use client'
import { createSignal, createMemo } from '@barefootjs/client'

export function CallbackParamShadowsProp({ title }: { title: string }) {
  const [items, setItems] = createSignal([{ a: 'x' }, { a: 'y' }])
  const [first] = createSignal([{ a: 'p' }].map((title) => title.a).join(','))
  const joined = createMemo(() => title + ':' + items().map((title) => title.a).join(','))
  return (
    <div>
      <span>{first()}</span>
      <output>{joined()}</output>
      <button onClick={() => setItems([...items(), { a: 'z' }])}>add</button>
    </div>
  )
}
```

## Observed (rendered by real Go, `props: { title: 'T' }`)

```html
<span>[]</span><output></output>
```

Expected (Hono reference; Jinja renders it correctly too): `<span>p</span><output>T:x,y</output>`. The signal whose initializer is a `.map().join()` chain renders as `[]`, and the memo over signal + prop renders empty — hydration later snaps both to the correct values.

## Mechanism

The initializers are pure and compile-time computable, and the template-string adapters seed them at SSR. Go's Input/constructor seeding path (cf. the `todo-app-ssr` graduation notes in `packages/adapter-go-template/src/render-divergences.ts`) doesn't cover a signal/memo initial value produced by a JS method chain — it falls back to the zero value. Related in kind to the `ssr-seed-plan` derived/opaque classification (#2482 U16), but this is the Go seeding side, not the classifier.

## Pins

Conformance fixture `callback-param-shadows-prop` (all other adapters pass) + a Go `renderDivergences` entry carrying this URL.

kfly8:
[...]

Issue #2490, opened 2026-08-02
Go: spreading a loop row object mangles attribute names (`id` → `-i-d`, `title` → `-title`)


kfly8:
Found while landing the `loop-param-shadows-spread-const` fixture (#2482 audit follow-up). Distinct from the template-adapter const-shadow hole (#2489): Go picks the ROW value correctly but corrupts the attribute names.

## Repro

```tsx
export function SpreadConstShadow({ items, big }: { items: { id: number; title: string }[]; big: boolean }) {
  const attrs = big ? { role: 'note' } : {}
  return (
    <div {...attrs}>
      {items.map((attrs) => (
        <p key={attrs.id} {...attrs}>x</p>
      ))}
    </div>
  )
}
```

## Observed (rendered by real Go, `props: { items: [{ id: 1, title: 'first' }, …], big: true }`)

```html
<p -i-d="1" -title="first" bf="s0" data-key="1">x</p>
```

Expected (Hono reference): `<p id="1" title="first" …>x</p>`. The values are the row's own (the spread itself is row-scoped and correct); the NAMES come out as `-i-d` / `-title` — a camelCase→kebab conversion applied to keys that are already lowercase, inserting a dash before every letter it treats as a word boundary.

## Mechanism (suspected)

The row-map spread path (`bf_spread_attrs` seeding or its Go-side key normalization in `packages/adapter-go-template`) runs the JSX-prop→HTML-attr kebab-case pass over map keys that were already normalized once, or applies it with a per-rune uppercase test that misfires on the Go-side field names. Needs a short dig — the non-loop spread path renders the same keys correctly.

## Pins

Conformance fixture `loop-param-shadows-spread-const` + a Go `renderDivergences` entry carrying this URL.

Issue #2487, opened 2026-08-02
Go: `inLoop` is cleared (not restored) by an inner loop's exit — outer-row content after a nested loop renders through non-loop arms


kfly8:
Found by probing the unguarded name-resolution sites from the #2482 audit (U31).

## Repro

```tsx
export function Grid({ groups }: { groups: { id: number; name: string; tags: string[]; extra: Record<string, string> }[] }) {
  return (
    <div>
      {groups.map((g) => (
        <section key={g.id}>
          <ul>{g.tags.map((t, j) => <li key={j}>{t}</li>)}</ul>
          <footer {...g.extra}>{g.name}</footer>
        </section>
      ))}
    </div>
  )
}
```

## Observed (Go template)

The spread AFTER the inner loop emits the non-loop arm:

```gotemplate
…{{range $j, $t := .Tags}}…{{end}}…</ul><footer {{bf_spread_attrs .Spread_0}}>…
```

`.Spread_0` is the component-root slot mechanism — inside the outer `range` it resolves against the row item, which has no such field. The control pins the mechanism: moving the same `<footer {...g.extra}>` BEFORE the inner `<ul>` emits the correct row-scoped `{{bf_spread_attrs .Extra}}`.

## Mechanism

`packages/adapter-go-template/src/adapter/go-template-adapter.ts` sets `this.inLoop = true` on loop entry and `= false` on exit with **no save/restore**, so the inner loop's exit clears the flag for the remainder of the outer row body; the spread / child-component arms that consult `inLoop` then take their non-loop paths. The Twig-family analogue does save/restore `prevInLoop` (`twig-adapter.ts:946-947, 964`).

## Fix sketch

Save/restore: `const prev = this.inLoop; this.inLoop = true; …; this.inLoop = prev` at every loop render site (there are two pairs, ~lines 6284/6364 and 6432/6465).

## Pins

Conformance fixture `nested-loop-tail-content` (expectedHtml from the Hono reference; every other adapter passes) + a `renderDivergences` entry in the Go adapter carrying this issue's URL.

kfly8:
[...]

Issue #2486, opened 2026-08-02
Go: renderConditionExpr resolves destructured loop bindings to the root scope (`{{if $.Active}}`)


kfly8:
Found by probing the unguarded name-resolution sites from the #2482 audit (U25/U26).

## Repro

```tsx
export function DestructuredCondRows({ rows }: { rows: { id: number; label: string; active: boolean }[] }) {
  return (
    <ul>
      {rows.map(({ id, label, active }) => (
        <li key={id}>{active ? <b>{label}</b> : <i>{label}</i>}</li>
      ))}
    </ul>
  )
}
```

## Observed (Go template)

```gotemplate
{{range $i, $__bf_item0 := .Rows}}<li data-key="{{$i}}">{{if $.Active}}<b>…{{$__bf_item0.Label}}…</b>{{else}}<i>…</i>{{end}}</li>{{end}}
```

TEXT positions resolve the destructured binding correctly (`$__bf_item0.Label`), but the ternary CONDITION emits `$.Active` — the parent/root scope. With a root Input that has no `Active` field this fails at execute time; with a same-named root field every row renders the parent's branch.

## Mechanism

`renderConditionExpr` in `packages/adapter-go-template/src/adapter/go-template-adapter.ts` checks a narrower set of the adapter's four loop-scope stacks than `identifierToGoRef` does — its `identifier` and `member` arms omit `loopBindingStack`, which is exactly the stack destructured bindings live on. The omission is invisible without diffing the two guards side by side (#2482 §"Go's four loop stacks are checked in four different combinations").

## Fix sketch

Mirror `identifierToGoRef`'s full stack consultation (including `loopBindingStack`) in `renderConditionExpr`'s identifier/member arms.

## Pins

Conformance fixture `loop-destructured-param-condition` (expectedHtml from the Hono reference; every other adapter passes) + a `renderDivergences` entry in the Go adapter carrying this issue's URL. Graduate by fixing the condition path and deleting the entry.

kfly8:
Fixed by PR #2504 (merged as c580024): `renderConditionExpr`'s identifier/member arms now consult the full loop-scope stack set including `loopBindingStack`, mirroring `identifierToGoRef`. Verified on `main`: the `loop-destructured-param-condition` pin is gone from `packages/adapter-go-template/src/render-divergences.ts` and `ui/compat.lock.json` records no divergence for the fixture (locks regenerated post-fix), so the fixture now serves as the regression test. Closing.

---
[...]

PR #2504, opened 2026-08-03
fix(go-template): destructured loop bindings in conditions + inLoop save/restore (#2486, #2487)

kfly8:
**Stacked on #2503** — review that one first; this PR's diff against it is only the Go adapter.

Second of the series addressing the seven adapter-side defects from the #2482 audit (#2486–#2492). Both fixes here match an existing correct precedent in the same file rather than inventing a mechanism.

## #2486 — condition position ignored destructured loop bindings

```tsx
rows.map(({ id, label, active }) => <li key={id}>{active ? <b>{label}</b> : <i>{label}</i>}</li>)
```

emitted the root-scope `{{if $.Active}}` for the ternary condition while sibling **text** positions in the same row correctly emitted `$__bf_item0.Label`. Against a root Input without that field it fails at execute time; with a same-named field every row renders the parent's branch.

`renderConditionExpr`'s `identifier` arm consulted `resolveModuleStringConst`, `loopParamStack`'s top, `isOuterLoopParam` and `loopVarRefCount` — but never `loopBindingStack`, which is the *only* stack destructured callbacks record their names on (they push `''` onto `loopParamStack`). The canonical resolver `identifierToGoRef` checks all four, and checks `loopBindingStack` **first** so a binding whose name collides with a module string const still resolves to the row.

The fix gives the condition arm the same innermost-first lookup in the same position. The `member` arm needed no change — it recurses through the identifier arm, which I verified rather than assumed.

## #2487 — `inLoop` was clobbered by a nested loop's exit

`inLoop` was set with a plain assignment and cleared with `= false`, so an inner loop's exit cleared it for the rest of the **outer** loop's body: outer-row content after a nested loop rendered through the non-loop arms (a spread lowered to the component-root `.Spread_0`). The same content placed *before* the inner loop emitted correctly — that contrast pins the mechanism.

Both pairs now save/restore. Every other loop-scope stack in these functions (`loopParamStack`, `loopBindingStack`, `loopWrapperStack`, `loopKeyDepthStack`, `loopScalarItemStack`) was already push/pop; only this boolean was not. The idiom matches `wasInLoop` used elsewhere in this same file, and the Twig adapter's `prevInLoop`.

## Graduation, and one entanglement worth flagging

`loop-destructured-param-condition` graduates — it renders at reference parity and its pin is deleted.

`nested-loop-tail-content` does **not** graduate yet, and the reason is interesting: the `inLoop` fix demonstrably works — the spread now takes the loop arm and renders each row's own value (`data-kind="a"` / `"b"` instead of the component-root slot). What it exposed underneath is that the row spread's attribute *names* are mangled by the Go-cased-key kebab conversion (`data-kind` → `-data-kind`), which is #2490. So fixing #2487 is what made #2490 visible on this fixture. The pin stays with its reason rewritten to say exactly that, and it graduates in the #2490 PR later in this stack.

## Validation

- `bunx tsc --noEmit -p .` clean.
- `loop-destructured-param-condition` renders through real Go (`GOTOOLCHAIN=go1.25.6`) and passes — confirmed it is no longer reported as skipped.
- Full `packages/adapter-go-template` suite is running as of opening this draft; I'll post the counts here and mark the PR ready once it's green. Note for anyone re-running locally: each Go conformance test spawns a real `go run`, so a cold cache produces spurious "timed out after 5000ms" failures — re-run warm before believing them.

https://claude.ai/code/session_01MpPtSg6rTe158671pknoD2

---
_Generated by [Claude Code](https://claude.ai/code/session_01MpPtSg6rTe158671pknoD2)_

### DIFF
--- a/.changeset/go-loop-scope-condition-and-nesting.md
+++ b/.changeset/go-loop-scope-condition-and-nesting.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Fix two loop-scope resolution bugs found by the #2482 audit. `renderConditionExpr` now checks `loopBindingStack` (innermost-first, before module-const inlining) so a destructured `.map()` param used as a row ternary condition resolves to the row-scoped field instead of a root-scope one (#2486). `inLoop` is now saved/restored around both loop-rendering paths instead of being unconditionally cleared, so a nested inner loop's exit no longer clobbers the outer loop's flag for its remaining tail content, e.g. a spread attribute after a nested loop no longer misroutes to the component-root slot mechanism (#2487).
--- a/packages/adapter-go-template/src/adapter/go-template-adapter.ts
+++ b/packages/adapter-go-template/src/adapter/go-template-adapter.ts
@@ -5736,6 +5736,18 @@ export class GoTemplateAdapter extends BaseAdapter implements ParsedExprEmitter,
     switch (expr.kind) {
       case 'identifier':
         {
+          // Destructure-param bindings (`.map(({ id, active }) => active ? …)`):
+          // resolve the binding name to its accessor on the range var the same
+          // way the value path (`identifier` in `ParsedExprEmitter`) does.
+          // Without this, a destructured name falls through to `rootFieldRef`
+          // and the condition emits a ROOT-scope field instead of the row's
+          // (#2486). Innermost-first, and BEFORE module-const inlining so a
+          // binding whose name collides with a module string const still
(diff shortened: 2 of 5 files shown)


kfly8:
Full `packages/adapter-go-template` suite on this branch: **1540 pass, 0 fail** (run serially in a quiet environment, `GOTOOLCHAIN=go1.25.6`). `bunx tsc --noEmit` clean. Marking ready for review.

---

PR #2554, opened 2026-08-06
Migrate site/ui and site/core onto @barefootjs/vite; delete the legacy build machinery (#2537)

### LINKED ISSUE #2537: site/ still drives the compiler by hand — migrate it onto @barefootjs/vite and delete the copied legacy helpers
Surfaced by #2536, the Vite migration's final deletion. Not a regression — a gap the deletion made visible.

## What was found

`site/ui/build.ts` and `site/core/build.ts` — the component registry's and barefootjs.dev's own build scripts — invoke the compiler directly rather than going through any adapter or the CLI. They were written before the Vite migration and were never in its scope, so while nineteen integrations moved to `@barefootjs/vite`, our own site kept driving the pipeline that #2536 removes.

Deleting the legacy pipeline broke them, in two ways:

- `packages/cli/src/lib/resolve-imports.ts` (1302 lines) was on the deletion checklist and turned out to be **live**, reachable only through these two scripts. It is restored unchanged in #2536.
- `hasUseClientDirective`, `discoverComponentFiles`, `generateHash` (from the deleted CLI `build.ts`) and `addScriptCollection` (from the deleted Hono `build.ts`) were copied verbatim into a new `site/shared/lib/legacy-build-helpers.ts`.

## Why this needs fixing

That copy is duplicated code deleted three commits earlier in the same PR. It kept the deletion unblocked, which was the right call for that PR, but it leaves the repository in a state where:

- `resolve-imports.ts` survives solely for two site scripts, not for anything the framework ships;
- `addScriptCollection`'s regex-and-paren-counting rewrite of compiled TS lives on in `site/`, when removing it from `adapter-hono` was one of the wins of #2509 — it is precisely what CLAUDE.md's first code convention forbids;
- the site does not exercise the pipeline it is meant to showcase, so a plugin regression cannot surface there.

That last point is the real cost. The site is the most-used consumer of this compiler in the repo and it is currently the one place that would not notice if the Vite path broke.

## What to do

Migrate `site/ui` and `site/core` onto `@barefootjs/vite` the way the integrations were, then delete `site/shared/lib/legacy-build-helpers.ts` and `packages/cli/src/lib/resolve-imports.ts` along with its six test files.

Worth checking during the work whether these scripts need anything the plugin does not yet expose. They predate the three-option API and may rely on behaviour that was only ever available by calling the compiler directly — if so, that is a finding about the plugin's surface, not a reason to keep the scripts.

## Related

- #2536 — where this surfaced; carries the restored file and the copied helpers.
- #2511, #2518 — the other deferred follow-ups from this migration series.

kfly8:
Closes #2537 — the migration proper, now that all five survey blockers are resolved (#2545–#2548, #2553). Both sites stop driving the compiler by hand and build through `barefoot()` like the nineteen integrations; the copied legacy helpers and the CLI inliner they kept alive are deleted.

## What changes

**Both sites** get a `vite.config.ts` (`@barefootjs/hono/vite`, per-directory `components` entries, `templates: 'dist/components'`, `build.outDir: 'dist/static/components'` with `base /static/components/`) and a `build.ts` reduced to `vite build` + the non-compile assembly each site always owned (CSS/tokens, UnoCSS, icons, llms.txt, `_headers`; site/core additionally content.json, playground bundles, logos/snippets; site/ui additionally the passthrough copy of `shared/` modules and plain `.ts` data modules that emitted templates import). `emptyOutDir` stays scoped to `dist/static/components` — `dist/` is simultaneously the Workers asset root and an import source, so wiping it would delete content.json/templates/llms.txt/registry.

**Deleted** (the issue's actual ask):
- `site/shared/lib/site-build-helpers.ts` — the #2536 verbatim copy, including the `addScriptCollection` regex rewrite CLAUDE.md's first convention forbids.
- `packages/cli/src/lib/resolve-imports.ts` (1302 lines) + its six test files — Rollup owns import resolution/bundling now. Changeset: `@barefootjs/cli` patch.
- site/ui's import map, manifest-driven `BfPreload`, the form/chart/xyflow/zod/embla side bundles, the `@ui/`→`@/` regex rewrite over dist, `copyServerComponents`, `build:watch.ts`; site/core's import maps and per-file minification loop.

**Key mechanism swaps**
- Client JS: content-hashed Rollup chunks under `/static/components/assets/`; compiled templates carry `registerComponentScripts`/`registerComponentPreloads` with resolved URLs (dev: vite-origin; build: manifest-hashed). No import maps anywhere except the playground iframe, which still loads the standalone `barefoot.js` runtime by fixed URL (still built, now solely for that).
- site/ui aliases: `tsconfig` `@ui/*` now maps to `./dist/*` like `@/*` already did, so templates' verbatim `@ui/components/...` imports resolve to compiled output at SSR-bundle time — replacing the legacy post-build rewrite. Verified mechanically: no component file imports a non-component value through either alias, so client chunks carry no alias imports.
- `ui/package.json` now declares the library's real runtime deps (`@barefootjs/client`/`jsx`/`chart`/`xyflow`, `embla-carousel`) — the Vite graph resolves from the importing file, and `ui/` never declared them.
- Dev: `vite dev` (plugin's dev eager pass keeps `dist/components` fresh, chunks served from the vite origin) + the existing Bun SSR server, via `concurrently`.

**Semantics kept, not silently changed**: renderers keep importing server components from source where they did before; the plugin's templates for those files are emitted but unused (switching them over is a separate decision). `PlaygroundLayout` stays deliberately uncompiled (`skipDirs: ['shared']`), copied verbatim — exactly the legacy behavior.

**CI**: new `e2e-site-core` job — site/core e2e never ran in CI, proven by the `lp-demo` spec drift this PR fixes (b20d167aa regenerated demo-outputs.ts with a 9th adapter; the spec still said 8). `site/core/**`/`site/shared/**` added to path filters; `deploy.yml`'s site/core job now builds `@barefootjs/hono` (its `/vite` subpath resolves to `dist/vite.js`).

## Verification

- site/core: clean build; e2e 8/8 (includes the QuickstartCopy clipboard hydration path through the new hashed chunks); HTTP smoke over `/`, `/docs/*`, `/playground`, chunk + runtime URLs; `wrangler deploy --dry-run` bundles the Worker against the new dist.
- site/ui: clean build (258 sources, 246 chunks; templates verified to resolve `@ui/` to dist; chunks verified free of alias imports). Full e2e suite run in progress — results will be posted on this PR before it leaves draft.
- `bun test packages/cli` green (1296) after the deletions.

https://claude.ai/code/session_01C2ohTSt6GpQcQbSBADjKeQ

---
_Generated by [Claude Code](https://claude.ai/code/session_01C2ohTSt6GpQcQbSBADjKeQ)_


kfly8:
site/ui e2e verification is complete; the run surfaced (and this PR now fixes) one compiler-level defect and one class of site-level authoring drift. Full story in the last commit; the short version:

**Compiler fix — `@barefootjs/jsx` patch (changeset included).** The emitted init eagerly read props it never uses (`const children = _p.children`), and props are getters over the parent's reactive state — a slot-children getter *instantiates* child components when read, so a Label-wrapped Checkbox got a duplicate instance with a second click listener whose toggle cancelled the first. The legacy pipeline masked this via registration order (child modules registered after the parent queued its inits); Vite's ESM import graph registers children first, which is what made it bite. New `pruneUnusedPropExtractions` final pass (AST walk, `resolveFinalImports`'s shape) removes the dead extractions; unit-tested, doc-example snapshots regenerated.

**Site fix.** The legacy build loaded client JS *content-based*; the plugin's contract is the `'use client'` directive. The 14 files that emitted real client JS without the directive now declare it — restoring the gallery shells' badge/cross-page hydration through the plugin's own mechanism. Three shells also carried a private multi-return-switch `NavIcon` that only compiles as a verbatim server helper (#932) and was silently dropped from `'use client'` templates (SSR `ReferenceError`); each is extracted to its own nav-icon component with the dispatch as a compilable ternary chain.

**Where verification landed:**

| layer | result |
|---|---|
| packages/jsx full suite | 2907 pass / 2 fail — both pre-existing `map-body-no-silent-divergence`, reproduce on main |
| adapter-tests (CSR conformance) | 1872 / 0 fail |
| ui component IR tests | 1240 / 0 fail |
| packages/cli | 1296 / 0 fail |
| site/core | build + e2e 8/8 + `wrangler --dry-run` |
| site/ui | build + e2e **1120 passed**, all 66 gallery cross-page tests green |

The three e2e failures still visible locally (form-builder required-toggle ×2, studio copy-button) are **environment-specific, not from this diff**: a main-checkout **legacy** build run in the same container fails them identically — this container's Chromium dispatches two trusted clicks for a click on a label-wrapped control — and they pass in CI's Chromium (they were green on main's CI today).

CI on this PR is the remaining gate.

---

Issue #2608, opened 2026-08-10
SSR: generic HTML-escape of the `bf-p` props JSON is the dominant source of render-time variance


kfly8:
## Summary

Emitting the `bf-p` hydration payload costs ~0.5–0.6ms per render of the 1k-row benchmark document (~7–8% of barefoot's SSR time). Almost none of that is the serialization itself:

| step | median | range | note |
|---|---|---|---|
| `JSON.stringify(__hydrateProps)` | **0.07 ms** | [0.06, 0.2] | cheap, tight — inherent |
| hono's `escapeToBuffer` over that JSON string | **~0.9–1.1 ms** | **[0.2, 4.5] — ~20x spread** | generic re-scan, incidental |

The escape step re-scans an already-quote-dense JSON string for `/[&<>'"]/` and allocates one small string per quote — thousands per render. It is both ~15x the cost of producing the JSON and, more importantly, **the dominant source of barefoot's SSR run-to-run variance**: across three CI runs barefoot's server-render median moved 5.57 → 8.84 ms (~59% spread) while solid (0.47–0.53) and react (20.57–21.32) stayed within a few percent. Allocation-heavy work varying with heap/GC state is exactly this signature.

Measured on `benchmarks/ssr/apps/barefoot` with the artifacts landed in #2607 — `measure-props-escape-variance.ts` isolates the two steps (n=100), `measure-props-cost.ts` does the four-variant ablation that attributes the 7–8% share.

## Suggested direction

A **JSON-aware single-pass encoder** — escape while stringifying, rather than generic-escape after stringifying. The set of characters that can appear unescaped in `JSON.stringify` output is known and small, so the attribute-safe encoding can be produced in one pass with one output buffer instead of a scan plus per-quote allocations. Scoped to wherever `bf-p` is emitted; no change to the hydration contract or to what the attribute contains.

## Deliberately not claimed

This will **not** close the gap against solid. The same measurement found barefoot's SSR time is **75–85% hono/jsx runtime tree-walking** (JSXNode allocation + recursive `.toString()` for 1000 rows) — the category solid eliminates by compiling to static string chunks (`_$ssr(_tmpl$, …)`, verified from babel-preset-solid's real output). Marker emission is another ~20–25%, inherent to the per-slot hydration contract. So the realistic prize here is a few percent of mean and a large share of the *variance* — worth doing for benchmark stability and for not paying an avoidable cost, not as a performance strategy.

[...]

PR #2609, opened 2026-08-10
investigate(benchmarks): hono/jsx render-cost profile + cross-adapter SSR comparison

kfly8:
Stacked on #2607. Answers the follow-up question that PR raised: the ~75–85% "base structural render" cost was attributed to hono/jsx's request-time tree-walk — **is any of it fixable, and could it be contributed upstream?**

Investigation only; nothing here changes the compiler or any adapter.

## 1. The profile names a single surgical cost

Bun's JSC sampling profiler (`--cpu-prof`), 2000 renders after 50 warmup, real `compileJSX` → `HonoAdapter` → `renderToHtml` path. Per-render self-time:

| Share | Cost | file:line | Kind |
|---:|---|---|---|
| **28.3%** | native `Object.entries(props)` | `hono/dist/jsx/base.js:147` | **Incidental** — allocates a `[k,v]` array per element for typically 1–3 props; `for...in` does the same job with zero allocation |
| 23.0% | `toStringToBuffer`/`childrenToStringToBuffer`/`JSXNode` ctor | `base.js:111-199` | Inherent — the actual tree-walk |
| 19.9% | `escapeToBuffer` + `raw()` + regex probe | `utils/html.js:8-72` | Split — escaping unknown-safety values is inherent; re-escaping content the compiler already proved static is not (issue #2608) |
| 14.8% | `jsxFn` element factory | `base.js:249-279` | Mostly inherent; does `intrinsicElementTags[tag]` lookups the compiler already knows |
| 4.4% | `normalizeIntrinsicElementKey` / `isValidAttributeName` | `jsx/utils.js:12,41` | Incidental — re-validates attribute names validated at compile time |
| 3.8% | barefoot's own emitted component body | — | |
| 1.9% | `JSON.stringify` (bf-p) | — | confirms #2607's "props are a minority" finding |
| 1.5% | `jsxDEV` shim | — | **measurement artifact** — `NODE_ENV` unset, so Bun resolves the dev runtime; a production build drops this |

~93% of self-time is inside hono/jsx's runtime machinery; only ~4–6% is barefoot's emitted code. **`Object.entries` alone is bigger than the entire tree-walk** — that's the clean upstream-PR candidate.

## 2. hono/jsx already ships precompiler-shaped primitives, unused

**Observed**: `hono/jsx/jsx-runtime` exports `jsxTemplate` (tagged-template string builder), `jsxAttr`, and `jsxEscape` (`hono@4.13.1/dist/types/jsx/jsx-runtime.d.ts:10-12`) — exactly the shape a Solid-style precompiler targets. Grepping hono's `dist/` shows **nothing in the package calls them**; no precompiler, babel plugin, or CLI hook ships in this version.

More surprising: **barefoot's own `@barefootjs/hono/jsx` shim already re-exports all three** (`packages/adapter-hono/src/jsx/jsx-runtime/index.ts:14-24`) for API parity, yet the adapter's codegen never emits a call to any of them — `jsx`/`jsxs` route through `JSXNode` unconditionally.

**Inference, flagged as such**: this looks like substrate for an external precompiler that either wasn't finished upstream or was plumbed through here for type-completeness without a compiler pass targeting it. The code doesn't say why, and upstream's issue tracker isn't visible from this environment — no claim about their intent.

## 3. Barefoot can fix most of it without upstream

Side-by-side of the current emitted `.tsx` and a Solid-shaped string-building equivalent is in the writeup. Full replacement is a **large** project, and the blast radius is the iceberg, not the string-building: `TsxSourceText` is a deliberately user-visible artifact (`types.ts:824-840`, and the README leads with "TSX in. Your stack out."), `renderToHtml`'s contract, child-component calls, `<Suspense>`/streaming, hono's document-metadata hoisting and SVG namespacing all need equivalents.

**The cheap partial win**: hono's own `raw()` (already used by barefoot for `bfComment`/`bfText`) treats a prebuilt string as an escaped leaf — O(1) append, no recursive descent (`base.js:96-99`). A loop row that is **provably static-shaped** (pure HTML/text, no nested components or component-bearing conditionals) could compile to `.map(row => raw(chunk1 + esc(row.id) + chunk2 + …))`. That touches only `renderLoop` (`hono-adapter.ts:972`) — no adapter contract change, no `renderToHtml` change, no child-render change — while attacking the ~70% of per-row cost sitting in tree-walk + factory + attribute validation.

## 4. The reframe: Hono is NOT the slow backend

Measured on each backend's **real production entry point** (not the conformance harness, which under-counts — Go's `test-render.ts` never sets `BfIsRoot` so `bf-p` is never emitted; Twig's never calls `_props()`), same dataset, same box, all verified to emit 1000 rows and ~320KB:

| Backend | Entry point | Median |
|---|---|---:|
| Twig / PHP | `TwigBackend::render_named` (compiled to a real PHP class) | **7.5 ms** |
| Hono / JS | `renderToHtml` (JSXNode tree-walk) | 8.1 ms |
| Go `html/template` | `bf.Renderer.RenderFragment` | **13.0 ms** |
| Solid (reference) | `renderToString` (precompiled chunks) | **0.35 ms** |

**Go is the slowest, not the fastest**, despite parsing its template once at startup: a pprof profile shows `reflect.Value.call` at **47.9% cumulative** — `html/template.Parse` produces an AST, not bytecode, so every `{{...}}` still dispatches through reflection at execution time. Same category of request-time interpretation as hono/jsx's tree-walk, different implementation. Twig is fastest because its compile target differs in *kind* — a real PHP class with real method calls and a real `foreach`.

So: Twig/Hono/Go are within **1.7x of each other**, and Solid is **20–37x faster than all three**. This is not a Hono problem; it's "every current adapter interprets something per request."

## Artifacts (benchmark-only, no changeset)

`profile-hono-render.ts` + `analyze-profile.py`, `measure-hono-timings.ts`, `go-render-bench/` (Go harness + pprof), `twig-render-bench/` (PHP harness), `.gitignore` entries for per-run outputs.

https://claude.ai/code/session_01WcX9QTV6Ng2X6mry4HPbhT

---
_Generated by [Claude Code](https://claude.ai/code/session_01WcX9QTV6Ng2X6mry4HPbhT)_

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -115,3 +115,17 @@ Gemfile.lock
 benchmarks/results/
 benchmarks/**/.babel-out*/
 benchmarks/ssr/apps/*/.build-temp/
+
+# SSR render-gap investigation cross-adapter benches: per-run output
+# (timings.json / one.html / cpu.prof), the Go bench binary, and the
+# CPU-profile dumps profile-hono-render.ts produces — all regenerable,
+# some multi-MB.
+benchmarks/ssr/apps/barefoot/*.cpuprofile*
+benchmarks/ssr/apps/barefoot/hono-timings.json
+benchmarks/ssr/apps/barefoot/one.html
+benchmarks/ssr/apps/barefoot/go-render-bench/ssr-bench-go
+benchmarks/ssr/apps/barefoot/go-render-bench/timings.json
+benchmarks/ssr/apps/barefoot/go-render-bench/one.html
+benchmarks/ssr/apps/barefoot/go-render-bench/cpu.prof
+benchmarks/ssr/apps/barefoot/twig-render-bench/timings.json
+benchmarks/ssr/apps/barefoot/twig-render-bench/one.html
--- a/benchmarks/ssr/apps/barefoot/analyze-profile.py
+++ b/benchmarks/ssr/apps/barefoot/analyze-profile.py
@@ -1,0 +1,88 @@
+"""
+SSR render-gap investigation, item 1 (where does hono/jsx's per-request
+render time go?) -- NOT part of the product.
+
+Parses the `--cpu-prof-md` markdown dump `profile-hono-render.ts` produces
+(Bun's real JavaScriptCore sampling CPU profiler, not hand-rolled
+instrumentation) and buckets its "Hot Functions (Self Time)" table into
+named categories (hono's JSXNode tree-walk, native Object.entries(), the
+escape pass, the compiled component body, etc.), separating one-time
+compile/import cost (paid once before the measured loop, in the same
+process) from the per-render-loop steady-state total.
+
+Usage:
+    bun --cpu-prof --cpu-prof-md --cpu-prof-name=hono-render.cpuprofile.md \\
+      benchmarks/ssr/apps/barefoot/profile-hono-render.ts
+    python3 benchmarks/ssr/apps/barefoot/analyze-profile.py \\
+      benchmarks/ssr/apps/barefoot/hono-render.cpuprofile.md.md
+
+The .md profile itself is a multi-MB generated artifact -- regenerate it
+locally rather than expecting it committed.
+"""
+import re
+import sys
+from collections import defaultdict
+
+path = sys.argv[1] if len(sys.argv) > 1 else "hono-render.cpuprofile.md.md"
+totals = defaultdict(float)
+with open(path, encoding="utf-8") as f:
+    lines = f.readlines()
+
+start = next(i for i, l in enumerate(lines) if l.startswith("## Hot Functions (Self Time)"))
+end = next(i for i, l in enumerate(lines) if l.startswith("## Call Tree"))
+section = lines[start:end]
+
+for line in section:
+    m = re.match(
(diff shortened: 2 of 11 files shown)


kfly8:
Closing without merging — the findings are worth keeping, the tooling isn't worth carrying.

**Findings moved to #2611**: the cross-adapter numbers (including Go being the slowest, not the fastest, and why), the hono/jsx profile breakdown, the conformance-harness under-counting trap, and the ranked routes.

**Why not merge**: unlike #2607 — which fixed the harness everyone's numbers come from, and whose measurement scripts reproduce figures the repo now cites — everything here is one-shot investigation tooling for a question that is now answered. It adds a Go module, a PHP/Twig harness, and a Python analysis script to a repo whose CI runs none of them; they'd break silently when `bf.Renderer` or the Twig backend's API moves, and nothing would notice. Same reasoning that trimmed #2606 down to its spec entry plus the two scripts that back a cited number.

**Nothing is lost.** This PR's diff stays readable here, and the commits stay fetchable indefinitely via GitHub's PR ref:

```
git fetch origin refs/pull/2609/head && git checkout FETCH_HEAD
```

#2611 links back here, so anyone acting on the findings can pull the harnesses in one command rather than rebuilding them — which matters most for the Go pprof and Twig harnesses, where the non-obvious part is avoiding the conformance-harness trap.

---

Issue #2613, opened 2026-08-11
Enforce "loud-or-escapable": every adapter refusal must carry a verified escape


kfly8:
## The policy this encodes

Two things BarefootJS must never do:

1. **Compile clean, then be silently broken.** → must become a loud compile-time error.
2. **Refuse loudly with no working way out.** → the refusal must come with an escape that actually works.

Condition 1 has a mechanical backstop for its main surface: `map-body-no-silent-divergence.test.ts` ("sound-or-loud", shrink-only known-hole set). **Condition 2 has none.** Nothing today checks that a refusal is escapable — the escape is claimed in prose in each adapter's `suggestion` text and verified, if at all, by hand.

That gap was found the hard way: verifying it manually for #2320/#2321 (all 8 template adapters, escaped variants compiled and mounted in a real DOM) took a full session. It came back clean — but nothing would have told us if it hadn't, and nothing will tell us about the next refusal.

## Where an escape is owed — don't classify codes

Do **not** build a per-code "escapable: yes/no" registry. BF101 alone spans dozens of unrelated refusal sites with different correct escapes; such a registry is simultaneously too coarse and immediately stale.

The decidable boundary is already latent in the corpus: **a refusal owes an escape iff the same fixture compiles clean on the reference adapter (Hono) but is refused by adapter X.** The reference succeeding is the proof that the user's code was legitimate, in-subset JSX — refused only because this backend cannot host it. BF011 ("signal outside component") fails everywhere *including* the reference: the program is wrong in itself, the fix is a rewrite, no escape is owed — and it lands outside the boundary **mechanically**, with no human classification step.

The mechanical proxy already exists: **the error-severity entries in each adapter's `conformancePins`** are by construction fixtures the corpus accepts and one adapter refuses. So the invariant's domain is every `(adapter, fixture)` pair with an error pin. Nobody decides which side a new code falls on — **landing a pin is the decision**, and the floor test forces the follow-through at that moment.

**Known limit of this net, stated up front**: it covers only *adapter-specific* refusals. A compiler-wide refusal (rejected by every adapter, so never a pinned fixture) is invisible to it — see "Not in scope" below.

## What "the escape works" means — three tiers, not one predicate

1. **Compiles clean on the refusing adapter.** Asserted directly by the new test.
[...]

Issue #2321, opened 2026-07-19
Computed component-scope const as loop source can't be evaluated at SSR (`static-array-from-props` BF101 on template adapters)


kfly8:
## Summary

The `static-array-from-props` / `static-array-from-props-with-component` fixtures refuse with BF101 on all 8 non-Hono template adapters. The loop array is a component-scope `const` **computed at runtime from props**:

```tsx
const entries = Object.entries(props.reactions ?? {}).filter(([, users]) => users.length > 0)
// ...
{entries.map(([emoji, users]) => ...)}
```

Template adapters have no mechanism to evaluate such a computed const at SSR render time — only a pure-literal or module-string const is ever inlined/seeded (`resolveStaticLoopSource`, #2208). A computed const falls through to an unseeded template variable (e.g. ERB's `v[:entries]`) and would crash at request time, so the adapters honestly refuse with BF101 instead. Hono/CSR evaluate the expression natively and render correctly.

Surfaced on the [compatibility matrix](https://barefootjs.dev/docs/advanced/compatibility-matrix) render-honesty table; pinned via each adapter's `conformance-pins.ts`.

## Why a new issue

This gap was previously attributed to #2087 in the ERB/Jinja/MiniJinja pins as a placeholder ("This is a pre-existing, orthogonal gap the widened destructure gate merely exposes … not a destructure-lowering limitation, so it is NOT part of #2087's scope"). #2087 (destructure lowering, Phase B) is complete and closed — the `([emoji, users])` destructure itself lowers fine now. The remaining refusal is purely the computed-const evaluation gap, which never had its own tracker. The pins' `issue:` URLs now point here (including the five adapter packages that previously pinned BF101 with no issue attribution at all).

## Design decision required

- **Derived-const lowering through the ParsedExpr pipeline:** recognize a whitelisted derivation shape (`Object.entries(...)` + catalogued array methods on a props expression) and emit it as a template-local expression per adapter (needs an `Object.entries` lowering — a `LoweringNode` via the builtin-plugin registry — in each backend: Go, Ruby, Perl×2, Python, Rust, PHP×2).
- **Or seed at the render boundary:** evaluate derived consts in the JS domain and pass the materialized value into the template's data alongside props — changes the adapter data contract (derived values computed where? by whom?) and breaks the "template receives exactly the props" model.
- **Or declare it out of contract:** require users to precompute in the parent / pass the array as a prop, and keep the loud BF101 (document as by-design).

[...]

Issue #2320, opened 2026-07-19
Faithful SSR lowering for nested higher-order callbacks inside filter predicates (BF101 on template adapters) — successor tracker to #2038


kfly8:
## Summary

A `.filter()` predicate whose body contains a **nested** higher-order callback call refuses with BF101 on SSR template adapters (the #2038 nested-higher-order-callback gate in `@barefootjs/jsx`, shared by all template adapters):

- `filter-nested-callback-predicate` — nested `.some(...)` (`items().filter(t => !picked().some(p => p.id === t.id))`): refuses on **blade, go-template, jinja, minijinja, twig, xslate**. ERB and Mojolicious lower it to a real inline block / `grep` predicate and render to Hono parity.
- `filter-nested-find-predicate` — nested `.find(...)` truthiness check: refuses on **all 8** template adapters. `find*` returns an *element*, not a boolean, so even the ERB/Mojo inline-block path has no faithful scalar lowering today.

Hono/CSR evaluate the chain natively. Surfaced on the [compatibility matrix](https://barefootjs.dev/docs/advanced/compatibility-matrix) render-honesty table; pinned via each adapter's `conformance-pins.ts`.

## Why a new issue

#2038's scope was to **stop the silent lossy degradation** (the Kolon-lambda fallback collapsed the inner call to its receiver) and surface a loud BF101 instead. That landed and #2038 is closed. What was never scoped or implemented is the *faithful* lowering that would turn these matrix cells green. The pins referenced closed #2038; they now point here.

## Design decision required

- **Evaluator route (adapter-agnostic):** teach `serializeParsedExpr` / the runtime evaluator to serialize nested arrows so `renderPredicateEval` can handle the whole predicate — fixes all adapters at once, but expands the evaluator contract (nested scopes, shadowing) across every backend language (Go, Perl, Ruby, PHP, Python, Rust, PHP/Twig).
- **Per-adapter inline forms:** extend the ERB/Mojo precedent (real inline closures) to Blade (PHP closures exist); Go template / Jinja / MiniJinja / Twig / Xslate lack inline lambdas entirely, so they'd stay refusing without the evaluator route.
- **Nested `.find`:** needs element-returning semantics (truthiness of a found element) in whichever route is chosen — strictly harder than boolean-returning `.some`/`.every`.

Whichever slice lands must graduate the corresponding fixture pins in the same PR (change-time coupling rule, `spec/subset-conformance.md`).

Refs: #2037, #2038, #2018.

[...]

Issue #2101, opened 2026-07-05
New adapter: Pebble (JVM / Java)


kfly8:
## Goal

`packages/adapter-pebble` — a JVM backend adapter targeting [Pebble](https://pebbletemplates.io/) (Twig/Jinja-family syntax), opening the Spring/Ktor ecosystem.

## Approach

- **TS side**: Pebble's syntax is `{{ }}` / `{% if %}` / `{% for %}` / `{% set %}` — a near-verbatim port of `packages/adapter-jinja/src/` (or adapter-twig, whichever diffs smaller). Read the divergence catalogue in `jinja-adapter.ts`'s header first; most policies (truthiness routing, `bf.string` interpolation wrapping, undefined-guarded nullish, set-block children capture, reserved-word mangling) transfer directly.
- **Java runtime** (new language port): `packages/adapter-pebble/java/` — port of the BarefootJS runtime + evaluator, method surface kept verbatim (`bf.render_child`, `bf.filter_eval`, …) so templates are cross-engine-identical. Tests MUST be driven by the shared vector corpus (`packages/adapter-tests/vectors/*.json`) — write one JSON-driven runner, not hand-ported cases.
- JS-semantics watchpoints for Java: `long`/`double` split for JSON numbers (centralize in one number helper — same class of trap Rust hit), `String()`-compatible float formatting, SameValueZero `.includes`, JS `%` sign, and Pebble's strict-variables / null handling vs Jinja's `ChainableUndefined` (verify in a spike and pin the policy).
- **Conformance harness**: `test-render.ts` shells out per fixture. Use Java 11+ single-file source launch (`java Render.java`) for an interpreter-like harness, or prebuild one renderer jar once per run (the adapter-rust harness is the precedent for compiled targets — copy its caching approach, 172 fixtures × full builds is not viable).

## Deliverables

- [ ] Conformance green: no `skipJsx`; documented `expectedDiagnostics` pins (start from jinja/twig's set); 2+2 sanctioned skips only.
- [ ] Vector-corpus-driven Java runtime tests.
- [ ] Integration example — recommend **Spring Boot** for reach (accept the image size), `integrations/spring/`; standard demo set + blog; shared e2e green; SSE must not block the servlet thread pool (use async/SseEmitter — same lesson as fastapi's asyncio SSE).
[...]

PR #2615, opened 2026-08-11
test(compat): escape-coverage floor for "loud-or-escapable" (#2613 increment 1)

kfly8:
Increment 1 of #2613: make "every adapter refusal carries a verified escape" a mechanical floor instead of prose.

> **Revised after maintainer review.** The first version put the ledger in a central `KNOWN_UNESCAPABLE` set inside `packages/compat` — 112 hardcoded `"adapterId/fixtureId"` strings. That inverted the dependency direction: a community adapter (#2101/#2102/#2103) could not be added without editing a core package. The ledger now lives on each adapter's own pins. Details in the "Declaration shape" section.

## The seed number, up front — it crosses the line #2613 drew

#2613's Risks section named the seed run as the experiment, with **>30% ledgered** as the point where "an escape is owed wherever the reference succeeds" looks aspirational rather than a ratchet.

| | |
|---|---:|
| Domain — `(adapter, fixture)` pairs with an error-severity pin | **126** |
| Covered by a declared, verified escape twin | **14 (11.1%)** |
| Ledgered (`unescapable`) | **112 (88.9%)** |

**89%, nearly 3× the line.** Not presented as a clean win. The composition below is why I think the premise survives with a refinement rather than being invalidated — but the number is the number. (#2616, stacked on this, then tested that composition directly and found something better: see its verdict.)

The domain excludes `date-method-uncatalogued` mechanically — it's error-pinned on **hono too**, so the reference doesn't succeed and no escape is owed. That's #2613's boundary working with no human classification step: a compiler-wide refusal falls outside the net by construction.

## What the composition says

14 distinct fixtures × the same 8 DSL adapters — not a few weird fixtures inflating a count, but also not 14 independent problems:

| group | fixtures | read |
|---|---|---|
| `typeof`-guard family | `every`/`find`/`some`-`typeof-predicate`, `reduce`/`reduce-right-typeof-body`, `flatmap-typeof-projection` | Same class as `filter-typeof-predicate`, whose twin is wired and passing here. Looked mechanically closable |
| `fill-unsupported` | | Its docstring **claims** `/* @client */` is covered by the filter twins. The floor test turns that claim into a checked fact — currently **false** |
| `filter-nested-find-predicate` | #2320 | nested `.find()` sibling of the working `.some()` twin |
| `static-array-from-props{,-with-component}` | #2321 | increment 2's `prop-precompute` target |
| `map-array-builder-{body,escaping}` | | imperative array-builder BF021 refusal |
| `preamble-cells`, `tag-cloud` | | **Refused by design** — jsRuntime-only demos whose docstrings say a DSL escape would break every DSL integration build |

That last pair is a design finding, not debt: ledgering by-design refusals conflates "twin not authored yet" with "permanent by design", so the ledger could never legitimately reach zero. #2616 adds `escapeNotOwed` for exactly that.

## Declaration shape (post-review)

`ConformancePin` gains an optional field, declared inline on the same pin that already carries `code`/`severity`/`issue`, in each adapter's **own** `conformance-pins.ts`:

```ts
'fill-unsupported': [{ code: 'BF101', severity: 'error', unescapable: { issue: '…/2613' } }],
```

Folded into the pin rather than added as a parallel per-adapter export: a pin already *is* the adapter's complete statement about one refusal, and a sibling mechanism could drift (a pin disappears, its ledger entry lingers). `issue` nests inside `unescapable` because "why this refusal exists" and "why no escape exists yet" are sometimes different tickets.

The central test derives its whole domain and ledger from `loadCompatAdapters()`. **Zero adapter identifiers remain in `packages/compat`:**

```
$ grep -E "'(blade|erb|go-template|jinja|mojolicious|minijinja|rust|twig|xslate|hono)/" \
    packages/compat/src/__tests__/escape-coverage.test.ts packages/compat/src/escape-coverage.ts
(no output)
```

**Additivity verified by running it.** A synthetic `LoadedCompatAdapter` (id `stub-adapter`, borrowing go-template's real factory so compilation is meaningful) pushed through the same exported functions the real test uses, no repo file touched: refuses without `unescapable` → **fails**, naming `stub-adapter` / `@barefootjs/stub-adapter` so the message addresses *that adapter's author*; with `unescapable` → **passes**, zero changes elsewhere; pinning a fixture that already has a working twin → **passes** via the twin path.

**Regression backstop**: `escape-coverage-additivity.test.ts` derives its forbidden-name list from `loadCompatAdapters()` (never hardcoded) and fails if either compat file contains an `'<adapterId>/`-shaped literal for any loaded adapter. Verified it bites by temporarily reinserting `'go-template/some-fixture'`.

Honest about strength: nothing in the logic is keyed by adapter name anymore — no `Record<string, …>` or `Set<"adapter/fixture">` for a future edit to hang an adapter branch on. Someone could still type a literal by hand; nothing motivates it, and the test catches it. Structural, not compiler-enforced.

## Also built

- `EscapeKind = 'client-directive' | 'prop-precompute' | 'rewrite'` and `escapes?: ReadonlyArray<{ kind; fixture }>` on `JSXFixture` — **fixture-level**, because the same twin escapes the same fixture on every adapter. Only the "not escapable here yet" half is per-adapter.
- The floor test per pair: a verified twin (**tier 1** compiles clean; **tier 2** unpinned + non-divergent + not CSR-skipped) or a ledger declaration. Plus stale-twin, stale-declaration and freshness defences.
- Both existing twins wired (`filter-nested-callback-predicate`, `filter-typeof-predicate`), verified against their docstrings' mutual cross-references.
- `packages/adapter-tests/src/csr-skip-set.ts` — CSR skip set extracted so the floor test and `csr-conformance.test.ts` read one list.

**The incentive design is in the code, not just the issue**: the failure message leads with the ledger as the fast path — and now names *the adapter's own* `conformance-pins.ts` — while the header comment states the cost ordering (refuse + one declaration is deliberately cheaper than shipping a silent divergence, which costs the full three-piece known-limitation set).

## Bite-checks (current shape)

Removing `every-typeof-predicate`'s declaration on go-template:
> `[go-template/every-typeof-predicate] refused with an error-severity pin but has no verified escape.`
> `FAST PATH: add 'unescapable: { issue: '<url>' }' to the error pin for 'every-typeof-predicate' in @barefootjs/go-template's own conformance-pins.ts (point at #2613 if no more specific issue exists yet) …`

Adding a bogus `unescapable` to a fixture that has a working twin:
> `stale 'unescapable' on the 'BF021' pin for 'filter-typeof-predicate' on 'go-template': a working escape twin ('filter-typeof-predicate-client') now exists here — remove 'unescapable' from that pin in @barefootjs/go-template's own conformance-pins.ts`

Both restored; suite back to green.

## CI gap found and fixed

`ci-compat.yml` is the only workflow that runs `packages/compat`, hence the only one that runs this floor test — and it was gated on `packages/adapter-tests/coverage-map.json` alone. A PR editing a fixture's `escapes` field or `csr-skip-set.ts` would trigger `ci.yml` (whose test job never runs `packages/compat`) and the floor test would **silently not re-run**. Path filter widened to `packages/adapter-tests/**`.

## Verification

`packages/compat` **330 pass / 0 fail** (re-run at the tip after the refactor) · `packages/adapter-tests` **1897 / 0 fail** unchanged · `packages/jsx` **3043 / 0 fail** unchanged.

**Changeset added** (`@barefootjs/jsx` + the 8 adapter packages, patch): `unescapable` is a published type addition and the public mechanism a community adapter author uses to declare their own escape debt, so it belongs in the changelog rather than being waved through with `no-changeset`.

Refs #2613

https://claude.ai/code/session_01WcX9QTV6Ng2X6mry4HPbhT


kfly8:
## Fixed the inverted dependency (pushed as `65341e5ae`)

You're right, and it was the wrong shape to land. The ledger was 112 hardcoded `"adapterId/fixtureId"` strings in a core package's test — which would have meant a community adapter (#2101/#2102/#2103) couldn't be added without editing `packages/compat`. Adapters are supposed to be self-describing; this made them non-additive.

**The ledger now lives on the pin**, in each adapter's own `conformance-pins.ts`:

```ts
'fill-unsupported': [{ code: 'BF101', severity: 'error', unescapable: { issue: '…/2613' } }],
```

A pin already *is* the adapter's complete statement about one refusal, so folding "and there's no escape yet, tracked here" into it beats a parallel per-adapter export that could drift out of sync with the pin it qualifies. `issue` nests inside `unescapable` rather than reusing the pin's own — "why this refusal exists" and "why no escape exists yet" are sometimes different tickets.

The central test now derives its whole domain and ledger from `loadCompatAdapters()`.

**Zero adapter identifiers remain in `packages/compat`:**

```
$ grep -E "'(blade|erb|go-template|jinja|mojolicious|minijinja|rust|twig|xslate|hono)/" \
    packages/compat/src/__tests__/escape-coverage.test.ts packages/compat/src/escape-coverage.ts
(no output)
```

**Additivity verified by running it, not by asserting it.** A synthetic `LoadedCompatAdapter` (id `stub-adapter`, borrowing go-template's real factory so compilation is meaningful) was pushed through the same exported functions the real test uses, with no repo file touched:

- refuses a fixture, no `unescapable` → **fails**, and the message names `stub-adapter` / `@barefootjs/stub-adapter` — i.e. it tells *that adapter's author* what to declare in *their own* package;
- same stub with `unescapable` added → **passes**, zero changes anywhere else;
- stub pinning a fixture that already has a working corpus twin → **passes** via the twin path, again with no central edit.

**Backstop against regression**: a new `escape-coverage-additivity.test.ts` derives its forbidden-name list from `loadCompatAdapters()` (never hardcoded) and fails if either compat file contains an `'<adapterId>/`-shaped literal for any loaded adapter. Verified it bites by temporarily reinserting `'go-template/some-fixture'`.

Worth being precise about how strong that is: nothing in the logic is keyed by adapter name anymore — there is no `Record<string, …>` or `Set<"adapter/fixture">` for a future edit to hang an adapter-specific branch on. Someone could still type a literal by hand, but nothing in the architecture motivates it and the test catches it immediately. Structural, not compiler-enforced.

Ledger after the move: **11 `unescapable` pins per adapter × 8 = 88**, unchanged in substance from the Set it replaced. `packages/compat` 330 pass / 0 fail (I re-ran this myself at the tip, given the branch churn below).

### One process note

Two of my agents ended up working these stacked branches concurrently — my coordination error, not theirs; each behaved sensibly given it couldn't see the other. Nothing was lost (nothing had been pushed, and both efforts' commits are in the history), and the ledger's stale-declaration check caught one migration gap on its own during the merge — `filter-nested-find-predicate`'s `unescapable` needed graduating now that its twin works, which is `7e299a785`. The invariant catching its own migration error is a small point in its favour.

`claude/2613-typeof-escape-twins` (#2616) was force-updated onto this fixed base.

---

kfly8:
Copilot's three comments (two here, one on #2616) were all valid and all landed in `870fdc121`. Each was the same shape: a guard whose stated contract was wider than what it actually checked.

**1. `evaluateFixtureEscapeCoverage` throws on an unpinned fixture** (#2616 thread) — accepted, and sharper than "defensive nicety": the function's own docstring says *"Pure function (no `expect`/`throw`) so it can run outside `bun:test`"*, and `adapter.pins[fixtureId].filter(...)` violated that. Unreachable from the test path (domain ids are derived *from* `adapter.pins`), reachable from the standalone use the docstring advertises and that this PR's additivity claim was verified with. Now returns an outcome.

**2. Additivity backstop missed backticks** — accepted as-is. The key it guards is a composed `adapterId/fixtureId`, so a template literal is the *most* natural way to rebuild one; checking only `'` and `"` let precisely the likeliest regression through.

**3. Reference-adapter guard** — accepted, but deliberately **narrower** than suggested. Copilot asked for a guard "that includes skipped adapters/reasons"; `compat-pins.test.ts:29` already asserts `expect(skipped).toEqual([])` package-wide, so a general load guard here would duplicate it. What is *not* covered elsewhere is specific: hono's absence silently no-ops the domain **subtraction**, changing what this file *means* rather than merely reducing coverage. So the guard targets hono only — while still surfacing skip reasons inline, since re-running a sibling file to get the diagnosis is a detour.

I also corrected my own first draft of that message: it said "treat every failure in this file as spurious," which is false. Only failures naming reference-refused fixtures are spurious; others stay real. The message now says that.

### Each verified to bite

| Fix | Check | Result |
|---|---|---|
| 1 | probed `xslate` with an unpinned fixture (`counter`) | returns the outcome; previously a `TypeError` |
| 2 | temporarily added a `` `xslate/some-fixture` `` literal | additivity test failed, naming the file; restored green |
| 3 | simulated hono's absence | guard fired; 9 spurious failures appeared on reference-refused fixtures, exactly as predicted |

### One finding worth flagging, unrelated to the review

This PR's body claims `packages/compat` **330 pass / 0 fail**. That reproduces only against a **current `bun run build`**. `loadCompatAdapters` imports each adapter from its `dist`, and `unescapable` is declared in `src` — so with a stale build the field is simply absent at runtime and the suite reports **~112 spurious "refused … but has no verified escape" failures**, i.e. every ledgered pair at once. I hit this before touching anything and briefly mis-attributed it to my own edits; `bun run build` clears it entirely (now **331 pass / 0 fail**, +1 for the new guard).

Not a code problem — CI always builds fresh — but the failure mode is loud, plausible-looking, and points at the ledger rather than at the build, so it is worth knowing before someone else loses time to it.

`bun run lint` clean.

---

Issue #2483, opened 2026-08-02
General-case marker elision — the only route to closing the HTML-size gap against solid


kfly8:
Promoting the §8 follow-up in `spec/slot-unification.md` to a tracked work item, so it can be scheduled once the correctness backlog (#2460's remaining adapters, #2466 re-measurement, #2470) is clear.

## Why

BarefootJS ships 318.6KB of HTML on the benchmark document against solid's 235.9KB (raw; CI benchmark comment). The gap is marker bytes — `<!--bf:sN-->…<!--/-->` pairs around dynamic text. Marker elision is **the only route** to closing it: client JS and heap are already ahead of solid, and the document is otherwise the same content.

The narrow case already shipped: `/* @client */` text-only slots elide their markers, justified by node count and memory (not transfer size). This issue is the general case.

## Why it is not a small change

A text span's rendered width is **data-dependent**, so the absolute positions of following siblings differ per request. Eliding a marker therefore requires the claim plan to know, at compile time, that a position is recoverable without the marker — which today's pipeline cannot ask, because **SSR generation runs before client JS generation**, and it is the client-JS side (the claim planner) that knows which slots need marker-anchored positions.

So the prerequisite is inverting (or splitting) the compiler's call order — and that inversion is **unverified against the loop-row planner's invariants**. The §5a measurement that justified deferral still stands: the risk is concentrated in loop rows, where claim paths are hoisted per-loop and adopted rows read positions back from SSR DOM.

## Suggested first step (independent, before any elision work)

An investigation-only change: swap or split the generation order behind a flag, run the full conformance + fixture-hydrate + e2e-hono suites, and **catalogue what breaks and why**. That catalogue decides whether the elision lands as one change or needs the planner's invariants reworked first. Do not fold the investigation into the elision PR — the §5a history shows estimates in this area have been wrong in both directions, and the reusable artifact is the verified list of constraints.

## Constraints to carry in

- The claim-once rule (§2) must hold: elision changes where positions are *anchored*, never how often they are *resolved*.
- Adopted rows (SSR DOM read-back, `lazyClaimSlots`) are the hard case — their paths must survive without the marker they currently anchor to.
- The declared-divergence gate: any adapter that cannot elide a given shape must keep emitting the marker rather than silently diverging; the fixture corpus should compare marker-elided output across all 9 adapters the way it compares everything else.

## References

[...]

PR #2618, opened 2026-08-11
fix(jsx): consult markerless in generateCsrTemplateWithOpts — graduates 7 twins (#2617)

### LINKED ISSUE #2617: SSR/CSR divergence: `generateCsrTemplateWithOpts` ignores `markerless`, emitting slot markers SSR elides
## The bug

`generateCsrTemplateWithOpts` (`packages/jsx/src/ir-to-client-js/html-template.ts:2229`) never consults `IRExpression.markerless` before emitting `<!--bf:sN-->…<!--/-->` for a `clientOnly && slotId` expression.

The result is a real SSR/CSR divergence for `/* @client */` text expressions in **bare, non-loop positions**: SSR correctly elides the markers, while the standalone CSR template still embeds them.

Observed across 7 new corpus fixtures (all failing `csr-conformance.test.ts`), with SSR emitting the correctly-elided
```html
<div bf-s="test" bf="s1"></div>
```
while the CSR template for the same component embeds `<!--bf:s0-->…<!--/-->`.

## It contradicts two things already written down

1. **`client-only-elision.ts`'s own docstring** claims "every CSR emitter … reads the single markerless flag." It does not.
2. **`spec/slot-unification.md` §5a**, added by #2606, states the CSR side is already general. That investigation checked `irToHtmlTemplate` (`html-template.ts:737`), which *does* check `markerless` (line 813) — and missed that there is a **second** CSR emitter. The only `markerless` reference in the whole file sits inside the first one.

**The spec correction is part of this issue's fix**, not optional: as written, §5a tells whoever picks up general-case marker elision (#2483) that the CSR side needs no work, which is now known to be false. That statement is currently misleading in `main`.

## Why it went unnoticed until now

No corpus fixture exercised the shape. The two pre-existing escape twins (`filter-typeof-predicate-client`, `filter-nested-callback-predicate-client`) are both **loop-shaped**, where the container element is the mount anchor and no separate markerless slot is in play. The bare-text-expression `/* @client */` position had never been covered.

It surfaced while authoring escape twins for #2613 — the "loud-or-escapable" invariant — where 7 of 8 twins compiled clean, unpinned and non-divergent, yet failed real CSR conformance. That is the invariant earning its keep in a way its own design didn't anticipate: not by finding a refusal without an escape, but by driving a *new shape* through the corpus and finding a latent emitter gap.

## Pins (the reproducible defect is already landed as fixtures)

PR #2616 lands 7 fixtures pinning this, CSR-skipped with the cause traced in each docstring and in `csr-skip-set.ts`:

`every-typeof-predicate-client`, `find-typeof-predicate-client`, `some-typeof-predicate-client`, `reduce-typeof-body-client`, `reduce-right-typeof-body-client`, `flatmap-typeof-projection-client`, `fill-unsupported-client`

Graduation: make `generateCsrTemplateWithOpts` consult `markerless` (mirroring `irToHtmlTemplate:813`), delete the seven CSR skips, declare `escapes` on the seven refused fixtures, and remove their `KNOWN_UNESCAPABLE` entries — at which point #2613's ledger drops by a further 56 pairs and the fixtures flip into this fix's regression tests.

## Knock-on: it also blocks #2613's stated boundary

#2613's seed comment guessed that most ledgered pairs were merely "twin not authored yet". This bug is why that guess was wrong: the invariant is **a ratchet for the loop-shaped class and aspirational for the bare-text-expression class** until this is fixed. Fixing it converts the larger of the two groups.

## Related

#2613 (the invariant that surfaced it) · #2616 (the pinning fixtures) · #2483 (general-case marker elision — the work §5a's incorrect claim would mislead) · #2606 (where that claim landed)

kfly8:
Stacked on #2616. Fixes the SSR/CSR divergence #2616 discovered, then graduates the 7 fixtures it had to pin.

## The bug, in one line

`generateCsrTemplateWithOpts` never consulted `IRExpression.markerless` before emitting the marker pair for a `clientOnly && slotId` expression. SSR elided the markers; the standalone CSR template still embedded them.

```ts
if (node.clientOnly && node.slotId) {
+ if (node.markerless) return ''
  return `<!--bf:${node.slotId}--><!--/-->`
}
```

The nested placement is the point. All nine SSR adapters use exactly this shape — `if (expr.markerless) return ''` inside their own identical `clientOnly && slotId` branch (`hono-adapter.ts:813`, `go-template-adapter.ts:3640`). Copying the *sibling CSR emitter's* shape instead (`irToHtmlTemplate:813`, a bare `${expr}`) would have introduced a new bug: eagerly evaluating a deferred `/* @client */` read at template-build time.

## Why the two CSR emitters were not collapsed into one helper

They have genuinely different `clientOnly` semantics, so a shared helper would have to re-introduce the distinction internally. `irToHtmlTemplate` "does not honour `clientOnly`" by its own docstring; `generateCsrTemplateWithOpts` does, matching the SSR adapters.

The comment justifying that split **originally stated the invariant backwards**, and was corrected in `5d5b2fec5` during review. It claimed a markerless node reaching `irToHtmlTemplate` "is, by construction, never today's sole eligible case (`clientOnly && slotId`)". The opposite holds — `markElided` has a single call site, gated on `child.clientOnly && child.slotId`, so `markerless` **implies** `clientOnly && slotId`; no non-`clientOnly` markerless slot exists.

What actually makes eager evaluation safe there is *reachability*:

1. `markerless` ⟹ `clientOnly && slotId` (single call site, `client-only-elision.ts:193`).
2. The elision walk never descends into loops/conditionals — its `default:` case freezes the level and returns without recursing.
3. `irToHtmlTemplate` only ever runs **on** loop bodies and conditional branches (verified across all 19 call sites).

(2) + (3) make the branch dead code; (1) means that if it ever came alive, evaluating eagerly would be wrong. That distinction is load-bearing rather than pedantic: the erroneous version erred in the reassuring direction, and #2483 widening elision into loop/conditional subtrees is exactly what would falsify it. Both the comment and `spec/slot-unification.md` §5a now record that the branch must gain a `clientOnly` deferral check at that point rather than being carried over as-is.

## Graduation

| | #2616 | Now |
|---|---:|---:|
| Covered by a verified twin | 22 (17.5%) | **78 (61.9%)** |
| `escapeNotOwed` | 16 (12.7%) | 16 (12.7%) |
| `unescapable` ledger | 88 (69.8%) | **32 (25.4%)** |

7 CSR skips removed, `escapes` declared on the 7 refused fixtures, 56 `unescapable` declarations removed across 8 adapters. Ledger verified independently by grep (4 per adapter × 8) and by scripting `evaluateFixtureEscapeCoverage` over all 56 pairs.

This flips #2613's boundary: the bare-text-expression class is now a **ratchet** alongside the loop-shaped one, not aspirational.

## Pre-existing snapshots that moved — and why that is the fix working

6 CSR snapshots changed (5 doc-example, 1 `todo-app.client.js`), all the same shape:

```
- <strong bf="s7"><!--bf:s0--><!--/--></strong>
+ <strong bf="s7"></strong>
```

All are bare top-level `/* @client */` text expressions inside elision scope. The corresponding **SSR** snapshots are unchanged and already read `<strong bf="s7"></strong>` — so CSR converged onto SSR rather than drifting from it. Verified directly against `todo-app.html`, not inferred from the regeneration.

## Verification

`packages/jsx` **3074 pass / 0 fail** · `csr-conformance` **266 / 0** (up from 259 = +7 twins) · `packages/compat` **330 / 0** · `bun run lint` clean. `expectedHtml`, `coverage-map.json`, `compat.lock.json`, `support-matrix.lock.json` all regenerated — the latter three with zero diff.

Changeset patches `@barefootjs/jsx` (real emission change) plus the 8 adapters (declaration-only pin changes; SSR output was never affected).

## Follow-up found, not fixed here

A **third** emitter, `irToComponentTemplateWithOpts` (`registerTemplate()`'s static-template fast path), has no `clientOnly` handling in its `expression` case at all — unlike its own `conditional` case, which does. Provably unreachable today (`canGenerateStaticTemplate` rejects expressions with non-trivial calls, and every `clientOnly` corpus fixture has them), so no fixture reproduces it and nothing is pinned. Out of #2617's scope; I'll file it separately rather than widen this PR.

Closes #2617. Refs #2613, #2483

---
_Generated by [Claude Code](https://claude.ai/code/session_01WcX9QTV6Ng2X6mry4HPbhT)_


kfly8:
Both editorial points actioned in `a3306b3b7`, though neither diagnosis was quite right — worth recording so the reasoning isn't lost:

**`client-only-elision.ts` docstring.** Not a subject/verb agreement error: the subject was still the singular *"Every consumer"*, with `both ... emitters` sitting inside an em-dash aside. But the underlying report is sound — the sentence had three stacked em-dash clauses and stranded its verb about four lines from its subject. Split into two: the invariant first, the consumer inventory second.

**`spec/slot-unification.md` code span.** This one does render correctly — CommonMark converts line endings inside a code span to spaces, so `` `clientOnly &&⏎slotId` `` renders as `clientOnly && slotId`, not broken formatting. Joined anyway, since single-line spans are easier to grep and edit.

Comments and spec only; no code change, so the verification in the PR body still stands.

---

PR #2624, opened 2026-08-12
docs(compat): make the matrix answer "does it work?" first (#2613)

kfly8:
Follow-up to #2622 on maintainer feedback:

> Make the compatibility matrix more intuitive. Answer *does it work?* first, before issue information. If it works with a client directive, users should perceive it as **working**.

#2622 made escapable refusals *distinguishable* (`BF101†`), but the cell still led with an error code — so a construct that genuinely works, by adding one `/* @client */` comment, still read as a failure. This inverts that.

## The cell now answers "does it work?"

| state | was | now |
|---|---|---|
| parity | `✓` | `✓` |
| refused + verified escape | `BF101†` | **`✓†`** |
| refused, no escape | `BF101` | `BF101` (unchanged) |
| refused, no escape owed | `BF021‡` | `BF021‡` (unchanged) |
| render divergence | `≠` | `≠` |

The check mark answers *whether*; the dagger answers *how*. The diagnostic code moves out of the escapable cell into the detail list — that is the "issue info second" instruction.

## Fully-working fixtures leave the table

This goes beyond the literal ask and is the biggest change, so flagging it explicitly for your call: a fixture that works on **every** adapter — even if some need `/* @client */` — is no longer listed as a divergence. Listing it under "these diverge" contradicts the premise that it works.

```
315 / 320 fixtures work on every adapter (12 of those need a /* @client */
comment on at least one adapter). The 5 below need attention:
```

The table drops from **17 rows to 5**. Nothing is lost: every escapable fixture remains in the per-fixture detail list, now reading

> `blade`, `erb`, `go-template`, … — works with a `/* @client */` comment — **verified escape** (†): [`every-typeof-predicate-client`](…) compiles clean here, suppressing `BF101`

Verified on the real docs page, not just the CLI: I mounted `registerCompatMatrixRoutes` on a Hono app and fetched `/advanced/compatibility-matrix.md`. The CLI's `--md` output doesn't carry the detail list, so checking only that surface would have looked like those fixtures had vanished.

## The 5 that need attention

```
| date-method-uncatalogued               | BF021 | BF021  | … (refused on hono too)
| preamble-cells                         | ✓     | BF021‡ | …
| static-array-from-props                | ✓     | BF101  | …
| static-array-from-props-with-component | ✓     | BF101  | …
| tag-cloud                              | ✓     | BF021‡ | …
```

`date-method-uncatalogued` is refused on the reference adapter too, so no escape is owed anywhere; `preamble-cells` / `tag-cloud` are `‡` (no escape owed, by design). **The `static-array-from-props` pair (#2321) is the only genuine dead end left.**

#2623 merged during this PR and is included here: it graduated `map-array-builder-*` to escapable, moving both rows out of the table and into the working count. The total rises to 320 because #2623 also adds the twin fixture itself to the corpus.

## Section prose

Rewritten to lead with the question rather than the taxonomy:

> This section answers the question that actually matters: **does the construct work**, on each adapter's real backend … A construct works either **as written** (`✓`) or with a documented `/* @client */` comment (`✓†`, a verified, supported escape — most refusals have one).

"most" is load-bearing and checked: 90 of 126 in-domain pairs are escapable (71%).

Copilot caught a self-contradiction in an earlier draft of this prose — it said absent fixtures work "no escape needed" while the headline counts fixtures that work *because of* an escape. Fixed on both surfaces; see the comment below.

## One honest wrinkle

`✓†` currently appears in **no table cell** — every fully-escapable fixture is filtered out, and no fixture today is escapable on some adapters *and* broken on others. The state is real and the code produces it (the filter is per row, not per cell, so a mixed row renders `✓†` beside a bare code), but it is unpopulated right now. `†` itself does appear, in the detail-list prose. I left the legend entry in rather than hiding a state the code can reach.

## Scope

Rendering only. `classifyFixtureEscapeState` / `evaluateFixtureEscapeCoverage` are untouched — confirmed by the `ui/compat.lock.json` diff containing nothing but the note string, and `escape-coverage-additivity.test.ts` still passing. No adapter identifier entered `packages/compat`.

`packages/compat` **352 pass / 0 fail**. `bun run lint` clean. Locks regenerated with the repo's own commands.

Refs #2613, #2622, #2623


kfly8:
Both review comments are the same defect in two copies of one sentence, and both are correct — fixed in `0acd27756`, along with merging #2623.

**The contradiction.** The section said fixtures absent from the table *"work on every adapter, no escape needed"*, while the headline immediately below counts fixtures that work **only because** of a `/* @client */` comment. I read this prose myself before opening the PR and didn't catch it. Both surfaces now read:

> Fixtures absent from the table below work on every adapter — most as written, some via that documented escape; the headline says how many of each.

**#2623 merged in**, which moved the numbers:

| | before | after |
|---|---|---|
| work on every adapter | 312 / 319 | **315 / 320** |
| …of those needing `/* @client */` | 10 | **12** |
| need attention | 7 | **5** |

The total rises to 320 because #2623 adds the twin fixture itself to the corpus.

**The test conflict needed resolving, not just picking.** Two assertions named `map-array-builder-body` as the still-broken example — precisely the fact #2623 inverted. Taking either side unchanged would have left assertions that pass while claiming the opposite of the truth. Both now use `static-array-from-props` (#2321, genuinely unescapable), and I added a positive assertion that `map-array-builder-body` *does* work everywhere, so #2623's graduation is pinned as observable through this renderer rather than silently dropped.

The 5 remaining: `date-method-uncatalogued` (refused on hono too, so no escape is owed anywhere), `preamble-cells` / `tag-cloud` (`‡`, no escape owed by design), and the `static-array-from-props` pair (#2321) — the only genuine dead ends left.

`packages/compat` **352 pass / 0 fail**, locks regenerated with the repo's own commands, lint clean.

---

Issue #2273, opened 2026-07-14
Close the unknown-type method-call silent passthrough (roadmap 4 prerequisite)


kfly8:
`spec/subset-conformance.md` ("Date: first catalogued rich type") documents the verified failure mode: a method call on a prop whose type has no known lowering compiles with **zero diagnostics** and is transliterated mechanically. The canonical probe:

```tsx
function Stamp({ createdAt, label }: { createdAt: Date; label: string }) {
  return <time>{createdAt.toISOString()}</time>
}
```

Go emits `{{.CreatedAt.ToISOString}}` on an `interface{}` field — `html/template` dies at request time ("can't evaluate field ToISOString") while CSR renders correctly, the worst possible split. The only thing that "works" today is an undocumented accidental contract: hand the adapter a Go type with PascalCase-mirrored methods.

This violates the loud-boundary principle (everything outside the subset is BF021/BF101, never a silent passthrough) and blocks roadmap 4: the Date catalogue needs the refusal in place first so the catalogue OPENS holes in a closed wall rather than papering over an open one (growing-only, refuse-first).

Work:
- A method call on a prop whose resolved type has no catalogued lowering becomes a diagnostic (BF021 class), with `/* @client */` as the escape hatch.
- Decide severity on evidence: if the accidental mirror-method contract has real users, stage through a warning release first; otherwise go straight to error.
- Conformance: an `expectedDiagnostics`-pinned fixture for the Date probe shape (per the tests-with-change rule).

Unblocks the Date catalogue issue (filed separately).

PR #2647, opened 2026-08-17
feat(jsx): BF049 — refuse a JSON-unsafe rich-typed prop read (not just method-called) by client code

### LINKED ISSUE #2643: Hono SSR throws on a BigInt prop, silently serializes Map/Set props to {}, with no diagnostic
Split out of #2642's design pass (which deferred the larger "typed-prop revival across `bf-p`" question — this is a narrower, independent bug that doesn't need that decision to be fixed).

## Problem

BF021 (`checkRichTypeMethodCalls`, `packages/jsx/src/rich-type-refusal.ts`) only fires on a **method call** on a host rich-typed prop. It does not fire when a rich-typed prop is simply **passed through** with no method call on it — which is legitimate in general (an untouched prop value doesn't need a template-lowerable expression), but for a `Map`/`Set`/`BigInt`-typed prop this legitimate case still breaks at SSR, silently or loudly depending on the type:

- **`BigInt`**: Hono's SSR serializer (`JSON.stringify(__hydrateProps)` in `packages/adapter-hono/src/adapter/hono-adapter.ts`, ~line 621) **throws** `TypeError: Do not know how to serialize a BigInt` — the whole SSR render fails, with no compile-time warning that this shape was ever unsound.
- **`Map` / `Set`**: the same `JSON.stringify` call **silently** serializes either to `{}`, dropping all entries with no error at all. The client then hydrates against an empty object where the component expected a populated `Map`/`Set`.

Neither case has a diagnostic today. A user passing `<Foo data={myMap} />` gets no compile-time signal that this will misbehave at runtime.

## Scope

This is deliberately narrower than #2642 (deferred — general typed-prop revival is a real design question, a 9-adapter protocol change). This issue is just: **make the existing failure loud and, ideally, compile-time**, without attempting revival.

## Proposed fix direction

Not yet designed in detail — needs its own look, but the shape is:
1. A compile-time check (mirroring `checkRichTypeMethodCalls`'s existing type-provable-from-propsType approach in `rich-type-refusal.ts`) that flags a rich-typed prop being passed to a JSX element/component when that type is known to be JSON-unsafe (`Map`, `Set`, `WeakMap`, `WeakSet`, `BigInt`, `Symbol`, `Function`, `Promise` — i.e., every `HOST_RICH_TYPE_NAMES` entry except `Date`/`URL`/`RegExp`/`Error`, which at least degrade to *something* rather than losing data or throwing).
2. Alternatively/additionally, a runtime-side serialize-time check in Hono's own prop-serialization step that throws a clear, actionable error instead of `JSON.stringify`'s generic `TypeError` (for `BigInt`) or silently succeeding with data loss (for `Map`/`Set`).

## Note

Per #2642's design pass, `URLSearchParams`/`RegExp`/`BigInt` are the cheapest/highest-value candidates for an eventual *real* revival mechanism (type-directed use-site revival, not a wire sentinel) if that's ever built — but this issue only asks for the failure to become loud/diagnosable, not for revival.

Related: #2642 (the deferred general design question this was split from), #2636/#2639/#2640/#2641 (the BF021/`/* @client */` family this is adjacent to).

kfly8:
## What

Closes #2643.

BF021 (#2273) refuses at compile time when client code calls a **method** on a prop provably typed as a host rich type (`Date`, `Map`, `Set`, …) with no catalogued lowering. It has no coverage for the sibling shape: a rich-typed prop that client code merely **reads and passes through untouched**, with no method call anywhere. That prop still crosses the `bf-p` JSON hydration boundary with no type-aware revival (`parseProps` is a bare `JSON.parse`), so today:

- A `BigInt` prop throws an unhandled `TypeError` at hydrate.
- A `Map`/`Set`/`WeakMap`/`WeakSet`/`Promise` prop silently de-serializes to `{}`/`null` — no diagnostic anywhere in the pipeline.

## Root cause

`checkRichTypeMethodCalls` (BF021) only looks for method-call sites. A prop that's read (e.g. passed into a closure, or just referenced) without ever having a method called on it was invisible to it — the exact "silent passthrough" hole #2273 closed for the method-call shape still existed for the plain-read shape.

## Fix

Two layers, per the recommended design:

1. **Compile-time (adapter-neutral):** `checkRichTypePropSerialization` (`packages/jsx/src/rich-type-refusal.ts`) — BF021's sibling. It reuses the same `resolvePropDeclaredType` evidence chain BF021 already builds, restricted to `JSON_UNSAFE_RICH_TYPE_NAMES` (host rich types with **no** JSON-safe revival path — `Date`/`URL` are excluded since `new Date(json)`/`new URL(json)` round-trip). It refuses any prop in that set that `clientAnalysis.usedProps` says the component's own client code reads. The check mirrors the Hono adapter's own `propsToSerialize`/`hasPropsToSerialize` filter (`on*` handlers and `__`-prefixed names excluded) so it fires exactly where the adapter would actually attempt serialization.

2. **Runtime backstop (Hono only):** compile-time proof is necessarily conservative — a loosely-typed prop can still be a `Map`/`BigInt` at runtime with no static evidence. `serializeHydrationProps` (`packages/adapter-hono/src/utils.ts`) throws a clear `TypeError` — naming the prop, the component, and linking BF049's docs — for `BigInt`/`Symbol`/`Map`/`Set`/`WeakMap`/`WeakSet`/`Promise` at the top level of the hydration payload.

`RegExp`/`Error`/`URLSearchParams` are **deliberately excluded** from the runtime throw set: they degrade to a non-empty, tolerated shape rather than losing all data, and the site's own InputOTP demo knowingly passes a live `RegExp` prop today — throwing for them would be a breaking behavior change outside this issue's scope.

## Verification

Full suite reruns, clean across all five affected packages:

| Package | Result |
|---|---|
| `packages/jsx` | 3082 pass / 0 fail |
| `packages/adapter-tests` | 2006 pass / 0 fail |
| `packages/adapter-hono` | 1466 pass / 0 fail |
| `packages/client` | 627 pass / 1 fail (pre-existing #1247 timeout, unrelated — reproduces green in isolation) |
| `packages/compat` | 481 pass / 0 fail |

New fixtures `rich-prop-client-read` (the refusal) and `rich-prop-precompute` (the sound escape — pass a JSON-safe entries array, reconstruct the `Map` client-side) are pinned BF049 across all nine adapters. `coverage-map.json` and `ui/compat.lock.json` regenerated to match; `generated-data-points.json` and `support-matrix.lock.json` confirmed already fresh (no diff). `site/core/landing/components/shared/demo-outputs.ts` regenerated (it embeds Hono's serialization line verbatim).

Docs (`docs/core/advanced/error-codes.md` BF049 section + quick-reference row, `spec/compiler.md` error table row, `spec/subset-conformance.md` passthrough-closure note) updated to match.

## Design process

Per this session's standing instruction, design/investigation was delegated to and verified by a `fable` sub-agent: it empirically reproduced both failure modes, corrected the issue's own draft rationale (RegExp/Error don't merely "degrade" — they lose all data, a distinction that mattered once the InputOTP counter-example surfaced), and recommended the two-layer compile-time + runtime approach along with the runtime throw-set scoping. Implementation, testing, and full-suite verification were done directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_011Y7832bD8Ph4NHfq7B42kb)_


kfly8:
Thanks for the review — both points check out and are fixed in 03ff19db0:

1. **"No method call involved" was wrong.** `checkRichTypePropSerialization` fires whenever `clientAnalysis.usedProps` contains the prop name, regardless of whether client code later calls a method on it — `checkRichTypeMethodCalls` only walks template-lowered expression positions, so a handler/effect body's method call (e.g. `data.get(...)`, which the `rich-prop-client-read` fixture's own source already used) is just as invisible to it as a bare read. Reworded every occurrence (module/function docstrings, `errors.ts`, all 9 `conformance-pins.ts` files, both changesets, `spec/compiler.md`, `spec/subset-conformance.md`, `docs/core/advanced/error-codes.md`, the fixture's `description` field, the test file header) to describe the actual condition instead.

2. **RegExp/Error/URLSearchParams rationale was wrong.** Verified: `JSON.stringify(/x/)` and `JSON.stringify(new Error())` both produce `{}` — identical data loss to `Map`/`Set`. Only `URLSearchParams` actually round-trips. The real reason they're excluded from the Hono runtime throw is compatibility (the site's InputOTP demo already ships a live `RegExp` prop that degrades to `{}` today), not data preservation. Fixed the docstring in `utils.ts` and both changesets to state that instead.

Also tightened the per-offender throw message in `serializeHydrationProps` (`Symbol`/`Promise` no longer say "contents dropped" — that phrasing didn't fit either).

No logic changed — full suite reruns clean (jsx 3082, adapter-tests 2006, adapter-hono 1466, compat 489, all 0 fail).

---

Issue #2649, opened 2026-08-17
CSR: grandchild-composition (3rd-level nested component) reuses parent's scope id instead of deriving its own


kfly8:
## Summary

The `grandchild-composition` CSR conformance fixture is skipped (`packages/adapter-tests/src/csr-skip-set.ts`) with no tracking issue — a gap in the three-piece-set convention (`CLAUDE.md`: fixture + `known-limitation` issue + pin, each carrying the issue URL). This issue fills that gap.

## The bug

At the third composition level (a grandchild component nested two levels deep), CSR reuses the parent's scope id (`test_s0`) instead of deriving its own (`test_s0_s0`, as SSR correctly does).

#2444 fixed the sibling case — a child component nested directly inside a dynamic loop row getting a random scope id on every adapter except Hono — but this **grandchild** case (composition depth 3, CSR only) is a distinct, still-open gap.

## Where it comes from

Deriving the correct id via `renderChild`'s `_parentScopeId` push collided with `comment: true` wrapper self-lookup — specifically `$cSingle`'s short-suffix fallback in `query.ts` — breaking hydration for a `renderNode`-style callback prop whose inner component's own first slot id coincides with the wrapper's slot number.

Discovered via a real regression in `site/ui`'s xyflow Highlight-Depth demo.

## Scope

CSR only — the fixture and skip live in `packages/adapter-tests/src/csr-skip-set.ts` (`grandchild-composition`). Not yet confirmed whether the SSR template-adapters (Go/ERB/Blade/etc.) have an equivalent gap; worth a quick audit when this is picked up, since #2444's fix covered SSR-side scope-id derivation for the shallower nesting case.

## Suggested fix direction

Investigate `renderChild`'s `_parentScopeId` push against `$cSingle`'s short-suffix fallback (`query.ts`) — the collision needs either a longer/disambiguated suffix for the grandchild derivation, or a fix to `$cSingle`'s lookup so it doesn't false-match a coincidentally-equal slot number from an unrelated wrapper.

## Acceptance

Fix the derivation (or the `$cSingle` lookup), delete the `grandchild-composition` entry from `csr-skip-set.ts`, and confirm the fixture passes CSR conformance without further changes.

Refs: #2444 (fixed sibling case, closed).

Issue #2648, opened 2026-08-17
Rich-typed prop hydration-transport refusal stays permanently unrevivable — BF049


kfly8:
## Summary

A prop typed as a JSON-unsafe host rich type (`Map`, `Set`, `WeakMap`, `WeakSet`, `URLSearchParams`, `RegExp`, `Promise`, `Error`, `Symbol`, `BigInt`, `Function`) refuses with **BF049** on all 9 adapters (including Hono) when the component's own client code (a handler, an effect) uses it — regardless of whether a method is called on it:

```tsx
'use client'
export function RichPropClientRead({ data }: { data: Map<string, number> }) {
  return <button onClick={() => console.log(data.get('x'))}>go</button>
}
```

Fixture: `rich-prop-client-read` (`packages/adapter-tests/fixtures/rich-prop-client-read.ts`), pinned identically via each adapter's `conformance-pins.ts` and surfaced on the [compatibility matrix](https://barefootjs.dev/docs/advanced/compatibility-matrix) render-honesty table.

## Why this is a permanent (by-design) limitation, not open work

The `bf-p` hydration boundary transports props as `JSON.stringify`d data with no type-aware revival (`parseProps` is a bare `JSON.parse`). `Map`/`Set`/`WeakMap`/`WeakSet` JSON-serialize to `{}`, losing all entries; `BigInt` throws in `JSON.stringify` outright; the rest degrade similarly. There is no sound escape for these types short of pre-computing a JSON-safe value server-side and reconstructing the rich value client-side (`rich-prop-precompute`'s twin fixture).

A general typed-prop revival mechanism (e.g. a `{ $map: [[k, v], ...] }`-style envelope) was explicitly considered and **rejected as out of scope** in #2642 — a 9-adapter serialization-protocol change with per-type revivability caveats (`WeakMap`/`WeakSet` can't be carried by any envelope at all; `Promise`/`Function` can't cross a serialization boundary ever, by construction). #2642 was closed `not_planned` in favor of the narrower, loud compile-time + runtime-backstop refusal implemented by #2643/PR #2647 (`checkRichTypePropSerialization`, BF049).

This mirrors #2356's own precedent for BF021's `toLocaleDateString()` refusal: the enabling/implementation work (#2643) is complete and closed, but the refusal it left behind is correct and is expected to remain forever absent a full revival-mechanism redesign.

## Why this tracker exists

[...]

Issue #2630, opened 2026-08-13
go-template: prop-backed child-component loop renders empty at SSR (blocks the prop-precompute escape for the #2321 with-component shape)


kfly8:
## What diverges

`static-array-from-props-with-component-precomputed` (added in #2629) compiles clean on go-template and `go run`s without error, but renders its loop host **empty**:

```
Expected: <ul bf-s="test" bf="s1"><span bf-s="Tag_*" … data-key="a">…a…</span><span … data-key="c">…c…</span></ul>
Received: <ul bf-s="test" bf="s1"></ul>
```

The fixture is a plain prop-backed loop whose body is a child component:

```tsx
export function TagList(props: { entries: { id: string; variant: 'on' | 'off' }[] }) {
  return <ul>{props.entries.map(entry => <Tag key={entry.id} id={entry.id} variant={entry.variant} />)}</ul>
}
```

Its sibling `static-array-from-props-precomputed` — identical except the loop body is inline elements rather than a child component — renders correctly on every adapter. So the divergence is specific to **child-component bodies over a prop-backed array**.

## Likely cause — possibly the harness, not the adapter

`render-divergences.ts`'s own `todo-app-ssr` note already describes this shape:

> `.TodoItems []TodoItemProps` — the loop-body CHILD COMPONENT slice the template actually ranges over — which has no server-side population path in this harness (documented as route-handler-populated in production). `buildDynamicChildLoopSeeding` (this package's `test-render.ts`) now replicates that documented contract for a **signal-backed dynamic** child-component loop.

That seeding covers the signal-backed dynamic case. A **prop-backed static** child loop appears to have no equivalent path, which would explain an empty range with no `go run` error. If that's the whole story this is a test-harness gap rather than emitted-template breakage, and the fix is a static analogue of `buildDynamicChildLoopSeeding` — worth confirming before assuming the adapter is at fault.

## Why it matters beyond the fixture

The BF101 diagnostic for the #2321 loop-source shape offers two escapes and lists **pass-as-prop first**, because it is the one that keeps full server output. #2629 added the twins that prove that claim. On go-template the claim is now proven for the plain shape but **not** for the child-component shape — the fixture is pinned in `renderDivergences` pointing here.

Until this is fixed, a user on go-template who follows the diagnostic's first suggestion for a child-component loop gets an empty container at SSR, with no error telling them so. That is the quiet half of the "loud-or-escapable" contract (#2613): the escape exists, compiles, and silently under-renders.

[...]

PR #2650, opened 2026-08-17
chore: strip graduated-fix changelog comments from adapter skip/pin declarations

kfly8:
## What

The adapter skip/pin declaration files — `conformance-pins.ts` (×9), `render-divergences.ts` (×8), `csr-skip-set.ts` — had accumulated ~350-400 lines of "no longer pinned / graduated (#NNNN)" changelog narration for entries that no longer exist in the file. Per `CLAUDE.md`'s comment philosophy (explain a non-obvious *current* constraint, not a history of past fixes), this PR removes all of it. Every comment on a still-live skip/pin is kept unchanged, since those explain the current reason a fixture is skipped, not history.

## What else this surfaced

Reading every file in full to separate live reasoning from dead narration turned up two real issues, not just stale comments:

1. **A factually wrong comment.** `adapter-go-template/src/render-divergences.ts`'s header claimed the file was empty ("the last live entry graduated"), but it has carried a live entry (`static-array-from-props-with-component-precomputed`, #2630) since #2629 landed. Fixed to a standing note that won't go stale next time an entry graduates.

2. **A pin pointing at a just-closed issue.** BF049's `rich-prop-client-read` pin (all 9 adapters) referenced #2643 — the enabling/implementation issue for BF049 itself, closed earlier today by PR #2647. Per this repo's own precedent (#2356: *"a permanent known-limitation should point at an open known-limitation tracker, not a completed enabling issue"*), filed **#2648** as that open tracker and repointed all 9 pins.

3. **A missing tracker.** `grandchild-composition`'s CSR skip (`csr-skip-set.ts`) carried real technical detail (the `$cSingle` short-suffix collision) but no issue URL at all — a gap in the fixture/issue/pin three-piece convention. Filed **#2649** and added the reference.

## Verification

No logic changed anywhere — every `Set` member, every pin object, and every `render-divergences` entry is byte-identical except the two issue-URL swaps above (confirmed via `git diff` filtered to non-comment lines). `ui/compat.lock.json` regenerated to match (the 9 URL swaps only, verified diff); `ui/support-matrix.lock.json` confirmed already fresh, zero diff.

**Authoritative check** (no real subprocess spawning, exercises `expectedDiagnostics`/`conformancePins` directly): `packages/adapter-tests` — **2006 pass, 0 fail**.

**Real-runtime adapter suites**: this session's container showed elevated failure counts when running all 9 real-runtime suites (blade/erb/go-template/jinja/mojolicious/rust/twig/xslate/hono) under concurrent load — traced to resource contention (a `cargo build --bin bf-render` killed with SIGTERM mid-compile) and PATH gaps for spawned subprocesses (`go`/`perl` reported "not found" inside test-spawned shells despite being on `PATH` interactively). None of it traces to this change: every failing test name (`toggle-shared`, `todo-app`, `checkbox`, `dropdown-menu`, `accordion`, `date-tolocale-named-tz`, claim-plan `pagination`/`data-table`, …) is unrelated to any file this PR touches, each takes 5-13s (a timing signature consistent with subprocess starvation, not a fast diagnostic assertion), and re-running individual adapters in relative isolation converges toward 0-2 failures. `adapter-blade`'s own suite, run in isolation, passed clean: **1395 pass, 0 fail**. `adapter-xslate` in isolation: 1394 pass, 1 fail (an unrelated slow hydration test, `toggle-shared`). `adapter-go-template` in isolation: 1674 pass, 2 fail (both unrelated — `dropdown-menu` portal timing, and a `go command not found` PATH gap on `todo-app-ssr`'s claim-plan check).

## Design process

Per this session's standing instruction, the cleanup plan (which comments to keep vs. drop, per file, line-by-line) was designed and verified by a `fable` sub-agent that read every affected file in full and cross-checked every cited issue number against its live GitHub state. Implementation (mechanical text edits), the two new issue filings, and full-suite verification were done directly.

---
_Generated by [Claude Code](https://claude.ai/code/session_011Y7832bD8Ph4NHfq7B42kb)_

### DIFF
--- a/packages/adapter-blade/src/conformance-pins.ts
+++ b/packages/adapter-blade/src/conformance-pins.ts
@@ -52,24 +52,12 @@ export const conformancePins: ConformancePins = {
   // verbatim (and patches the region on same-key updates, #2389), a DSL
   // adapter refuses with BF021 + `/* @client */`. See spec/callback-fidelity.md.
   'preamble-cells': [{ code: 'BF021', severity: 'error' }],
-  // `todo-app` / `todo-app-ssr` no longer pinned (#2205) — the conformance
-  // harness now passes `siblingTemplatesRegistered: true` for fixtures with
-  // sibling `components`, matching `bf build`'s real semantics, so the
-  // BF103 loop-body cross-template check no longer fires spuriously. (Both
-  // fixtures are still skipped on this adapter via `render-divergences.ts`
-  // — #2209 — for an unrelated signal-seeding gap.)
-  // `static-array-children` no longer pinned (#2208) — `items`'s
-  // array-literal initializer is now recognized as fully-static
-  // (`resolveStaticLoopSource`) and inlined as a native PHP array literal
-  // in the `@foreach` header, the same way a module-scope const's value is
-  // already seeded.
   // The `([emoji, users]) => …` array-destructure param itself now lowers
-  // (#2087 Phase B — see the destructure comment below), but the loop
-  // ARRAY is a function-scope computed const (`const entries =
-  // Object.entries(props.reactions ?? {}).filter(...)`) that the adapter
-  // can't bind as a template variable — refused loudly with BF101 (same
-  // check and policy as Jinja / ERB) instead of silently iterating zero
-  // times over an unbound name.
+  // (#2087 Phase B), but the loop ARRAY is a function-scope computed const
+  // (`const entries = Object.entries(props.reactions ?? {}).filter(...)`)
+  // that the adapter can't bind as a template variable — refused loudly
+  // with BF101 (same check and policy as Jinja / ERB) instead of silently
+  // iterating zero times over an unbound name.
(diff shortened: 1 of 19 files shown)


kfly8:
CI red on the Twig/Blade adapter checks and `deno-check` — all three are hitting `HTTP/2 429` from `codeload.github.com` (GitHub's own archive-download CDN), not a code issue: Composer downloads for unrelated PHP packages (symfony/translation, illuminate/support, etc.) and a completely unrelated `denoland/setup-deno` action download are failing the same way. This is external GitHub infrastructure rate-limiting, unrelated to this diff (which touches zero PHP/Deno files). Retried once already (still rate-limited); will retry again once it clears and re-check.

Also added the `no-changeset` label — the `check` gate is otherwise correct that this touches published package source, but the change is comment-only with zero behavioral/API impact, so no changeset applies.

---

Issue #2751, opened 2026-08-28
Pure CSR mount throws `ReferenceError: val is not defined` — the module-scope `template` thunk references a component-scope signal binding (getter-elided form)


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. Six cases, **`csr-mount` leg only** — SSR + hydration is unaffected.

## Source (representative case)

`state-getter-elided-signal__structure-fragment__binding-class__event-direct-handler__callback-function-reference`:

```tsx
"use client";
import { createSignal } from "@barefootjs/client";
export function PairwiseCase() {
    const [val, setVal] = createSignal(0);
    const handleClick = () => { setVal(val() + 1); };
    return (<>
      {<span className={Boolean(val) ? "on" : "off"}></span>}
      <button onClick={handleClick} data-pw-event="1">go</button>
    </>);
}
```

`Boolean(val)` is the **getter-elided** form — the signal referenced without calling it, which the compiler is expected to resolve to the signal's value.

## Emitted client JS

```js
export function initPairwiseCase(__scope, _p = {}) {
  // …
  const [val, setVal] = createSignal(0)          // ← line 8, INSIDE init
  const handleClick = () => { setVal(val() + 1); }
  // …
}

hydrate('PairwiseCase', {
  init: initPairwiseCase,
  template: (_p) => `<span ${(`${Boolean(val) ? 'on' : 'off'}`) != null ? 'class="' + escapeAttr(`${Boolean(val) ? 'on' : 'off'}`) + '"' : ''}></span><button data-pw-event="1" bf="s0">go</button>`,
  comment: true,
  fragmentRoot: true,
})
```

The `template` thunk is at **module scope** and closes over nothing, but its body references `val`, which is declared inside `initPairwiseCase`. `createComponent()` (pure CSR mount) invokes this thunk, so it throws `ReferenceError: val is not defined` and the component mounts to an empty body. Hydration never invokes the thunk — the SSR markup is already in the document — which is exactly why the crash is csr-mount-only.

Two things are wrong here and it is worth deciding deliberately which is the fix (they are separable):

1. The getter elision that was applied inside the init body (`val()` on lines 8–10) was **not** applied inside the template thunk, which kept the bare `val`.
2. Independently of elision, a module-scope thunk should not be able to reference a component-scope binding at all. Even lowered to `val()` it would still be out of scope there.

## Observed

[...]

Issue #2750, opened 2026-08-28
Emitted client JS throws `ReferenceError: handleMount is not defined` — a `ref` const used inside a nested loop has its call site emitted but its declaration dropped


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. Case id:

```
state-signal__structure-nested-loop-depth-2__binding-boolean-attr__event-ref-callback__callback-function-reference
```

**This throws in the normal SSR + hydration path, not only in pure CSR mount.** Both legs error.

## Source

```tsx
"use client";
import { createSignal } from "@barefootjs/client";
export function PairwiseCase() {
    const [val, setVal] = createSignal(0);
    const [items, setItems] = createSignal([
        { id: 1, label: "Alpha", n: 3, active: true, children: [{ id: 10, label: "Alpha-child" }] },
        { id: 2, label: "Beta",  n: 1, active: false, children: [{ id: 20, label: "Beta-child" }] },
        { id: 3, label: "Gamma", n: 2, active: true, children: [{ id: 30, label: "Gamma-child" }] },
    ]);
    const handleMount = el => {
        el.setAttribute("data-mounted", String(val()));
    };
    return (<ul>
      {items().map((row, idx) => (<li key={row.id}>
          {row.children.map((child, cidx) => (<span key={child.id} ref={handleMount} data-pw-event="1">{<input type="checkbox" disabled={!Boolean(val())}/>}</span>))}
        </li>))}
    </ul>);
}
```

## Emitted client JS

The inner-loop row body emits the ref **call site**:

```js
{ const __rf_s1 = qsa(__innerEl1_0, '[bf="s1"]')
if (__rf_s1) (handleMount)(__rf_s1) }
```

but `handleMount` is **never declared anywhere in the emitted module**. `initPairwiseCase`'s preamble emits `val` and `items` and stops:

```js
export function initPairwiseCase(__scope, _p = {}) {
  if (!__scope) return
  const __scopeId = __scope.getAttribute('bf-s')

  const [val, setVal] = createSignal(0)
  const [items, setItems] = createSignal([ /* … */ ])

  const [_s3] = $(__scope, 's3')
  // … no `handleMount` declaration, ever
```

So the module throws `ReferenceError: handleMount is not defined` the moment the first inner row renders.

## Observed

Scanning all 85 `status: 'ok'` cases × both host legs for page errors, this case is the only one that throws in **both**:

```
[...]

Issue #2749, opened 2026-08-28
`ref` on a child-component call site is stringified into a `ref="…"` DOM attribute by hydration (source text leaks into markup, callback never runs)


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg (first run). Case id:

```
state-signal__structure-child-component__binding-class__event-ref-callback__callback-function-reference
```

## What is measured

Component under test (as generated by `packages/adapter-tests/pairwise/compose.ts`):

```tsx
"use client";
import { createSignal } from "@barefootjs/client";
function PairwiseRow({ children, ...rest }: { children?: unknown; [key: string]: unknown }) {
    return <div className="child" {...rest}>{children}</div>;
}
export function PairwiseCase() {
    const [val, setVal] = createSignal(0);
    const handleMount = el => {
        el.setAttribute("data-mounted", String(val()));
    };
    return <PairwiseRow ref={handleMount} data-pw-event="1">{<span className={Boolean(val()) ? "on" : "off"}></span>}</PairwiseRow>;
}
```

**SSR markup** (read straight off the generated artifact, not off an oracle diff — no `ref` attribute, as expected):

```html
<!--bf-scope:PairwiseCase_test--><div class="child" data-pw-event="1" bf-s="PairwiseCase_test_s1" bf-h="PairwiseCase_test" bf-m="s1" bf-r=""><span class="off" bf="^s0"></span></div><!--bf-/scope:PairwiseCase_test-->
```

**After hydration** (`snap` oracle, SSR vs hydrated):

```html
<div bf-s="PairwiseCase_*_s1" class="child" data-pw-event="1"
     ref="el =&gt; { el.setAttribute(&quot;data-mounted&quot;, String(val())); }">
  <span bf="^s0" class="off"></span>
</div>
```

So hydration writes the **source text of the callback** into a live `ref` DOM attribute. Two separate consequences:

1. Function source leaks into the rendered markup.
2. `handleMount` is never invoked — the element gets no `data-mounted`, which is the whole observable point of the `ref`.

Fails `snap` and `three-point`. Three cases in the covering array hit it (all `structure: child-component` × `event: ref-callback`).

## Why this is a contract violation, not a capability gap

`jsx-to-ir.ts` already states the invariant for element call sites, and enforces it there — `packages/jsx/src/jsx-to-ir.ts:6109`:

[...]

Issue #2745, opened 2026-08-28
`select` fails the idempotence oracle intermittently in CI


kfly8:
## What

`[idempotence] select: replayed actions agree between hydrated and csr-mount` (`e2e/oracle.playwright.ts:148`) fails in CI intermittently, on commits that do not touch the `select` path. It is not tracked anywhere, and each occurrence currently looks like a fresh, unexplained red.

Example: run [33152507825](https://github.com/piconic-ai/barefootjs/actions/runs/33152507825/job/98787558163) — `1 failed, 1 skipped, 162 passed`, the failure being this single test, on a commit whose diff is eight adapter `data-key` conditions and a generated JSON artifact.

## Why this is filed as a bug rather than dismissed as a flake

"Flaky" is not a diagnosis, and the failure is not an infrastructure death — the test body runs and the oracle's own comparison fails. The idempotence oracle asserts that the DOM after a replayed interaction sequence agrees between the hydrated leg and the pure-CSR-mount leg (`assertSnapshotsAgree`, `oracle-core.ts:109`). A comparison that passes most of the time and fails sometimes means the two legs are **genuinely diverging under some timing**, which is exactly the class this oracle was built to catch (#2481, oracle 3, the #2466 class).

The honest reading is one of:
- a real race between the two legs for `select` (portal-mounted listbox + controlled selection state), or
- a harness timing gap — `replayActionsAndCapture` waits one animation frame after `goto` and bounds each step at 5s, which may not be enough for the portal path under CI load.

Both are worth knowing; neither is "ignore it".

## Observations so far

- **Does not reproduce locally.** All six `select` oracle tests pass in isolation:
  ```
  ✓ [snap] select-native-shared / select
  ✓ [three-point] select-native-shared / select
  ✓ [idempotence] select-native-shared / select   (869ms)
  6 passed (5.1s)
  ```
  Repeated isolated runs on an earlier commit were 3/3 clean, plus a clean full-suite run.
- **Predates the changes it lands on.** It was observed independently before the fixes in #2744, on a branch that touched none of the select machinery.
- The neighbouring `select-native-shared` fixture has not been seen failing — only `select`, the portal-listbox one.

## Suggested next step

[...]

PR #2752, opened 2026-08-28
Browser-oracle leg for the pairwise sweep (#2481 step 5)

kfly8:
Runs #2481's three DOM-state oracles (`snap` / `three-point` / `idempotence`) against every `status: 'ok'` case the t=2 covering-array generator produces, on a nightly schedule mirroring `mutation-sweep.yml`. The compile leg landed in #2744/#2747; this is the half that actually executes the cases in a browser.

Unlike the mutation sweep there is no `baseAlreadyQuarantined` analogue — a pairwise case is synthesized from an axis tuple and has no base fixture to inherit a known failure from, so every failure here is new. `pairwise-quarantine.ts` ships **empty** on purpose: the first sweep is triage input, not something this PR quarantines on its own initiative. #2759 stacks on this and fills it.

## What the first sweep found

**145 passed / 111 failed** of 256, stable across a contended and an uncontended run.

The interesting part was the 39 `TimeoutError: locator.click` failures, which arrived looking like one undifferentiated pile of harness noise. They were three unrelated populations:

| n | what | disposition |
|---|---|---|
| 10 | render no click target at all — `conditional-ternary`/`early-return` seeded falsy select the row-less branch | skipped, driven by the shared `LEGITIMATELY_ROW_LESS` table |
| 9 | target exists but has **zero height** — a binding whose only content is an attribute leaves its element with no line box | fixed by a `hostStyles` floor size |
| 20 | target **absent from the `csr-mount` leg entirely** | product defects — left failing, filed as #2754 |

That last row is the reason this was worth pulling apart. Twenty of the 39 were never a harness problem at all, and the shape that gave it away was structural rather than statistical: 19 of 20 land on `child-component` / `component-row-root-loop` / `fragment` while spreading evenly across every `binding` and `state` value. Starvation scatters; this didn't.

## Issues filed from this sweep

Triage is complete — every one of the 111 failures now has a named bucket, with nothing unexplained. Ten issues, populated into the ledger by #2759.

- **#2755** — getter-elided signal reaches the controlled-value effect, the loop row builder, *and* class-expression sites uncalled. In scope, so unlike #2751 it doesn't crash: it stringifies the accessor. A live input's `.value` becomes the source text of `@barefootjs/client`'s internal `read()`; one case bakes `"() => 0"` into the **SSR** markup. 16 generated modules carry an uncalled `String(val)`. Visible only to `diffDomState` — no serialized-HTML comparison can see a wrong DOM *property*.
- **#2753** (15 cases) — hydration stamps `data-key` on rows SSR never emitted: an *index* key on an unkeyed loop, and a second key beside the depth-suffixed `data-key-1` on nested rows. Verified **not** a regression from #2744 — at the pre-merge commit the SSR HTML and generated client JS are byte-identical, and that fix took the same 48 tests from 4→7 passing.
- **#2754** — caller-supplied props on a child-component root are dropped under `createComponent`. `class` (written inside the child) and `data-key` (runtime-generated) survive; only the call-site prop vanishes.
- **#2714** — corroborated with a synthesized minimal case rather than only the three hand-written UI components it was filed from.
- **#2751** — pure CSR mount throws `ReferenceError: val is not defined`: the module-scope `template` thunk references a signal binding declared inside `init`. For 4 of its cases #2755 fires first and masks this crash, so fixing #2751 alone will leave them failing.
- **#2756** — controlled `value` diverges between hydrate-reused and freshly-built loop rows after a row-count change, and the direction flips by element type. Confirmed **not** #2745: no portal involved, and 36 re-runs showed zero intermittency.
- **#2749** — `ref` on a child-component call site is stringified into a live `ref="el => {…}"` DOM attribute. Source text leaks into the markup and the callback never runs. `jsx-to-ir.ts:6109` states the invariant ("ref … never appears in the rendered HTML") and enforces it for element call sites; it doesn't hold at a component call site.
- **#2757** — a top-level root-is-a-child-call component never threads `_parentScopeId`, so a nested `renderChild` names its `bf-s` after itself instead of the SSR-established `Caller_sN` convention. Distinct from #2731, which is a compile-time flag on an if-statement-rooted `ComponentDef`.
- **#2758** — a controlled `select` whose bound value matches no option: SSR emits no `selected` attribute and the browser defaults to the first option, while hydration assigns `.value` explicitly and gets no selection. See the correction below.
- **#2750** — emitted client JS throws `ReferenceError: handleMount is not defined`. A `const` arrow used as a `ref` inside a nested loop gets its call site emitted and its declaration dropped. **Throws in the SSR + hydration path, not only pure CSR.**

Neither crash, and neither of the two silent-property defects, was reachable from any single-feature fixture — which is the argument for the t=2 layer existing.

## Correction: the "empty props" attribution below was wrong

An earlier revision of this body closed with an open question about hydrate mode always mounting with `_p = {}`, and attributed one unexplained failing case to it. **That attribution is wrong**, and the case has its own root cause as #2758. Measured:

```
PRE  (SSR markup, no JS):  selectedIndex=0,  value="0",  bf-p={"val":7}
POST (after hydration):    selectedIndex=-1, value=""
```

`bf-p` carrying the prop **is** present in the served DOM, and the two legs diverge identically whether the prop arrives or not, because `String(String(7))` and `String(String(undefined))` both match no `option`. So empty props neither cause nor explain it.

Leaving the record of the mistake rather than quietly deleting it: the case was the last of the 111 to be attributed, and the pressure to fold it into a plausible-sounding neighbouring cause is exactly what a quarantine ledger must not encode, since an entry is read as a diagnosis.

Whether hydrate mode *should* reconstruct `_p` from `bf-p` remains a real design question about what the sweep measures — it is just not this case's answer, and nothing currently observed depends on it.

## A detection gap this exposed, not fixed

All three oracles compare serialized DOM plus a live DOM-state table. **A component whose init throws just produces different DOM** — so it lands in the failure list as an ordinary structural divergence, indistinguishable from a mis-rendered attribute. #2750 and #2751 were only found because I scanned all 85 cases × both legs for page errors as a separate probe, not because an oracle flagged them.

Two masking situations make the same point from the other side: #2755 masks #2751 on 4 cases, and #2749 masks #2754 on 3. Both are recorded in the ledger's `reason` fields by #2759.

A page-error assertion in the oracle legs would catch this whole class directly. Still deliberately not in this PR: it would reclassify failures, and it is not a prerequisite for the ledger now that triage is complete. Tracked on #2750.

## Notes on the harness changes

The `hostStyles` rule is tag-qualified, and every qualification was forced by measurement rather than chosen up front:
- a blanket `[data-pw-event]{display:inline-block}` flips `li` elements out of `list-item` flow, collapsing sibling rows onto one line so one covers another's click point;
- a `1px` floor makes adjacent `fragment-row-loop` rows share an edge that Playwright's hit-test resolves to the **wrong sibling**;
- the same `1px` floor on `div` targets does it again on the vertical axis — raised in review, and measurement found it already live on two `component-row-root-loop` cases, where row 0's click point resolves to row 1.

A click that silently lands on the wrong element is worse than the timeout it replaces, because nothing reports it.

That third one also exposed a hole in this PR's own backstop: the floor test asserted non-zero bounding-box **area**, and a 1px-tall row root has area 1264 — so it passed on both broken cases. Area was a proxy. It now asserts the property that matters (`document.elementFromPoint` at the click centre resolves to the target or a descendant) and names the offending sibling on failure. Perturbation-checked both directions.

CSS is safe here because every oracle compares DOM, never computed style or geometry, and `hostStyles` applies identically to all three host modes so it can't introduce an asymmetry between compared legs. The floor test covers only the `hydrate` leg on purpose: the `csr-mount` absences are product defects with their own issues, and a geometry floor that can only be satisfied by fixing the compiler is not a geometry floor.

`LEGITIMATELY_ROW_LESS` moved to `pairwise/axes.ts` so the contract test and the sweep read one table. The skip asserts the SSR HTML genuinely contains no target, so a combo that starts rendering one fails loudly rather than staying quietly skipped — and it is deliberately *not* "skip when the selector is missing", which would also swallow a row that should have rendered and didn't.

## Verification

- `bun run pairwise:generate` — 97 cases, ok=85 refused=12 broken=0
- full sweep — 145 passed / 111 failed / 20 timeouts, reproduced on a quiet box and again after the `div` floor fix
- `bun test packages/adapter-tests/src` — **2192 pass, 4 skip, 0 fail** (634s)
- typecheck: `bunx tsgo --noEmit -p packages/adapter-tests/tsconfig.json` reports an identical **554** errors with and without these edits (dominated by `TS5097` and `bun:test` resolution failures — artifacts of invoking it outside the workspace build, not real repo state). Zero non-module-resolution errors in the touched files.
- biome does **not** lint this tree — `biome.jsonc`'s `files.includes` carries `!packages/adapter-tests/**`, and the pre-commit hook confirms it with "Checked 0 files".

Not on per-PR CI by design (#2481's budget); nightly at `0 5 * * *` plus `workflow_dispatch`.

## Next

#2759 populates the ledger (111 entries, sweep to 0 failed). After that, the conformance fixtures the ten issues need — the pairwise corpus is generated and gitignored, so it is a detector, not durable regression armor.

---


kfly8:
Good catch, and the answer to "verified rather than just not yet triggered" is: **not yet triggered, and already wrong.** Fixed in 1d1a8e8d7.

Measured Playwright's actionability criterion directly — `document.elementFromPoint` at each target's click point, across all 85 `ok` cases:

```
PROBED 85 cases
WRONG-TARGET: 2
  state-memo__structure-component-row-root-loop__binding-style__event-direct-handler__callback-function-reference   h=1 w=1264 WRONG(->DIV sibling#1)
  state-prop__structure-component-row-root-loop__binding-class__event-handler-reading-outer-signal__callback-inline-arrow  h=1 w=1264 WRONG(->DIV sibling#1)
```

Exactly the shape you described: three `component-row-root-loop` row roots at 1px tall, stacked edge-to-edge, and row 0's click point resolves to row 1. So the `<div>` rule now takes the same `1em` as the span rule.

Your reasoning about why it didn't surface as an oracle failure holds — those cases' handlers don't mutate distinguishable per-row state, so the mis-hit was silent. That is what makes it worth fixing rather than shrugging at: a click landing on the wrong element while the suite stays green is strictly worse than the timeout the rule replaced, because nothing reports it.

**The more useful half of your finding is what it says about the floor test.** It asserted non-zero bounding-box area, and `1264 × 1` is non-zero — so the backstop written specifically to prevent this class passed on both cases. Area was a proxy for the property I actually cared about. It now asserts the real property (the click point hit-tests to the target, or a descendant) and names the offending sibling on failure:

```
Error: 2 case(s) have a [data-pw-event] target the click cannot reach:
  …component-row-root-loop__binding-style…: click point resolves to sibling #1 (box=1264x1)
  …component-row-root-loop__binding-class…: click point resolves to sibling #1 (box=1264x1)
```

Perturbation-checked both directions: red at `min-height:1px` naming both cases, green at `1em`. Full sweep unchanged at 145 passed / 111 failed / 20 timeouts, so the taller div floor doesn't perturb any other case's layout.

On the pinned-SHA note in your footer: that's the repo's convention for third-party actions across every workflow here (`pairwise-sweep.yml` mirrors `mutation-sweep.yml` exactly, including its pinning style), so I've left it as-is rather than diverging in one file. Worth raising separately if the cleanup-step freeze is a real problem.

---

Issue #2758, opened 2026-08-28
Controlled `select` whose value matches no option: SSR shows the first option, hydration shows none


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. The last of the 111 first-sweep failures to be attributed — it was deliberately left out of the quarantine ledger until it had its own root cause rather than being folded into a neighbouring issue.

Case: `state-prop-shadowing-signal__structure-early-return__binding-controlled-select__event-direct-handler__callback-function-reference`, failing `snap` and `three-point` on `DOM state diverges`.

## Measured

```
PRE  (SSR markup, no JS run):  selectedIndex=0,  value="0",  bf-p={"val":7}
POST (after hydration):        selectedIndex=-1, value=""
```

SSR:

```html
<div style="display:contents" bf-s="PairwiseCase_test" bf-r="" bf-p="{&quot;val&quot;:7}">
  <div bf-c="s2" data-pw-event="1" bf="s1">
    <select bf="s0"><option value="0">Zero</option><option value="1">One</option></select>
  </div>
</div>
```

The signal is seeded from a prop whose value (`7`) matches neither option. SSR therefore emits **no `selected` attribute at all**, and the browser applies its own default — the first option. Hydration then runs the controlled-value effect, which assigns the property explicitly:

```js
const __val = String(String(val()))
if ('value' in __ra_s0) { if (__ra_s0.value !== __val) __ra_s0.value = __val }
```

Assigning a `select.value` that matches no option yields `selectedIndex = -1` and `value = ""`. So the two legs disagree: server says "Zero", client says nothing selected.

User-visible consequence: a server-rendered page shows the first option, and the moment hydration runs the field goes blank. Nothing warns.

## Not the harness's empty-props question, and not #2755

Both were plausible and both were checked:

- **Not a harness artifact.** `bf-p` carrying `{"val":7}` is present in the served DOM (measured above). And the outcome does not depend on it either way: `String(String(7))` is `"7"` and `String(String(undefined))` is `"undefined"`, and neither matches `option value="0"` or `"1"`. So whether hydrate mode reconstructs `_p` from `bf-p` or mounts with `{}`, this case diverges identically. An earlier note on PR #2752 attributed it to the empty-props path; that attribution was wrong and this issue supersedes it.
- **Not #2755.** This case's generated client JS calls `val()` correctly everywhere — it is not the getter-elided-uncalled pattern. Verified by reading the emitted module.

## The general shape

[...]

Issue #2757, opened 2026-08-28
Top-level root-is-a-child-call component never threads `_parentScopeId`, so a nested `renderChild` invents its own `bf-s` prefix under pure CSR mount


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. 3 `(caseId, oracle)` pairs, all `structure: child-component`, leg `hydrated vs csr-mount`:

- `state-memo__structure-child-component__binding-controlled-textarea…`
- `state-prop__structure-child-component__binding-attr…`
- `state-signal__structure-child-component__binding-style…`

## Measured

Orientation resolved against the on-disk SSR artifact, not inferred from the diff labels:

```html
SSR (.pairwise/*.html):  <div class="child" data-pw-event="1" bf-s="PairwiseCase_test_s1" bf-h="PairwiseCase_test" bf-m="s1" bf-r="">…
hydrated:                <div bf-s="PairwiseCase_*_s1" class="child" data-pw-event="1">…
csr-mount:               <div bf-s="PairwiseRow_3306gf_s1" class="child" …>…
```

SSR and hydrated agree; **`csr-mount` is the leg that diverges from both**. The scope id's prefix is the child's own display name (`PairwiseRow_…`) instead of the caller-derived convention SSR established (`PairwiseCase_…_s1`).

## Mechanism (traced, worth re-confirming before fixing)

`PairwiseCase` compiles with `comment: true, fragmentRoot: false` — the "root is a child call" shape. In `packages/client/src/runtime/component.ts`, `materializeComponent` deliberately leaves `scopeId = null` for that shape (its own docstring cites #1211/#2649), and only sets `_parentScopeId` when `scopeId` or `slot?.parent` exists (~L331-334). A **top-level** `createComponent('PairwiseCase', {})` has no `slot`, so neither applies and `_parentScopeId` stays `undefined` while the template runs.

`renderChild` then computes its prefix as roughly `(_parentScopeId && slotSuffix) ? _parentScopeId : \`${displayName}_${generateId()}\`` (~L653-655). With `_parentScopeId` falsy it takes the fallback and names itself after the child.

Hydration is unaffected because it reads the scope id off the existing SSR markup rather than generating one.

## Not #2731

[...]

Issue #2756, opened 2026-08-28
Controlled `value` diverges between hydrate-reused and freshly-built loop rows after a row-count change — direction flips by element type


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. 5 `(caseId, oracle)` pairs, all involving a keyed-loop row **removal** (a filter/flatMap that shrinks the row count):

- `state-signal__structure-preamble-builder-body__binding-controlled-textarea…` (three-point + idempotence)
- `state-prop-shadowing-signal__structure-nested-loop-depth-2__binding-controlled-textarea…` (idempotence)
- `state-memo__structure-signal-array-loop__binding-controlled-input…` (idempotence)
- `state-memo__structure-fragment-row-loop__binding-controlled-select…` (idempotence)

## Not #2745

#2745 is the real portal-based `select` component (`bf-po`, `oracle.playwright.ts`) failing *intermittently* in CI. These are synthetic pairwise cases, none involve a portal, and re-running all 4 underlying cases × 3 oracles × 3 repeats (36 runs) showed **100% stable failure, zero intermittency**. Different file, different component, different mechanism.

## Two sub-mechanisms, opposite directions

Both are faces of one gap: a row *rebuilt* by client code does not carry the same value representation as a row *reused* from SSR. Which leg ends up with the stray or missing attribute depends on which of two value-binding code paths the element type compiles to.

**textarea / select — rebuilt rows gain an attribute SSR never emits.** The row builder embeds the value into an HTML *template string* parsed via `.innerHTML`:

```js
`<textarea ${cond ? 'value="' + … + '"' : ''}>`
```

so any freshly built row (full `csr-mount`, or a row rebuilt by the keyed-loop reconciler) gets a literal `value` attribute. SSR's renderer never emits that attribute for `textarea`/`select` — it encodes the value as child text or a `selected` option. A row *reused* by hydration keeps SSR's attribute-free markup.

**input — reused rows keep an attribute nothing removes.** The opposite. The static template is `<input bf="s0" />` with no value attribute, and a `createEffect` sets only the live **property** (`el.value = …`), never `setAttribute`. But the SSR adapter *does* bake `value="0"` as a literal attribute. Hydration reuses the SSR node, and a property write doesn't touch the attribute, so it becomes a permanent leftover that a freshly built `csr-mount` input never has.

## Why a row-count change is the trigger

[...]

Issue #2755, opened 2026-08-28
Getter-elided signal reaches the controlled-value effect and loop row builder uncalled — a form field's live value becomes the reactive core's `read()` source text


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg, via the `DOM state diverges` assertion (`diffDomState`, `e2e/dom-state.ts`) rather than the serialized-HTML one — which is why it was invisible to every structural comparison.

**This is a different emission site from #2751**, and #2751's fix will not close it. Same root defect (a `getter-elided-signal` reference not lowered to a call), three sites:

| site | where | symptom | tracked |
|---|---|---|---|
| 1 | module-scope `template` thunk | `val` out of scope → `ReferenceError`, csr-mount only | #2751 |
| 2 | controlled-value `createEffect`, non-loop structures | **in scope, so no crash** — stringifies the accessor | **this issue** |
| 3 | loop row builder / patcher (`createRow`/`applyOuter`) | same, and never reachable via #2751's thunk | **this issue** |

## Measured

Generated client JS for `state-getter-elided-signal__structure-child-component__binding-controlled-select__event-direct-handler__callback-inline-arrow`, line 17 — inside `initPairwiseCase`, so it runs without error:

```js
const __val = String(String(val))   // `val` is the accessor; never called
```

**16 of the generated modules carry an uncalled `String(val)`** (22 carry a correct `String(val())`).

The runtime consequence, straight from the oracle's DOM-state diff:

```
value: "" !== "() => {    if (Listener) {      if (profilerSink) {        subscribers.add(Listener);        subscri…
value: "() => 0" !== "() => {    if (Listener) {      if (profilerSink) {        subscribers.add(Listener);       …
selectedIndex: 0 !== -1
value: "0" !== ""
```

The live `.value` of an `<input>`/`<textarea>` becomes **the source text of `@barefootjs/client`'s internal `read()` function** (`Function.prototype.toString()` of the accessor). For `<select>`, that giant string matches no `<option>`, so `selectedIndex` goes to `-1`.

Note the second line: `"() => 0"` is on the **SSR** side — a memo's source text baked into the served markup.

If this reaches a real app, a form field contains the framework's internals and submits them.

## Affected

12 unique cases / 24 raw failures on the `SSR vs hydrated` leg. Of those:

[...]

PR #2759, opened 2026-08-28
Populate the pairwise-sweep quarantine ledger (#2481 step 5)

kfly8:
**Stacked on #2752** — base is `claude/barefootjs-spec-oracle-m3z7o3-s24`, not `main`. Merge that first.

#2752 landed the browser-oracle leg with `pairwise-quarantine.ts` shipping empty on purpose, because the first sweep is triage input rather than something to quarantine on its own initiative. Triage is now done, so this fills the ledger: **111 entries, one per failing `(caseId, oracle)` pair**, each carrying a `reason` and the issue it belongs to.

The sweep goes from 111 failed to **0 failed** — not because anything changed, but because every failure is now a recorded, attributed, rot-checked claim.

## Attribution

| issue | rows | what |
|---|---|---|
| #2753 | 32 | hydration stamps `data-key` SSR never emitted |
| #2755 | 22 | getter-elided signal reaches the value effect / row builder / class sites uncalled |
| #2754 | 20 | caller-supplied props dropped on a child-component root under `createComponent` |
| #2714 | 12 | ref callbacks never run at SSR |
| #2751 | 8 | `ReferenceError: val is not defined` in the module-scope template thunk |
| #2749 | 6 | `ref` stringified into a DOM attribute |
| #2756 | 5 | controlled `value` differs between hydrate-reused and freshly-built rows |
| #2757 | 3 | top-level root-is-a-child-call invents its own `bf-s` prefix |
| #2758 | 2 | controlled `select` with an out-of-range value: SSR shows option 0, hydration shows none |
| #2750 | 1 | `ReferenceError: handleMount is not defined` |

Counts are per `(caseId, oracle)` pair, so a case failing both `snap` and `three-point` contributes two rows — that is deliberate, since the ledger keys on the pair and rot-checks each independently.

## #2758 is new, and is why this took an extra pass

The ledger was first built with 109 entries and **one case left deliberately red**, because it fitted none of the nine existing issues and forcing it into a neighbouring one would have recorded a wrong diagnosis. That was the right call, and measuring it produced a tenth issue.

The attribution it nearly received was "the harness mounts hydrate mode with empty props" — which #2752's own body claimed. That is wrong, twice over:

```
PRE  (SSR markup, no JS):  selectedIndex=0,  value="0",  bf-p={"val":7}
POST (after hydration):    selectedIndex=-1, value=""
```

`bf-p` is present in the served DOM, and the outcome is identical whether the prop arrives or not, because `String(String(7))` and `String(String(undefined))` both match no `option`. The real defect is that SSR emits no `selected` attribute for an out-of-range value and lets the browser default to the first option, while the client assigns `.value` explicitly and gets no selection. A user sees the first option server-rendered, then a blank field the instant hydration runs.

#2752's body has been corrected accordingly.

## Two masking situations recorded in the reasons

A quarantine entry is read as a diagnosis, so where one bug hides another the `reason` says both:

- **#2755 masks #2751** on 4 cases — the DOM-state assertion fires before `assertSnapshotsAgree` reaches the csr-mount crash. Fixing #2751 alone will leave those pairs failing.
- **#2749 masks #2754** on the 3 `child-component × ref-callback` cases — the ref-stringification fails the SSR-vs-hydrated comparison before `three-point` ever reaches the leg where the prop-drop would appear.

Both were found by measurement rather than inferred, and the second was not anticipated.

## Verification

- regenerated the sweep and confirmed the same 111 failing pairs as the pre-ledger baseline (sorted pair lists diff identical)
- full sweep with the ledger: **256 passed, 0 failed, 0 stale-entry errors**
- rot-check proven live on two independent entries: removing one makes its pair fail with the genuine underlying error (not the "stale" message), restoring it makes it pass again. Verified on #2750's entry and, separately, on one of the #2758 entries added here:
  ```
  entry removed  → ✘ [snap] …early-return__binding-controlled-select…: DOM state diverges   (1 failed, 2 passed)
  entry restored → ✓ [snap] …                                                               (3 passed)
  ```
- `bun test packages/adapter-tests/src` — 2192 pass / 4 skip / 0 fail
- biome does not lint this tree (`!packages/adapter-tests/**`); the pre-commit hook reports "Checked 0 files"

## What this does not do

It does not fix anything. Every entry is a pointer to an open defect, and the ledger shrinks only as those are fixed — at which point the rot-check turns each stale entry into a loud failure telling the reader to delete it.

The conformance fixtures the ten issues need are still outstanding. The pairwise corpus is generated and gitignored, so it detects these defects but does not protect against their return; that is what the fixtures are for.

---


kfly8:
CI note — the `test` job failure on 9d59ea16c is not this PR's.

**Failing check:** `test` ([job 98963018535](https://github.com/piconic-ai/barefootjs/actions/runs/33204804328/job/98963018535)), one test:

```
(fail) JSX Conformance Tests > [data-table] site/ui DataTable demo — keyed-loop reorder on sort:
       column-header clicks resort the memo and the runtime reorders rows in place [1825.04ms]
 1447 pass / 22 skip / 1 fail
```

**Why it is not this PR's**, structurally rather than by inspection: this PR changes exactly one file, `packages/adapter-tests/e2e/pairwise-quarantine.ts`. That file has exactly one importer —

```
$ grep -rln "pairwise-quarantine" --include=*.ts packages/
  packages/adapter-tests/e2e/pairwise.playwright.ts
```

— and the `.playwright.ts` suffix exists precisely to opt that file out of `bun test`'s discovery (`playwright.config.ts:9-13`, `testMatch: '**/*.playwright.ts'`). The failing job runs `bun test`, which therefore never loads the changed file. There is no path by which this diff reaches an xslate DataTable sort test.

**Reproduced locally on this branch** — the same suite, including the failing test:

```
$ bun test packages/adapter-xslate/src
 1527 pass / 22 skip / 0 fail   (502s)
```

It also passed on the base branch's head (1d1a8e8d7 / #2752), where all checks were green.

So: green on the base, green locally on this head, and structurally unreachable from the diff. That leaves an environment-dependent failure in a click-and-reorder test. Re-running the job once, which is the single re-run this warrants; a second failure would be real and I would treat it as such rather than repeating this comment.

No fix ported because there is nothing to port — no existing PR addresses this, and it does not reproduce outside CI. If it recurs on a later head I will open a `known-limitation` issue for the flake itself with the runs that exhibited it, rather than leaving it to be rediscovered.

---

Issue #2760, opened 2026-08-28
BF044 refuses a getter passed directly to a component prop but allows the same getter wrapped in an object literal


kfly8:
Surfaced while measuring BF044's detection surface for #2755. It predates that work and is deliberately **not** being resolved inside it — recording it so it isn't silently settled in either direction.

## Measured

`compileJSX(src, 'C.tsx', { adapter: testAdapter })`, one signal `val`, one child component `Foo`:

```
DIAG BF044   <Foo x={val} />              ← refused
SILENT       <Foo value={{ val }} />      ← allowed
SILENT       <Foo value={{ open: val }} />← allowed
```

Both hand a live accessor to a child component. One is a compile error, the other is fine.

## Why it is inconsistent rather than merely strict

Passing accessors to children **is** this codebase's idiom, not an accident. `ui/components/ui/select/index.tsx:113-120`:

```tsx
<SelectContext.Provider value={{
  open,                                                    // bare accessor
  onOpenChange: (v) => { setOpen(v); … },
  value: () => isControlled() ? (props.value ?? '') : internalValue(),   // explicit thunk
  …
}}>
```

`open` sits directly beside an explicit `() => …` thunk: the contract of that object is "every member is an accessor, the consumer calls it at its own read site and subscribes there". Evaluating it eagerly at the provider would freeze the value at provider-render time and break reactivity for every consumer.

A blast-radius scan over `ui/components/**` (115 files), `site/**` (327 files) and `packages/adapter-tests/fixtures/**` (345 fixtures) found **66** instances of this shape and **zero** genuine forgotten-`()` bugs — `dropdown-menu`, `menubar`, `context-menu`, `select`, `carousel`, `navigation-menu`, `command`, `radio-group`, `sidebar`, `toggle-group`, `combobox`, and `chart` (~44 hits across five sibling components), plus two site demos.

So the pattern is deliberate and pervasive — and it survives the gate only because BF044 currently tests `if (!ts.isIdentifier(expr)) return`, i.e. only an expression's top-level node. `{{ val }}` is an `ObjectLiteralExpression`, so it returns early. **The idiom is legal by accident of nesting, not by design.**

## The question to settle

Either direction is defensible and both are subset decisions, not bug fixes:

- **Allow both.** Passing an accessor to a component prop is in subset; the child decides when to call it. This makes `<Foo x={val} />` legal, which is a permissive subset extension and needs fixtures per `spec/subset-conformance.md`.
[...]

Issue #2754, opened 2026-08-28
Caller-supplied props on a child-component root are dropped in pure CSR mount (present in SSR and after hydration, absent under `createComponent`)


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. **9 of 40 unique `(caseId, oracle)` structural divergences**, plus it is the direct cause of most of the sweep's 20 remaining `idempotence` click timeouts — the oracle replays its click against the `csr-mount` leg, where the target attribute no longer exists.

Fails the `hydrated vs csr-mount` half of `three-point`. SSR and hydration are correct; only `createComponent()` (pure CSR mount, no SSR markup) is wrong.

## What is measured

A static attribute passed **at the child-component call site** is present in SSR and after hydration, and absent when the same component is mounted via `createComponent`.

`state-signal__structure-child-component__binding-class__event-ref-callback__callback-function-reference`:

```html
<!-- hydrate leg -->
<div class="child" data-pw-event="1" bf-s="PairwiseCase_test_s1" bf-h="PairwiseCase_test" bf-m="s1" bf-r="">…</div>

<!-- csr-mount leg -->
<div bf-s="PairwiseRow_3306gf_s1" class="child" ref="…">…</div>
```

`state-signal__structure-component-row-root-loop__binding-controlled-input__event-handler-reading-loop-param__callback-flatmap-callback` (3 rows):

```html
<!-- hydrate leg -->
<div class="row" data-pw-event="1" bf-s="PairwiseRow_zpdgo9" bf-h="…" bf-m="s1" data-key="1"><input value="0" bf="^s0"></div>

<!-- csr-mount leg -->
<div class="row" bf-s="PairwiseRow_3m3y73" data-key="1"><input value="0" bf="^s0"></div>
```

Note precisely what survives and what does not:

| attribute | origin | csr-mount |
|---|---|---|
| `class="child"` / `class="row"` | written **inside** the child component | ✅ present |
| `data-key="1"` | generated by the runtime's loop reconciliation | ✅ present |
| `data-pw-event="1"` | passed by the **caller**: `<PairwiseRow data-pw-event="1">` | ❌ **dropped** |

So this is not "attributes are lost on CSR mount" generally — it is specifically the props supplied at the call site and destined for the child's root element.

Measured directly with a locator count in each leg (four representative cases):

```
child-component × ref-callback           hydrate: count=1  csr-mount: count=0
[...]

Issue #2753, opened 2026-08-28
Hydration stamps `data-key` on loop rows that SSR never emitted — unkeyed rows gain an index key, nested rows gain a second key alongside `data-key-1`


kfly8:
Found by the #2481 pairwise t=2 sweep's browser-oracle leg. **Largest single group in the first sweep: 15 of 40 unique `(caseId, oracle)` structural divergences.** Fails `snap` (hydration is not a no-op) and therefore `three-point`.

Two shapes, both a disagreement between the SSR adapters and the client runtime about the row-key attribute. Filing as one issue because the emission site is shared; split if the fix turns out not to be.

In both blocks below, `Received` is the SSR markup and `Expected` is the post-hydration DOM (that is the orientation `assertSnapshotsAgree` prints — it reads backwards, and has already caused two misfilings, see the Correction sections on #2714 and #2715).

## Shape A — unkeyed loop gains an *index* key

`state-signal__structure-fragment-row-loop__binding-text__event-direct-handler__callback-inline-arrow`

```
SSR:      <div bf-s="PairwiseCase_*" bf="s3"><span bf="s2" data-pw-event="1">…</span> ×3</div>
hydrated: <div bf-s="PairwiseCase_*" bf="s3"><span bf="s2" data-key="0" data-pw-event="1">…</span>
                                             <span bf="s2" data-key="1" …>
                                             <span bf="s2" data-key="2" …></div>
```

The values are `0,1,2` — the loop **index**, not any id. This structure's template carries no `key={}` at all (that is #2748: the pairwise `fragment-row-loop` template is unkeyed, which is why `carriesDataKey` is uncovered by it). So the client runtime is stamping a positional `data-key` onto rows of an unkeyed loop while every SSR adapter emits none.

Affected: `fragment-row-loop` and `nested-loop-depth-2`.

## Shape B — nested row gains a second key beside the depth-suffixed one

`state-signal__structure-nested-loop-depth-2__binding-boolean-attr__event-ref-callback__callback-function-reference`

```
SSR:      <span bf="s1" data-key-1="10" data-pw-event="1">…
hydrated: <span bf="s1" data-key-1="10" data-key="10" data-pw-event="1">…
```

SSR names the inner loop's key `data-key-1` (depth-suffixed); the client writes plain `data-key` with the same value, so the hydrated row carries both. Only the first inner row diverges in this case.

## Not a regression from #2744

[...]

Issue #2714, opened 2026-08-26
ref-effect-computed attribute state (aria-expanded/aria-checked/data-selected) never reaches SSR — compiled literals are baked and hydration corrects them


kfly8:
## Correction (2026-08-26)

The original report had the direction of the divergence **backwards** (a misread of the oracle diff's Expected/Received orientation). Measured directly (pre = SSR markup before any JS, post = after hydration):

- `accordion`: pre `aria-expanded="false"` → post `"true"`
- `radio-group`: pre `aria-checked="false"` → post `"true"`
- `command`: pre `data-selected="false"`, no `data-value` → post `"true"`, `data-value="Calendar"`

**SSR only ever emits the compile-time literal; hydration is the side that produces the correct value.** Not a hydration bug.

## Actual mechanism

`accordion`/`radio-group`/`command` write `aria-expanded="false"` etc. as hard-coded JSX literals and compute the real value inside `ref={handleMount}` (`useContext` + `createEffect`). `ref` callbacks never run during SSR — so the reference Hono SSR bakes the literal (confirmed in the accordion snapshot), and the first client effect pass fixes it. The accordion fixture's own comment documents this as intended: the interaction assertions pass only because hydration runs the mount effect.

## Status

Re-tiered `bug` → `known-limitation + enhancement`: the user-visible consequence (pre-hydration state differs from post-hydration state — a real snap at the hydrate boundary, which is exactly what the snap oracle flags) is an SSR capability gap, not a silent corruption. Closing it means either (a) rewriting these components to express initial state as compiler-analyzable JSX attribute expressions (like their sibling `data-state` attributes already do), or (b) an SSR-side mechanism to evaluate context-derived initial state. Until one lands, the `oracle-quarantine.ts` rows stay (with corrected reasons) as the executable record of the gap.

<details><summary>Original (incorrect) report</summary>

When a component's initial state renders a NON-default value for a reactive attribute at SSR time, hydration's first effect pass was believed to re-apply the attribute from the compiled default instead of adopting the SSR value. Measurement showed the opposite direction.

</details>

kfly8:
Independent corroboration from the #2481 pairwise t=2 sweep's first browser-oracle run.

[...]

PR #2761, opened 2026-08-28
Widen BF044 to descend into rendered positions (upstream of #2755 / #2751)

kfly8:
Family 1, the upstream cause of #2755 and #2751.

The pairwise sweep produced ten issues. Fixing them ten separate ways would be a pile of patches that the next sweep re-opens from another angle, so they were grouped by root cause first. #2755 and #2751 turned out not to be three emitter bugs but downstream symptoms of **one gate whose detection surface was too shallow**.

## Root cause

`checkBareSignalOrMemoIdentifier` (`packages/jsx/src/jsx-to-ir.ts`) — the gate behind BF044, "a signal/memo getter was passed without calling it" — opened with:

```ts
if (!ts.isIdentifier(expr)) return
```

It only ever inspected an expression's **top-level node**. A bare getter nested anywhere below slipped through, and each downstream emitter then broke differently: #2755 (the accessor stringified into a DOM property, so a form field contains the source text of `read()`), #2751 (a module-scope `template` thunk referencing an init-scope binding, so CSR mount throws `ReferenceError`).

## Why not simply recurse

The first plan was "make it recursive". Measuring the blast radius **rejected** that: scanning `ui/components/**` (115), `site/**` (327) and fixtures (345) produced **66 hits**, and every one was the Context Provider idiom:

```tsx
// ui/components/ui/select/index.tsx:113
 isControlled() ? (props.value ?? '') : internalValue(),  // explicit thunk
}}>
```

As the explicit `() => …` sibling shows, this context object's contract is "every member is an accessor". Evaluating them at the provider would freeze the values at provider-render time and break every consumer. Auto-fixing that is a correctness regression.

"Exempt object-literal property values wholesale" was measured and rejected too — it has a hole:

```
SILENT      emitted: style={{ color: val }}   ← no bf="s0", no reactive slot
```

No diagnostic and no reactive slot. Exempting that would permanently bless a real bug class.

## The rule adopted — recurse only into rendered positions

| position | behaviour |
|---|---|
| top-level bare identifier | fires everywhere, as before (**unchanged**) |
| nested bare identifier × element attribute value / JSX text child | **fires** (it gets rendered) |
| nested bare identifier × component prop | does not fire (it gets handed on) |

The discriminator is "rendered, or passed to a child". `getAttributeValue(attr: ts.JsxAttribute, …)` already receives the attribute node, so `attr.parent.parent.tagName` distinguishes element from component with no new state threaded through.

**Nothing that compiles today breaks.** All 66 Provider hits are component props and go silent; `` fires as before. Nothing was widened in the permissive direction, so no subset-extension fixture is required.

## Two by-products found while implementing

- **The walk has to stop at a JSX boundary.** Descending into an expression that returns JSX misreads a nested element's own **attribute name** as a bare reference (`checked=` in `checked={checked()}` fired). Stops at `JsxElement` / `JsxSelfClosingElement` / `JsxFragment`. Shadowing introduced *outside* the expression (loop rows, callback params) was also invisible, so both checks now go through the existing `BindingScope` service (`ctx.scope.isBound()`) — the same shape `csrSubstituteOnce`'s `enclosingScope` uses.
- **Three real bugs in the docs.** `docs/core/adapters/{custom,go-template}-adapter.md` had `tsx` fences writing `items.map(...)` where `items` is a signal (correct: `items().map(...)`). The shallow gate never looked inside a `.map()` callee, so these compiled silently down the static-array codegen path and emitted working code. Anyone copying the adapter-authoring docs was quietly writing broken components.

## Review round: three false positives in the widened walk

The first revision of this PR shipped the new rule with **no committed test**: the dedicated BF044 suite was untouched and nothing in the repo referenced the new walk. Review caught that, and closing it surfaced three defects — all in the **false-positive** direction, which is the one that refuses working code.

| fix | mechanism |
|---|---|
| **Visit binding and parameter defaults** | A default is a value expression, but the walk only looked at names. `const { x = count } = obj` in a rendered position emits a module-scope `template` thunk referencing a component-scope binding — `ReferenceError` on CSR mount. **#2751's mechanism, surviving inside the check meant to close it.** |
| **Match JS parameter binding order** | All defaults were visited before any parameter was bound. JS binds left to right — `((a, b = a) => b)(7)` returns `7`, verified against V8 — so `(count, x = count) =>` was flagged whenever a parameter name collided with a signal. Now one left-to-right pass, and the same rule inside a pattern (`{ a, b = a }`). |
| **Stop walking type positions** | `({} as { count?: unknown })` in a rendered position read the type literal's property name as a bare reference to a same-named signal and refused the component. Guarded with `ts.isTypeNode` (covers `as` / `satisfies`, type assertions, generic type arguments). |

A docstring was also corrected: it claimed `(count = count) => …` reads the outer `count`. It does not — that is the parameter's own binding in TDZ, `Cannot access 'count' before initialization`. A comment describing behaviour the language does not have was rewritten rather than kept as plausible-sounding rationale.

## Tests

`signal-getter-not-called.test.ts` gains a nested-descent block (31 pass total): positives (ternary, template-literal span, call argument, array-literal member, `style={{}}` property, JSX text child), negatives (component-prop nesting in three shapes, loop-row param shadowing, correctly-called getters), the JSX-boundary attribute-name guard, and the three false positives above.

**Every case is perturbation-verified** — the implementation was reverted and the corresponding tests confirmed to fail. That process caught one test that pinned nothing: the pattern-ordering case sat at a function body's top level, where `collectBlockDeclarations` pre-scans the whole `VariableStatement` and binds every name up front regardless of order, so removing the code under test left it green. It now sits inside an `if` block the pre-scan does not reach, and its default references a name that is actually a signal (a default reading a non-signal name is silent either way and pins nothing).

## Verification

| | |
|---|---|
| rule behaviour | 13 / 13 cases matched |
| blast-radius rescan | **0 hits** across `ui/components/**` / `site/**` / fixtures |
| `bun test packages/jsx/src` | 3157 pass / 28 skip / 3 todo / **0 fail** (3188 across 246 files) |
| BF044 suite alone | **31 pass / 0 fail** |
| `bun test packages/adapter-tests/src` | 0 fail |
| `pairwise:generate` | 97 cases, **ok 85→62 / refused 12→35 / broken 0** |
| pairwise sweep | **187 passed / 0 failed**, 0 stale |

`pruneUnusedPropExtractions: generated code did not parse; skipping prune` appears once during the run. It is a defensive warn path in `prune-unused-prop-extractions.ts` that exists on `main` as well, and no test fails on it.

## Quarantine ledger: 111 → 67

The sweep going 111 fail → 0 fail is a **pass** under sound-or-loud: 23 cases that were silently diverging are now loudly refused.

- **#2755: 22/22 deleted** (section gone)
- **#2751: 8/8 deleted** (section gone)
- #2753: 8/32 deleted; #2754: 4/20; #2714: 2/12

The rot-check fires on none of the surviving entries (the 187/0 is that confirmation).

## Asymmetry left alone

`` being refused while `` passes predates this change and is **not touched here** — it is a subset judgement, not something that should be inherited implicitly from a `ts.isIdentifier` early return. Filed separately as #2760.


kfly8:
レビュー指摘に2件対応した。

## 主指摘: 再帰降下に commit されたテストが無い — 対応済み (`2985e3089`)

そのとおりだった。門に272行足していながら BF044 専用スイートは無変更で、新しい walk を参照するものがリポジトリ内に1つも無かった。本文の「13/13」は promote しなかった開発時プローブで、規則は一度検証したきり何にも支えられていなかった。

`signal-getter-not-called.test.ts` に移植し、**perturbation で検証した**: 広げる前の門に対して positive 6件が落ち、negative は通ったまま。指摘の "Required outcome" に挙がっていた形（nesting positive / component-prop 免除 / ループ行 shadowing / JSX 境界の属性名ガード）は全部ピンした。

書いていて**自分のテストラッパのバグを1つ見つけた**ので記録しておく。子コンポーネントを被テストコンポーネントの**前**に宣言していたため、`analyzeComponent` がモジュール先頭の関数を解析する仕様で `Child` のほうを解析し、全ケースが診断ゼロを返していた。negative が全部「正しい理由でない理由で」通る状態で、これに気づかなければ何も証明しないスイートを commit するところだった。子を後ろに移して解決し、positive ブロックが同じラッパを共有することでこのハザードに対する生きたコントロールになるようにした。コメントにも残した。

## nitpick: binding / parameter の default — nitpick ではなかった (`45fad0e72`)

「very narrow, not blocking」とあったので、実際に何にコンパイルされるか測った:

```js
hydrate('DD', { init: initDD, template: (_p) => `<div ${((() => { const { x = count } = ({}); return String(x) })()) != null ? ... }></div>` })
```

この `template` は**モジュールスコープの thunk** で、`count` は `initDD` の内側で宣言された束縛。つまり CSR mount で `ReferenceError` になる。これは #2751 の機構そのもので、**この PR が閉じるために存在しているバグクラスの生きた実例が、閉じるはずの検査の内側に残っていた**。narrow ではあっても性質が違うので、後回しにせず直した。

parameter default は params を bind する**前**に訪問している — default は外側スコープで評価されるので `(count = count) => …` は外側の `count` を読む。先に bind すると自己 shadowing として誤って隠れる。

一般的な形（`{ size = 'md' }`）は default がリテラルで reactive 名ではないため無反応のまま。これもテストでピンした。

## 検証

| | |
|---|---|
| `bun test packages/jsx/src` | 3151 pass / 28 skip / 3 todo / **0 fail** (3182 across 246 files) |
| BF044 スイート単体 | 28 pass / **0 fail** |
| perturbation | 広げる前の門に対して positive 6件が赤 |

`pruneUnusedPropExtractions: generated code did not parse; skipping prune` が実行中に1度出るが、`prune-unused-prop-extractions.ts` の防御的な warn 経路で main にも存在する。この変更由来ではなく、落ちたテストも無い。

---

kfly8:
第2ラウンドの指摘2件に対応した (`835354179`)。追う過程でもう1件出たので合わせて3件。**3件とも false positive（正当なコードを新たに拒否する）方向**で、門を広げる変更としては一番まずい向き。

## 1. parameter default の訪問順 — 指摘どおり

「全 default を訪問してから全 param を bind」は JS のスコープ規則と食い違っていた。V8 で確認:

```
$ node -e 'console.log(((a, b = a) => b)(7))'
7
```

後続の default は**既に bind 済みの先行 param** を読む。つまり `(count, x = count) => …` で `count` が signal 名と衝突していると、`x` の default を裸の signal 読みとして誤検出していた。左から右への1パスに変更。

指摘には無かったが**パターン内部も同じ規則**なので (`{ a, b = a }`)、そちらも同時に直した。

## 2. docstring の TDZ 記述 — 指摘どおり、私の誤り

前コミットで `(count = count) => …` は「外側の `count` を読む」と書いた。誤り:

```
$ node -e 'try { (function(count = count){})() } catch(e) { console.log(e.message) }'
Cannot access 'count' before initialization
```

自分自身の binding が TDZ にあるので ReferenceError。**言語に無い挙動を説明するコメント**だったので、もっともらしい理由づけとして残さず書き直した。

## 3. 型注釈の中を歩いていた — 上の1を検証中に発見

pattern-element のテストが落ちたので調べたところ、原因はテストではなく walk のほうだった:

```
type annotation only        -> BF044 が1件出る（出てはいけない）
pattern elem, no type ann   -> 0（正しい）
pattern elem WITH type ann  -> 1（型注釈のせい）
```

`({} as { count?: unknown })` が rendered position にあると、**型リテラルのプロパティ名 `count`** を同名 signal への裸参照と読んで、正当なコンポーネントを拒否していた。型は値ではないので `ts.isTypeNode` で遮断。`as` / `satisfies` / 型アサーション / ジェネリック型引数を一括でカバーする。

## 検証

| | |
|---|---|
| `bun test packages/jsx/src` | 3157 pass / 28 skip / 3 todo / **0 fail** (3188 across 246 files) |
| BF044 スイート単体 | 31 pass / **0 fail** |
| false positive 3形 | いずれも 0 件に（修正前は 1 件ずつ出ていた） |
| true positive | ternary / destructure default / param default とも発火継続 |

3件とも回帰テストでピンした。特に「後続 default が先行 param を読む」形と「型注釈」は、外すと即座に赤くなる。

---

kfly8:
数値の訂正を1件。

会話とレビュー対応の中で quarantine 台帳のエントリ数を「111 → 70」と述べた箇所がありますが、**誤りです。正しくは 111 → 67**（本文の記載どおり）。

原因は私の計測で、`grep -c 'caseId:'` が実エントリ以外に3箇所を数えていました:

```
54:  caseId: string                                              ← interface のフィールド定義
62:function key(caseId: string, oracle: OracleKind): string {    ← 関数引数
204:export function pairwiseQuarantineEntry(caseId: string, ...  ← 関数引数
```

`^  \{ caseId:` で数え直した確定値:

| ref | エントリ数 |
|---|---|
| `main` | 111 |
| このPR の head (`0585546`) | **67** |

graduation 数 44 と、issue 別の内訳（#2755 22/22、#2751 8/8 が節ごと消滅、ほか部分削除）は厳密な方のパターンで数えていたので**変わりません**。差分ではなく端点だけの誤りでした。

同じ誤りが stack 上の #2762 側の会話にもあります（「70 → 52」と述べた箇所）。あちらの確定値は **67 → 49** で、削除18件という差分は変わりません。#2762 の本文には端点を書いていないので、そちらの本文の訂正は不要です。

---

kfly8:
**English restatement of my two earlier comments on this PR** (they were written in Japanese; GitHub has no comment-edit API available here, so this supersedes them rather than replacing them in place). No new information — everything below is already reflected in the PR body.

## Review round 1 — the missing tests

The finding was correct: 272 lines were added to the gate with no committed test touching it, and the "13/13" figure in the original body came from a dev-time probe that was never promoted. Ported into `signal-getter-not-called.test.ts` and **perturbation-verified**: against the pre-widening gate the six positive cases fail and the negatives still pass.

Writing those tests exposed a bug in my own test wrapper worth recording: the child component was declared *before* the component under test, and `analyzeComponent` analyzes the first function in the module — so it analyzed the child and reported zero diagnostics everywhere. Every negative case would have passed for the wrong reason. The child now sits after, and the positive block shares the same wrapper so it stays a live control on that hazard.

## Review round 2 — the "nitpick" was not a nitpick

Flagged as "very narrow, not blocking". Measuring what such code actually compiles to:

```js
hydrate('DD', { init: initDD, template: (_p) => `<div ${((() => { const { x = count } = ({}); return String(x) })()) != null ? ... }></div>` })
```

That `template` is a module-scope thunk referencing `count`, a binding declared inside `initDD` — `ReferenceError` on CSR mount. That is #2751's mechanism, i.e. **a live instance of the bug class this PR exists to close, sitting inside the check meant to close it.** Fixed rather than deferred.

Following it up surfaced two more, both also false positives: the parameter binding order (JS binds left to right — `((a, b = a) => b)(7)` returns `7`, verified against V8) and the walk descending into type positions (`({} as { count?: unknown })` read the type literal's property name as a bare signal reference). All three are in the direction that refuses working code.

## Correction to a number I stated

I wrote "111 → 70" for the quarantine ledger in an earlier comment. **That was wrong; the correct figure is 111 → 67**, as the PR body says. My `grep -c 'caseId:'` was counting three non-entries — the interface field declaration and two function parameters:

```
54:  caseId: string
62:function key(caseId: string, oracle: OracleKind): string {
204:export function pairwiseQuarantineEntry(caseId: string, ...
```

Counting with `^  \{ caseId:` gives 111 on `main` and 67 on this head. The graduation count (44) and the per-issue breakdown were computed with the stricter pattern and are unaffected — only the endpoints were misstated. The same error appeared in my commentary on #2762, where the correct figures are 67 → 49.

---

Issue #2482, opened 2026-08-02
Design review: loop lowering — is the recurring fragility architectural?


kfly8:
Upfront investigation of the maintainer hypothesis raised in the #2481 discussion: *"the recurring loop fragility may indicate a design problem in the loop lowering itself, not a series of independent bugs."* Rather than waiting for the pairwise sweep's density map, this review builds the map from evidence that already exists: the tracker history (30 loop-matching known-limitation issues analyzed; 42 match total, first page analyzed exhaustively) and a structural inventory of the loop machinery.

**Verdict up front: the hypothesis is supported, but with precision — the fragility localizes to two specific architectural absences plus one measured complexity burden, not to "loops" as a whole.** The discovery→fix loop itself works well (28/30 analyzed issues closed, `map-body-no-silent-divergence`'s known-hole set is empty); what the history shows is that the *cost recurs by construction*.

## Finding 1 — There is no scope model for loop-bound names (largest cluster: 9 issues, 0 open)

#2221, #2222, #2231, #2235, #2236, #2237, #2245, #2087, #1677. The same defect — a name lookup that ignores loop-callback scope — re-manifested independently in **at least six code paths**: per-adapter const inlining (#2221: *"a flat `this.localConstants.find(...)` … with **no notion of AST scope**"*, replicated *"nearly identically in every adapter built on the shared `ParsedExprEmitter` architecture"*), the Go bare-identifier fast path (#2236), record-literal lookup (#2237), CSR template substitution (#2222), a shared-IR attribute fold (#2235), and ERB filter-param resolution (#2245).

Every fix was a **local guard at the consuming site**; no shared scoped binding-resolution service was ever introduced (#2235's own text names the general fix: *"a scope-accurate guard, not the coarse whole-component exclusion the per-adapter fixes use"*). Structurally, this guarantees the next consumer of `localConstants`/prop names re-introduces the bug — the #2468 CSR-scope family this month was the same class in a new consumer.

## Finding 2 — Per-row construction is bolted onto per-component machinery (chain: #2445 → #2448 → #2447 → #2389)

The adapter data contract assumes values are built **once per component**; loops need **per-row instantiation**, and each retrofit has exposed the next layer:

- #2445: child props emitted as *"a single struct field on the parent, built once outside the loop"* → every row identical.
[...]

PR #2762, opened 2026-08-29
Collapse the row-key attribute into one IR-resolved decision (#2753)

kfly8:
Family 2. Stacked on #2761 (base is `claude/barefootjs-bf044-recursive`).

## Why

**16** places decided "does this row root carry a key attribute, what is it called, and what is its value" — 7 client (`map-array.ts` ×3, `map-array-lazy.ts` ×1, `component.ts` ×3) and 9 SSR adapters. The client stamped unconditionally; SSR emitted only for keyed loops, with a depth-suffixed name when nested. #2753's two measured shapes are both that disagreement, and fixing one leaves fifteen other sites to produce the next one.

This is a failure mode CLAUDE.md already records twice — six independent `BindingScope` implementations collapsed by #2482, and the placeholder convention that "distributes a substitution obligation across every emitter". Same shape, different subject.

## What

`IRElement.carriesDataKey?: boolean` becomes `IRElement.keyAttr?: { name: string; value?: string }`, resolved once in `jsx-to-ir.ts`. **Absent means the element carries no key attribute** — the correct final answer for an unkeyed loop's row, not a hint adapters must gate further.

Two disjoint sources:

- **`value` present** — the element is directly a `.map()` row root compiled inline here; `IRLoop.key` / `.depth` already resolved the expression and nesting depth.
- **`value` absent** — the element is one of this component's own render roots, relaying a key its caller supplies at runtime (the `data-key` prop / Rust `bf.data_key` field / client `createComponent`'s `key` argument).

`keyAttrName(depth)` moved to `@barefootjs/shared` — the one package every consumer already depends on, which is what makes "compute it once" structural rather than aspirational.

Deleted: Hono's `loopKeyStack`; every DSL adapter's `currentLoopKeyDepth` (`loopKeyDepthStack` in Go) and its `attr.name === 'key'` rewrite; the `rootScopeNodes` / `collectRootScopeNodes` copy in each `lib/ir-scope.ts` (go-template's file had nothing else left and is deleted outright). All 9 adapters now read `element.keyAttr` and emit.

## `needsHydration`

`map-array.ts` decided this by reading the DOM:

```ts
existingRanges.length > 0 && (!existingRanges[0]?.primaryEl.hasAttribute('data-key') || scopes.size === 0)
```

That depends on an attribute the compiler may legitimately not emit, so it could not stay. Investigating it showed the condition was **vacuous**: `scopes` is a closure-local Map, and this branch runs at most once (guarded by a `hydrated` flag set on the line above), at which point `scopes` has had no writes — its first write is inside the branch it guards. So the `||` reduced to the first probe, and that probe was itself unsound: it is name-specific, so it misdiagnosed both Shape A (an unkeyed row legitimately carries no key attribute) and Shape B (a nested row carries a depth-suffixed name). Simplifying to `existingRanges.length > 0` is behaviour-preserving, not a new rule. Recorded in a Why-not comment.

The write guards use a falsy-value `getAttribute` check rather than `hasAttribute`: a compiled template can bake the attribute present-but-empty for some shapes (measured on the csr-mount leg of a keyed top-level loop), and `hasAttribute` would treat that as "already correct" and skip the corrective backfill.

## Verification

All reproduced independently of the implementation report.

| | |
|---|---|
| `bun test packages/client/__tests__` | 664 pass / 0 fail |
| `bun test packages/jsx/src` | 0 fail |
| `bun test packages/adapter-tests/src` | 0 fail (4 skip) |
| `pairwise:generate` | 97 cases, ok=62 refused=35 broken=0 — matches the #2761 baseline |
| pairwise Playwright sweep | 187 passed / 0 failed — same |

`packages/client` was rebuilt before the sweep, since e2e serves `dist/`.

## Quarantine ledger: 67 → 49

18 of the 24 #2753-attributed rows deleted (genuinely fixed). The remaining 6 survive the sweep but **not for #2753** — each was re-measured by un-quarantining it and reading the raw diff, then re-attributed to what actually keeps it red. Two mechanisms, both of which `assertSnapshotsAgree`'s first-diff-wins compare could not reach while the `data-key` difference sorted first:

- **4 rows** (2 caseIds × snap/three-point) — the residue is #2714's ref-callback `data-mounted`, which SSR cannot produce. The unkeyed `fragment-row-loop` now emits no row-key attribute in either leg, which is the correct output for an unkeyed loop.
- **2 rows** (three-point only) — a genuinely masked second divergence: #2750 and #2756.

**#2753 now has zero entries.** Leaving rows filed under a fixed issue is the exact misdirection this ledger exists to prevent.

## Remaining seam (not addressed here)

The client-JS emitters still compute `keyAttrName(inner.keyDepth)` from their own depth counters rather than reading `element.keyAttr.name`. Both derive from the same shared function and the same "does a key expression exist" gate, so they cannot disagree today — but it is two derivations of one fact. Stating it rather than leaving the next sweep to find it. #2763 is the concrete case where that seam already shows: `html-template.ts` reads the raw `key` attribute independently of what `extractLoopKey` decided, so a fragment-bodied `.map()` gets a client-side `data-key` that no adapter emits.


kfly8:
レビューで、集約先の `applyLoopKeyAttr` が届いていない形を1つ見つけたので記録する。**この PR の回帰ではない**（前後でエミッタ出力はバイト一致を確認済み）ので、この PR では直さず #2763 に切り出した。

`.map()` の本体が fragment を返す形（`{items().map(item => <><li key={item.id}/><li/></>)}`）で、SSR は行キー属性を出さず client の行テンプレートだけが `data-key` を焼く。#2753 Shape A と同じ署名。

当初は `applyLoopKeyAttr` に fragment 分岐が無いだけだと思って patch を当てたが、実際にレンダリングしたら出力が変わらなかった。上流の `extractLoopKey` にも fragment 分岐が無く `key === null` になるので、`applyLoopKeyAttr` の呼び出し自体が飛んでいる。直すと **その `.map()` が位置ベースから keyed 照合に変わる**ので、一行埋める話ではなく設計判断になる。詳細と選択肢は #2763 に書いた。

この PR に関係する点として1つ: `html-template.ts` が生の `key` 属性を `extractLoopKey` の判断と独立に読んで `data-key` に書き換えている（`a.name === 'key' ? keyAttrName(loopDepth) : ...`）。行キーの決定を1箇所に集約したこの PR の趣旨からすると、client テンプレートも `IRElement.keyAttr` を読むべき箇所。本文の「残っている継ぎ目」に書いた client-JS エミッタの深さ再計算と同じ根で、#2763 の解決時にまとめて閉じるのが自然。

---

kfly8:
**English restatement of my earlier comment on this PR** (it was written in Japanese; no comment-edit API is available here, so this supersedes it rather than replacing it in place). No new information.

Review of this change turned up one shape the new consolidation point does not reach. **It is not a regression from this PR** — emitter output is byte-identical before and after — so it is filed as #2763 rather than fixed here.

A `.map()` whose body returns a fragment (`{items().map(item => <><li key={item.id}/><li/></>)}`) emits no row-key attribute from SSR while the client row template bakes `data-key`. Same signature as #2753 Shape A.

My first read was that `applyLoopKeyAttr` was simply missing a `fragment` branch, and I wrote that patch — but re-rendering showed the output unchanged. `extractLoopKey` has the same gap upstream, so `key` resolves to `null` and the call to `applyLoopKeyAttr` is skipped by its own guard. Making it fire turns that loop from positional to keyed reconciliation, which is a semantics decision rather than a fill-in. Reverted the patch and wrote up both defensible resolutions in #2763.

One point does belong to this PR's subject: `html-template.ts` reads the raw `key` attribute directly (`a.name === 'key' ? keyAttrName(loopDepth) : ...`), independently of what `extractLoopKey` decided. That is why the client template bakes a key no adapter emits. Same root as the seam already noted at the end of the PR body — client emitters deriving from their own inputs instead of reading `IRElement.keyAttr` — so it belongs in #2763's fix rather than bolted on here.

Also correcting a number from that comment: I stated the ledger went 70 → 52. **The correct figures are 67 → 49.** My `grep -c 'caseId:'` was counting the interface field declaration and two function parameters alongside the real entries. The delta (18 deleted) and the per-issue breakdown were computed with the stricter `^  \{ caseId:` pattern and are unaffected.

---

Issue #2722, opened 2026-08-26
Fragment-wrapped root: createComponent's CSR mount loses or mis-stamps the root `bf-s` scope id


kfly8:
## Problem

For a fragment-rooted component, `createComponent()`'s pure-CSR mount path diverges from the SSR/hydration path on scope-id stamping:

- **Missing `bf-s` entirely** on the mounted root (`button`, `conditional-return-button`, `link`, `kbd`, `tooltip` mutants — 9 triples).
- **Wrong scope NAME**: after one interaction, `select`'s mount carries `Select_*` where the hydrated leg carries `SelectBasicDemo_*` — the child component's identity instead of the mounting component's (10 more triples surface the same class only after interaction replay: `radio-group`, `data-table`, `pagination`, `tag-cloud`, `todo-app`, `props-reactivity-comparison`, `textarea-native-shared`, `branch-root-prop-attr`, `reactive-props`).

Found by the #2481 mutation sweep (`fragment-wrap` mutation) three-point/idempotence oracles. Related machinery: the `comment: true` fragment-scope path (`commentScopeRegistry`) vs `materializeComponent`'s bf-s stamping — the #2649/#2708 work covered the component-root disjunct; this is the fragment disjunct's CSR-mount half.

## Repro / pins

Quarantined in `packages/adapter-tests/e2e/mutation-quarantine.ts` (rot-checked; deterministic regeneration via `test:mutation`).

## Graduation

Make CSR mount stamp the same scope identity the SSR/hydration path produces for fragment roots; confirm the 19 triples pass, delete their ledger entries.

Issue #1320, opened 2026-05-16
CSR: `children={<jsx/>}` attribute form emits duplicate `bf-s` and drops the inner scope marker


kfly8:
Surfaced limitation #9 from PR #1306's catalog. Parent: #1244.

## Problem

Passing JSX through the explicit `children` prop (`<Box children={<span>x</span>} />`) is functionally identical to nesting it (`<Box><span>x</span></Box>`), but the CSR path handles the two parse shapes differently. The attribute form generates a client template with **two `bf-s` attributes on the same element** (e.g. `bf-s=\"test_s0\" bf-s=\"test\"`) and **drops the inner scope marker** on the hoisted child. SSR renders the correct shape.

Same class of bug as the existing `record-index-lookup-via-child-prop` skip already on the CSR conformance test — CSR's child-component render emits the parent's scope marker alongside the child's.

## Current shape

- Fixture: `packages/adapter-tests/fixtures/children-jsx-expression.ts` — pins `<div bf-s=\"test_s0\"><span bf-s=\"test\">x</span></div>` (SSR-correct).
- CSR skip declaration: `packages/adapter-tests/src/__tests__/csr-conformance.test.ts:70-75` — `children-jsx-expression` listed in `skipFixtures` with the duplicate-`bf-s` symptom documented in-line.
- Existing precedent: `record-index-lookup-via-child-prop` (`packages/adapter-tests/src/__tests__/csr-conformance.test.ts:61`) carries the same duplicate-scope-marker symptom — fixing one likely fixes the other.
- Likely emit site: CSR's `IRComponent` handler in `packages/jsx/src/ir-to-client-js/html-template.ts` — the `children` attribute path probably re-attaches the parent's scope to the hoisted child rather than letting the child carry its own.

## Repro

```tsx
function Box({ children }: { children: any }) { return <div>{children}</div> }
export function ChildrenJsxExpression() {
  return <Box children={<span>x</span>} />
}
```

Compile via Hono / CSR and inspect the generated client JS template:

- Expected: `<div bf-s=\"test_s0\"><span bf-s=\"test\">x</span></div>`
- Actual: `<div bf-s=\"test_s0\" bf-s=\"test\"><span>x</span></div>` — duplicate `bf-s` on the wrapper, inner scope marker dropped.

To re-enable the CSR conformance test, drop `'children-jsx-expression'` from the `skipFixtures` Set.

## Intended behavior

[...]

PR #2768, opened 2026-08-29
Make the ref mirror and the top-level wrapper scope read the answers that already exist (#2749 / #2757)

kfly8:
Second half of family 3, completing the family #2764 started. Both fixed **without adding anything to `IRAttribute`** — in each case an authority already existed and one path ignored it.

## #2749 — `ref` on a child-component call site

`packages/shared/src/dom-prop.ts`'s `classifyDOMProp` describes itself as *"the single source of truth for 'how should this prop reach the DOM?' Both runtime helpers (applyRestAttrs, spreadAttrs) and compile-time emitters (emitAttrUpdate, html-template) consume this classification."*

Measured: **no compile-time emitter consumed it.** Only the two runtime helpers did — the one reference in `ir-to-client-js/utils.ts` was a comment about re-exporting a sibling function. `collectReactiveChildProps` (`collect-elements.ts`) hand-rolled

```ts
prop.name.startsWith('on') && prop.name[2] === prop.name[2].toUpperCase()
```

which covers events but has no `ref` case, so a `ref` fell through to the generic dynamic-prop mirror and was emitted as `setAttribute('ref', String())`. SSR never emits that attribute, so the hydrate leg alone added it and failed the SSR-vs-hydrated snapshot.

The mirror now takes the same three exclusions (`ref` / `event` / `skip`) from the same classifier that `applyRestAttrs` reads.

**The issue body was wrong on one point, and it matters for triage.** It claimed the callback never runs. It does: `initChild` → `applyRestAttrs` → `classifyDOMProp` → `kind: 'ref'` → the callback fires. Verified on the unmodified runtime. The mirror ran *in addition*, so the only real symptom was the leaked attribute — and that mistaken premise is why #2749 was filed as distinct from #2714 (see the ledger section).

Two small widenings come with reading the shared classifier, both intended and documented in the code: `skip` now also covers a `children` prop arriving as an expression rather than the `jsx-children` variant, and the classifier's event test requires a real A-Z at index 2 where the old check also matched `on-foo` / `on1x`.

## #2757 — top-level root-is-a-child-call scope

The framing this PR was briefed with was **wrong**, and the correction is the useful part.

The brief said SSR "already has the answer at compile time" because `renderChild(..., 's1')` passes a slot id, so only CSR re-derives at runtime. But the slot id was never the missing piece — the *client* template passes it too; it is the same compiled `template:` lambda. What is ambient is the **parent scope prefix**, which is genuinely per-instance (each mount gets a fresh id) and therefore cannot be a compile-time constant. So this is not "one side re-derives a compile-time constant at runtime"; it is **a value that was never computed at all** for one shape.

`materializeComponent` set `_parentScopeId` from its own `scopeId` or from `slot.parent`. A `comment: true` / `fragmentRoot: false` wrapper at a top-level mount has neither — `scopeId` is null by design (the parsed `firstChild` is the child's own already-scoped element) and there is no `slot`. So `renderChild` took its "no parent known" fallback and named the child after itself.

The fix derives a scope id **for threading only** — the same split #2722 already made for a genuine fragment root: non-null `scopeId`, attribute write skipped. Guarded on `!_parentScopeId` so a wrapper materialized inside an outer template eval still inherits the caller's ambient.

### On removing `_parentScopeId`

Considered and **rejected**. The ambient is not an accident: a per-instance parent prefix has to reach `renderChild` calls buried inside a template string literal, and the alternative is threading a context argument through every generated template lambda and every adapter's renderer. Its 22 references encode four interacting fixes (#1320 hoisted-children placeholder, #2444 grandchild composition, #2649 third-level collapse, #2722 fragment roots), each with its own save/restore discipline. The measured defect was a missing computation, not the mechanism.

## An existing test's expectation changed — why that is sound

`create-component-parent-scope.test.ts` › *"strips the placeholder when no slot.parent is provided"* went red. It was **not** assumed stale. The Hono reference for the equivalent source stamps the hoisted element with the **outer** component's scope:

```tsx
function Box({ header, children }) { return <div>{header}{children}</div> }
export function Outer() {
  const [n, setN] = createSignal(1)
  return <Box header={<h1>{n()}</h1>} />
}
```
```html
…
```

So the pre-fix CSR behaviour (no `bf-s` at all) was itself the divergence, and the post-fix behaviour matches SSR. The test now asserts the new value, and the evidence — source, the emitted `hydrate` line showing the `__BF_PARENT_SCOPE__` placeholder, the full Hono render — is committed into the test's comment rather than left in a review thread.

One precondition is easy to miss and is recorded there too: the placeholder is emitted **only for a JSX element passed as a prop** (`children={<X/>}` / `header={<X/>}`). Ordinary nested children (`<Box><X/></Box>`) compile down a different path with no placeholder at all, produce `bf="^s1"` and no `bf-s`, and are **not** a counter-example.

This is also a second consequence of the same one-line gate: it fixed a CSR/SSR divergence nobody had filed.

## Verification

Every number reproduced independently of the implementation report.

| | baseline (#2764) | this branch |
|---|---|---|
| `bun test packages/jsx/src` | 3197 run / 0 fail | **3201 run / 0 fail** (+4 new) |
| `bun test packages/client/__tests__` | 664 pass / 0 fail | **668 pass / 0 fail** (+4 new) |
| `bun test packages/adapter-tests/src` | 2208 run / 0 fail | **2220 run / 0 fail** (4 skip) |
| `bun test packages/compat` | 508 pass / 0 fail | **508 pass / 0 fail** |
| `pairwise:generate` | 97, ok=62 refused=35 broken=0 | **identical** |
| pairwise Playwright sweep | 187 passed / 0 failed | **187 passed / 0 failed** |

### Perturbation

| reverted | red | still green (reverse-direction guards) |
|---|---|---|
| the `classifyDOMProp` read in `collect-elements.ts` | `child-component-ref-not-mirrored` 1/4 | `get ref()` still reaches `initChild`; CSR template still has no `ref=`; an ordinary reactive `title` prop **still mirrors** |
| the `!_parentScopeId` branch in `component.ts` | `issue-2757-…` 1 **and the rewritten `create-component-parent-scope` test** | a slotted wrapper still inherits `slot.parent` (#1320); a non-comment top-level component untouched |

The rewritten test failing under perturbation is what shows it pins behaviour rather than having been bent to fit the implementation.

## Quarantine ledger: 34 → 25

- **#2757 — 9 rows deleted**, all stale.
- **#2749 — 6 rows re-filed to #2714**, not deleted. After the fix they still fail, so they were un-quarantined and the raw diff read: no `ref=` on either leg (#2749 is fixed), but the hydrated leg carries `data-mounted="0"` that SSR lacks — because a `ref` callback **cannot run during SSR**. That is exactly #2714, which #2749's body had explicitly distinguished itself from on the incorrect premise that the callback never ran. #2749's `setAttribute('ref')` check fired first and masked it.

A correction is recorded in the file docstring: the six #2757 `idempotence` reasons stated the divergence **backwards** (naming the csr-mount and SSR sides the wrong way round). The `three-point` reasons and the issue body had it right. Same Expected/Received swap #2714's own correction records — the rows are gone, but the pattern is noted.

## Fixtures

Two added, both registered, **both honest in their docstrings that they do not redden pre-fix**, with the executable pin named:

- `child-component-ref-prop` — the defect lived only in the `init` builder, and no conformance layer here evaluates `init` against a DOM. Pin: the jsx unit test.
- `child-component-root-scope-prefix` — CSR-mount-only, and `csr-render.ts` evaluates the compiled `template:` lambda against a mock runtime, never the real `createComponent`. Pin: the client runtime test.

Both `expectedHtml` blocks generated from the Hono reference; both pass the CSR conformance leg.

Generated artifacts regenerated in dependency order — `coverage-map` → `compat:lock` → `support-matrix:lock`. `coverage-map.json` is a third artifact the #2764 round did not need and its freshness test caught it. Diffs are exactly the two fixtures: `coverage-map.json` +2 entries; `ui/compat.lock.json` `totalFixtures` 347 → 349; `ui/support-matrix.lock.json` 96 lines, all `total`/`pass` counters +2 across the 9 adapters.

## Defects found, not fixed

1. **`classifyDOMProp`'s docstring still overclaims.** After this change exactly one compile-time emitter consumes it; `emitAttrUpdate` and `html-template` still re-derive attribute-vs-property by hand. That is the remaining surface of this family.
2. **`dangerouslySetInnerHTML` as a reactive child-component prop** would still mirror as `setAttribute('dangerouslySetInnerHTML', String({__html:…}))` → `[object Object]`. The `dangerous-inner-html*` fixtures only cover the element-level path. Same bug class as #2749, unmeasured, left out deliberately and documented as a why-not in the code.
3. **The `bf-r` marker is CSR-invisible.** Hono emits `bf-r=""` on client-component roots; `renderChild` never does. `normalizeHTML` strips it, so no conformance layer compares it. Not investigated further.

---


kfly8:
## `update-expected-html` is red, and it is not this PR's

Diagnosed and bisected. **This failure is inherited from `main`** — it reproduces on `main` with this PR's changes absent.

Ran `bun run packages/adapter-tests/scripts/generate-expected-html.ts` at each commit and counted drifting files:

| commit | drift |
|---|---|
| `724411118` — `main` before this session's three merges | **0** |
| `2ced28129` — #2761 merged | **0** |
| `935d6eadb` — #2762 branch head, before merge | **8** |
| `4ad1c0bbb` — #2762 merged | 8 |
| `50b2fa90a` — current `main` | 9 |
| this PR's head | 10 |

So 9 of the 10 predate this PR. #2762 was already drifting on its own branch head before it was merged; its `update-expected-html` was not verified by name before merge — my error.

### What the drift is

Eight production components whose root is a Context Provider (select, popover, accordion, carousel, combobox, command, dropdown-menu, radio-group) lost the caller-key relay from their scope-bearing element:

```diff
-  const { …, "data-key": __dataKey, …props } = __allProps
+  const { …, "data-key": _dataKey, …props } = __allProps
-    <div data-slot="select" … bf-s={__scopeId} … {...(__dataKey !== undefined ? { "data-key": __dataKey } : {})}>
+    <div data-slot="select" … bf-s={__scopeId} …>
```

`resolveRootKeyAttr` (`jsx-to-ir.ts`) descends `element` / `if-statement` / `fragment` but **not `component`**, so the walk stops at the Provider and never reaches the `<div>` that carries `bf-s`.

The deeper cause is that the two adapter families disagreed **before** #2762, and #2762 unified them onto the wrong one:

| | pre-#2762 gate |
|---|---|
| Hono (reference) | `if (element.needsScope)` — relay emitted unconditionally |
| the 9 DSL adapters | `(rootScopeNodes.has(element) && element.needsScope) \|\| carriesDataKey` |

`collectRootScopeNodes` does not descend into a `component` either, so on a Provider-rooted component Hono emitted the relay and the DSL adapters did not. #2762 collapsed all sixteen sites onto one IR field — correct in intent — but took the answer from the DSL side, changing the reference adapter's output. That is #2753's own shape (a keyed loop row whose root carries no key) reintroduced in a different guise.

### Status

A fix is in progress on `claude/barefootjs-fix-root-key-relay`, branched from `main`. The success criterion is that the eight snapshots return to their committed content **by the fix alone** — regenerating them would pin the regression as correct, so it is explicitly not being done.

Once that lands on `main`, `main` will be merged into this PR and the check re-runs. Nothing in this PR needs to change for it.

### The tenth file

`todo-app.client.js` drifts only at `50b2fa90a`, i.e. it appeared with #2764's merge, not #2762's. A reactive `checked` disappears from a conditional-branch template:

```diff
- <input id="toggle-all" … type="checkbox" ${todos().every(t => t.done) ? 'checked' : ''} bf="s2" />
+ <input id="toggle-all" … type="checkbox" bf="s2" />
```

Being triaged separately — it may share the root cause above or be independent. It is also not this PR's: this PR's head adds one further drifting file (`command.client.js`) beyond `main`'s nine, and that one is being checked as part of the same investigation.

---

kfly8:
## `command.client.js` — resolved, and it is not a defect

I said earlier this file was "being checked as part of the same investigation" and deferred it. Here is the answer, read off this PR's own `update-expected-html` job log (run `33254187343`), which is the authoritative record.

The job reports **10 drifted files**, in three groups:

| group | files | cause |
|---|---|---|
| `data-key` relay lost from Hono SSR | `accordion`, `carousel`, `combobox`, `command`, `dropdown-menu`, `popover`, `radio-group`, `select` — 8 `*.ssr.tsx` | inherited from `main` (#2762 regression) |
| stale `todo-app` snapshot | `todo-app.client.js` | inherited from `main` (#2764 left it stale) |
| **`command.client.js`** | 1 | **this PR — and it is the #2749 fix landing** |

The `command.client.js` hunk is exactly the mirror this PR deletes:

```diff
-    const [__Button_s1El] = $c(__scope, 's1')
-    if (__Button_s1El) {
-      { const __v = (el) => {
-          el.addEventListener('click', () => setOpen(true))
-        }; if (__v != null) __Button_s1El.setAttribute('ref', String(__v)); else __Button_s1El.removeAttribute('ref') }
-    }
```

That is `setAttribute('ref', String(<callback>))` on a child-component call site — the precise emission #2749 is about, and the only place in the shared-component snapshots where a `ref` reaches a child component. `classifyDOMProp` now classifies it as `kind: 'ref'` and the mirror skips it, so the effect body loses those six lines. The `ref` callback still runs, via `initChild` → `applyRestAttrs`, as the PR body describes.

So the snapshot is *correct output that has not been committed yet*, not a divergence. Nothing to fix in the change; the regenerated file just needs to land here.

### Correction on method

My earlier reproduction of this drift was invalid and I am recording why, because the trap is reusable. I ran `generate-expected-html.ts` in a git worktree whose `node_modules` were symlinked to the primary clone. Workspace packages resolve through those symlinks, so `@barefootjs/jsx` resolved to the **primary clone's** compiler, not the worktree's checkout — both the worktree at this PR's head and one at `main` produced identical output, and neither measured what it claimed to. This is the same symlinked-worktree hazard already known to break TypeScript type resolution; it also silently substitutes the compiler under `generate-expected-html`. A drift measurement has to run in a real, independently installed checkout — or be read from CI, as it was here.

### Order of operations

1. #2772 lands on `main` (restores the `data-key` relay, regenerates `todo-app.client.js`).
2. Merge `main` into this branch — 9 of the 10 files clear with no edit here.
3. Regenerate and commit `command.client.js`, which is this PR's own and intended.

---

Issue #2732, opened 2026-08-27
SSR omits data-key for a fragment-wrapped loop item's root, breaking hydration's snap invariant


kfly8:
Found while fixing #2721 (fragment-root hydration prop loss) via the mutation sweep's `fragment-wrap` mutant on `toggle-shared`, after #2721's actual fix (the `hydrateCommentScope` props bug) landed and unmasked this narrower, distinct divergence.

## Mechanism

For a normal (element-rooted) component used as a keyed `.map()` row (e.g. `<ToggleItem key={item.label} .../>`), `renderElement` (hono-adapter.ts) stamps `data-key` on the scope root inside its `element.needsScope` block:

```ts
if (element.needsScope) {
  hydrationAttrs += ` ${BF_SCOPE}={__scopeId}`
  // ...bf-h / bf-m / bf-r / bf-p...
  hydrationAttrs += ' {...(__dataKey !== undefined ? { "data-key": __dataKey } : {})}'
}
```

When that same component's root is a JSX Fragment (`return <>...</>`), `jsx-to-ir.ts`'s `transformFragment` deliberately clears `needsScope` on the wrapped element (`ctx.isRoot = false` before transforming children) — the fragment itself owns the scope via a wrapping `<!--bf-scope:-->` comment (`wrapWithScopeComment`) instead of `bf-s`/`bf-h`/`bf-m`/`bf-r`/`bf-p` attributes. That's correct for those five properties, but `data-key` gets dropped in the same stroke — `wrapWithScopeComment` never re-adds it anywhere, so a fragment-rooted loop row's SSR markup carries no `data-key` at all.

At hydration, `mapArray`'s first-run adopt loop (`map-array.ts`) always stamps the key onto each adopted row afterward (`range.primaryEl.setAttribute(BF_KEY, key)`) — for the element-rooted case this is a no-op (SSR already had the same value), but for the fragment-rooted case it's a genuine, observable DOM mutation the raw SSR bytes never had. That breaks the `snap` oracle's invariant (hydration must be a no-op on visible SSR state): pre-hydration capture lacks `data-key`, post-hydration capture has it.

## Reproduction

[...]

PR #2772, opened 2026-08-29
Relay the caller's row key on provider-rooted components (#2753 regression)

kfly8:
**Fixes a regression on `main`.** `update-expected-html` is currently red there; this restores it.

## What broke

Eight production components whose root is a Context Provider — select, popover, accordion, carousel, combobox, command, dropdown-menu, radio-group — stopped relaying the caller-supplied row key from the element that carries their scope:

```diff
-  const { …, "data-key": __dataKey, …props } = __allProps
+  const { …, "data-key": _dataKey, …props } = __allProps
-    <div data-slot="select" … bf-s={__scopeId} … {...(__dataKey !== undefined ? { "data-key": __dataKey } : {})}>
+    <div data-slot="select" … bf-s={__scopeId} …>
```

That is #2753's own shape — a keyed loop row whose root carries no key — reintroduced by the PR that set out to close it.

### Bisect

`generate-expected-html.ts` run at each commit, counting drifting files:

| commit | drift |
|---|---|
| `724411118` — before the three merges | 0 |
| `2ced28129` — #2761 merged | 0 |
| `935d6eadb` — **#2762 branch head, pre-merge** | **8** |
| `4ad1c0bbb` — #2762 merged | 8 |
| `50b2fa90a` — current `main` | 9 |

#2762 was already drifting on its own branch head. Its `update-expected-html` was not checked by name before merge.

## Root cause

#2762 collapsed sixteen duplicated row-key decisions onto one IR field, `IRElement.keyAttr` — correct in intent. But the two adapter families **already disagreed** on this shape, and the collapse took the answer from the wrong one:

| | pre-#2762 gate |
|---|---|
| Hono (reference) | `if (element.needsScope)` — relay emitted unconditionally |
| the 9 DSL adapters | `(rootScopeNodes.has(element) && element.needsScope) \|\| carriesDataKey` |

`collectRootScopeNodes` never descended into a wrapper node, so on a provider-rooted component Hono emitted the relay and the DSL adapters did not. `resolveRootKeyAttr` inherited that walk and therefore the DSL answer, changing the **reference** adapter's output.

The specific node is `provider`, not `component`: `transformProviderElement` is the only root transform that does not clear `ctx.isRoot`, so the element under a provider is a genuine render root with `needsScope: true` — and the walk stopped one node above it.

## Which family is correct — Hono, on three independent grounds

1. Hono is the reference adapter; every fixture's `expectedHtml` is generated from it.
2. Semantically the element carrying `bf-s` **is** the component's rendered root. If a caller uses the component as a keyed loop row, that element must carry the key or `mapArray` cannot reconcile it.
3. **The client runtime already behaves this way.** `renderChild` / `materializeComponent` splice `data-key` onto the rendered markup's *first element* via `spliceAttrsAfterFirstTag`, whatever wrapper nodes stand above it. So CSR always stamps the inner element; SSR omitting it there is a hydration divergence, not a stylistic difference.

The third is the strongest: it is not a judgement about which adapter to prefer, it is an existing half of the contract that already picked a side.

## The fix

`resolveRootKeyAttr` no longer enumerates the constructs `isRoot` passes through. It tests `needsScope` over the whole tree — byte-for-byte the predicate the reference adapter evaluated at every `renderElement` before #2762, so equivalence is provable rather than argued. `needsScope` cannot be true inside a loop body or a child's slot children (the first element / scope-comment fragment consumes `isRoot`), so the wider walk costs no reach.

`markDataKeyCarrier` (#2732) is unaffected — a `needsScopeComment` fragment's children are `needsScope: false` by construction, so the guard skips them with no double-marking. `applyLoopKeyAttr` (mechanism 1) is protected by the pre-existing `!keyAttr` guard and runs during the transform, before this post-pass. Both are pinned.

**The eight snapshots are restored by the fix alone** — no regeneration. `generate-expected-html.ts` leaves `git diff` empty.

## Tests

Two new files, because the defect is invisible to the fixture layer: a standalone render passes no key, so the relay never appears in `expectedHtml`. That is exactly why only `update-expected-html` caught it. Both assert on the emitted *template*, checking the relay lands inside the same opening tag as `bf-s`.

- `packages/jsx/src/__tests__/root-key-relay.test.ts` — compiler layer.
- `packages/adapter-tests/src/__tests__/root-key-relay-cross-adapter.test.ts` — **pins Hono and three DSL adapters to the same answer for the same shape**, which is the class of drift that made this invisible for as long as both families have existed.

### Perturbation

Reverting only `resolveRootKeyAttr` to the pre-fix walk, tests untouched:

| file | result |
|---|---|
| `root-key-relay.test.ts` | 4 pass / **4 fail** — the provider-root rows; non-provider rows stay green |
| `root-key-relay-cross-adapter.test.ts` | 4 pass / **4 fail** — all four provider-root rows across hono / go-template / mojolicious / xslate; plain-element-root rows stay green |

## Second commit: the `todo-app` snapshot #2764 left stale

Separate cause, same job caught it. #2764 made the conditional-branch builder honour `clientOnly`, which correctly stopped it baking `checked` onto `#toggle-all` — but the snapshot was not regenerated.

Regenerating is right rather than reverting the emission, on measurement:

- SSR has never emitted it. `todo-app.html` has zero matches for `toggle-all…checked` at `724411118`, `4ad1c0bbb` and `50b2fa90a` alike.
- The branch registers an unconditional effect that sets it on both the hydrate and rebuild paths: `__ra_s2.checked = !!(todos().every(t => t.done))`.

So the baked attribute was surplus that only a reconciler-rebuilt branch carried while a hydration-reused one did not — the divergence #2764 set out to close. Dropping it makes the three legs agree.

## Verification

| suite | result |
|---|---|
| `bun test packages/jsx/src` | 3205 ran / **0 fail** (+8 new) |
| `bun test packages/client/__tests__` | 664 pass / **0 fail** |
| `bun test packages/adapter-tests/src` | 2216 ran / **0 fail** (4 skip, +8 new) |
| `bun test packages/compat` | 508 pass / **0 fail** |
| `pairwise:generate` | 97 cases, ok=62 refused=35 broken=0 |
| pairwise Playwright sweep | 187 passed / 0 failed |
| `generate-expected-html` | **`git diff` empty** |

All nine adapter conformance suites pass, run sequentially: blade 1533 / erb 1504 / go-template 1826 / hono 1606 / jinja 1533 / mojolicious 1702 / rust 1531 / twig 1541 / xslate 1533, zero failures. No conformance test disagreed with the new output, so there was no "DSL `expectedHtml` diverged from Hono" case to adjudicate.

## What this says about the guard

`keyAttr` had **no compiler-layer test coverage**. #2762 introduced the field, deleted ten adapters' worth of emit-time logic onto it, and shipped with nothing asserting what it resolves to — which is why every unit suite was green on a broken branch and only a snapshot job noticed.

Worth a follow-up sweep: other places where a DSL `lib/ir-scope.ts` copy diverged from Hono's inline equivalent before #2762 deleted them. That class of drift had no test that could see it either, and the cross-adapter test added here covers only this one shape.

---

### DIFF
--- a/.changeset/relay-row-key-on-provider-roots.md
+++ b/.changeset/relay-row-key-on-provider-roots.md
@@ -1,0 +1,20 @@
+---
+"@barefootjs/jsx": patch
+"@barefootjs/hono": patch
+"@barefootjs/rust": patch
+"@barefootjs/mojolicious": patch
+"@barefootjs/xslate": patch
+"@barefootjs/twig": patch
+"@barefootjs/erb": patch
+"@barefootjs/blade": patch
+"@barefootjs/go-template": patch
+"@barefootjs/jinja": patch
+---
+
+Restore the caller-key relay (`IRElement.keyAttr`, #2753's "mechanism 2") on components whose root is a `<Ctx.Provider>`.
+
+`resolveRootKeyAttr` looked for the component's render root by walking down from the IR root through `element` / `if-statement` / `fragment` and stopping at anything else. A provider is none of those, but `transformProviderElement` passes `ctx.isRoot` through to its children — so the element under the provider IS a render root, carries `needsScope`/`bf-s`, and the walk never reached it. Every adapter then emitted that root without the relay, and a caller rendering such a component as a keyed loop row got a row with no key attribute for `mapArray` to reconcile against.
+
+The relay is now resolved by testing `needsScope` throughout the tree rather than by an enumeration of the constructs `ctx.isRoot` passes through, which is the same predicate the reference adapter applied at emit time before the decision moved into the IR, and matches the client runtime's own CSR half (`renderChild` / `materializeComponent` splice `data-key` onto the rendered markup's first element whatever wrapper nodes sit above it).
+
+Affects every provider-rooted component, including `select`, `popover`, `accordion`, `carousel`, `combobox`, `command`, `dropdown-menu` and `radio-group`. The DSL adapters had never emitted the relay for this shape; Hono had, and regressed when the decision was centralized.
--- a/packages/adapter-tests/fixtures/__snapshots__/todo-app.client.js
+++ b/packages/adapter-tests/fixtures/__snapshots__/todo-app.client.js
@@ -191,7 +191,7 @@ export function initTodoApp(__scope, _p = {}) {
   })
 
   insert(__scope, 's1', () => todos().length > 0, {
-    template: () => { const __slots = []; return { html: `<!--bf-cond-start:s1--><input id="toggle-all" class="toggle-all" type="checkbox" ${todos().every(t => t.done) ? 'checked' : ''} bf="s2" /><label for="toggle-all">Mark all as complete</label><!--bf-cond-end:s1-->`, slots: __slots } },
+    template: () => { const __slots = []; return { html: `<!--bf-cond-start:s1--><input id="toggle-all" class="toggle-all" type="checkbox" bf="s2" /><label for="toggle-all">Mark all as complete</label><!--bf-cond-end:s1-->`, slots: __slots } },
     bindEvents: (__branchScope, { isFirstRun: __bfFirstRun = false } = {}) => {
       const [_s2] = $(__branchScope, 's2')
       if (_s2) _s2.addEventListener('change', handleToggleAll)
--- a/packages/adapter-tests/src/__tests__/root-key-relay-cross-adapter.test.ts
+++ b/packages/adapter-tests/src/__tests__/root-key-relay-cross-adapter.test.ts
@@ -1,0 +1,129 @@
+/**
+ * Render-root row-key relay, across both adapter families (#2753).
+ *
+ * `IRElement.keyAttr` with no `value` means "this element is one of THIS
+ * component's rendered roots; relay whatever key the caller supplies at
+ * runtime". Every SSR adapter must emit that relay on the SAME element it
+ * puts `bf-s` on — the client runtime reconciles keyed rows by reading the
+ * key attribute off the row's primary element, and it splices `data-key`
+ * onto the rendered markup's first element on the CSR side.
+ *
+ * The shape this guards is a component whose root is a `<Ctx.Provider>`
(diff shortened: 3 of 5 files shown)


kfly8:
## Correction: this PR's body is wrong about how the regression got through

The body says of #2762:

> #2762 was already drifting on its own branch head. Its `update-expected-html` was not checked by name before merge.

The first sentence is right. **The second is wrong, and it misdescribes the mechanism in a way that matters.** There was no job to check.

`.github/workflows/update-fixtures.yml` is gated:

```yaml
on:
  pull_request:
    branches: [main]
```

#2762 was stacked — its base was `claude/barefootjs-bf044-recursive`, not `main` — so the workflow never triggered. Measured:

- `list_workflow_runs` for `update-fixtures.yml` on branch `claude/barefootjs-datakey-ir` returns `total_count: 0`.
- `git log -S 'branches: [main]' -- .github/workflows/update-fixtures.yml` dates that gate to `1c169bb41`, 2026-02-26. It did not change during this window, so this is structural, not a config edit mid-flight.
- The bottom-of-stack PRs, whose base *was* `main` — #2752, #2761, #2772 — all ran the job and were green. That contrast is why the gap stayed invisible.

So the failure was not inattention to a red signal. **The drift gate structurally cannot see a stacked PR**, and #2762 was merged to `main` without anyone running the generator by hand — which for a stacked PR is the only way to get the answer.

This is the same gap I reported on #2774 an hour ago as an observation about that PR. I did not connect it back to #2762 at the time, and instead repeated an account of the regression that put the fault in the wrong place — as attention rather than as a missing gate. The wrong version is more comfortable and less useful: "check more carefully" fixes nothing here, because there is nothing to check.

The remedy is in #2776, which now splits the rule by base: check the job by name when the base is `main`; run `generate-expected-html.ts` by hand in a real checkout when it is not. It also repeats the trap that a worktree with symlinked `node_modules` resolves `@barefootjs/*` to the primary clone's compiler and therefore measures the wrong tree.

Leaving the body as-is with this correction below it, rather than editing it, so the original claim and its correction both stay visible.

---

PR #2774, opened 2026-08-29
Unify the children-slot predicate across all eight SSR adapters

### LINKED ISSUE #2773: go-template fills the reserved children slot from the first JSX-valued prop of ANY name — a named prop is duplicated into `children`, and a real `children` is dropped
## The divergence

Two implementations answer "which prop supplies the reserved children slot?", and they answer **different questions**.

Seven DSL adapters (blade, erb, jinja, mojolicious, rust, twig, xslate) share a byte-identical `resolveJsxChildrenProp` in their `lib/ir-scope.ts` — *"the prop **named** `children`, if it carries a `jsx-children` payload"*:

```ts
const prop = props.find(p => p.name === 'children')
if (!prop) return []
if (prop.value.kind !== 'jsx-children') return []
return prop.value.children
```

go-template's `jsxChildrenPropNodes` (`go-template-adapter.ts:3169`) is *"the **first** prop of **any** name carrying a `jsx-children` payload"*:

```ts
for (const p of props) {
  if (p.value.kind === 'jsx-children') return p.value.children
}
return []
```

Both are reached the same way — `comp.children.length > 0 ? comp.children : <helper>(comp.props)` — at one call site per DSL adapter (`blade-adapter.ts:1397`, `erb-adapter.ts:1302`, `jinja-adapter.ts:1232`, `mojo-adapter.ts:1239`, `minijinja-adapter.ts:1284`, `twig-adapter.ts:1282`, `xslate-adapter.ts:1208`) and at three in go-template (`3267`, `7341`, `7368`). So it is the same decision, not a coincidence of naming.

They diverge whenever a component call site passes a **named** JSX-element prop and the reserved children slot has no content of its own.

## Measured

Source:

```tsx
export function Card(props: { header?: any; children?: any }) {
  return (
    <section>
      <header>{props.header}</header>
      <div class="body">{props.children}</div>
    </section>
  )
}

export function Parent() {
  return <Card header={<span>HEADER</span>} children={<b>REAL-CHILDREN</b>} />
}
```

**Hono (reference)** — two distinct payloads, correct:

```jsx
<Card header={<><span …>HEADER</span></>} children={<><b>REAL-CHILDREN</b></>} … />
```

**Twig** (representative of all seven DSL adapters) — two distinct captures, correct:

```twig
{% set bf_prop_0 %}<span …>HEADER</span>{% endset %}{% set bf_children_s0 %}<b>REAL-CHILDREN</b>{% endset %}{{ bf.render_child('card', {'header': bf_prop_0, '_bf_slot': 's0', 'children': bf_children_s0}) | raw }}
```

**go-template** — `REAL-CHILDREN` is gone, `HEADER` is duplicated into `Children`:

```go
CardSlot0: NewCardProps(CardInput{
    …
    Header:   template.HTML("<span bf-s=\"" + scopeID + "\">HEADER</span>"),
    Children: template.HTML("<span bf-s=\"" + scopeID + "\">HEADER</span>"),
}),
```

Card's own generated template does `<div class="body">{{.Children}}</div>`, so the duplicate reaches the serialized HTML.

The one-named-prop case is the same defect with a milder symptom — `<Card header={<span>Hi</span>} />` (no children at all) still populates `Children` with the `header` payload, so content the caller never passed as children appears in the body.

Both cases reproduced directly against `compileJSX` with each adapter. Not yet executed through a real `go run` of the generated code: `template.HTML` fields are direct string substitution consumed by `{{.Children}}`, so the rendered string follows from the generated source, but that last step is inference, not measurement.

## Why no test sees it

This is the same class as #2753/#2762: **one decision, two implementations, no test comparing their answers.** The `data-key` relay had exactly this shape — Hono's inline predicate and the DSL adapters' `collectRootScopeNodes` had silently disagreed for as long as both existed, and it surfaced only when #2762 collapsed them and picked the wrong side. This one has not been collapsed yet, so it is still diverging.

The cross-adapter test added by #2772 (`root-key-relay-cross-adapter.test.ts`) pins the two families to the same answer for one shape only. Widening that pattern to this decision is what would have caught this.

## Not the same as #2703, #2746, or #2700

- **#2703** (`jsx-element-prop-fragment-conditional`) is the named-slot *bake* path (`go-template-adapter.ts:2313-2345`), where `extractTextChildren` → `extractHtmlChildren` → `extractScopedHtmlChildren` all fail and there is no fallback, so the field is never emitted. Its fixture passes a real nested child as well, so `comp.children.length > 0` and neither helper is reached at all.
- **#2746** is the silent drop when the bake itself fails — same family as #2703, different from a predicate that selects the wrong prop in the first place.
- **#2700** is the NewProps baker being static-only.

This one fires on the orthogonal precondition: the reserved slot is empty **and** a named JSX-valued prop exists.

## Suggested fix

Make `jsxChildrenPropNodes` ask the DSL adapters' question — match on the name as well as the kind. The seven copies and this one then agree by construction, which is the property worth having rather than the one-line change itself.

Landing that as a shared helper the eight adapters call, rather than an eighth copy that happens to agree today, is the version of this fix that cannot re-diverge.

kfly8:
## Summary

Fixes #2773. Two independently-maintained answers to "which prop supplies the reserved `children` slot?" had drifted apart:

- The seven DSL adapters (Blade/ERB/Jinja/Mojolicious/minijinja/Twig/Xslate) each carried a byte-identical private `resolveJsxChildrenProp` in their own `lib/ir-scope.ts`, matching the prop **named** `children`.
- The Go template adapter's `jsxChildrenPropNodes` answered a different question: the **first** prop of **any** name carrying a `jsx-children` payload.

Hono (the reference adapter) resolves `children` by name — ordinary JSX semantics, agreeing with the DSL adapters. For a component with an earlier JSX-element prop (`header={<span/>}`) and no `children` at all, go-template's predicate silently backfilled `.Children` with `header`'s payload; when a `children` prop was also given explicitly, it lost `children`'s own payload entirely in favor of the earlier prop's.

- Moved the single correct implementation into `@barefootjs/shared` (generic + structurally typed over `IRProp`/`IRNode`-shaped input, since `@barefootjs/jsx` already depends on `@barefootjs/shared` and importing the other direction would cycle). All eight adapters now call this one function.
- Deleted the seven per-adapter copies. Three `lib/ir-scope.ts` files (erb, mojolicious, xslate) carried nothing else and are removed entirely, along with their now-dangling imports. The other four (blade, jinja, minijinja, twig) keep their unrelated `extractTopLevelIdentifiers` helper, with the file-header docstring's "SHARED CANDIDATE... kept per-adapter for now" rationale updated to reflect the move.
- Fixed go-template's own predicate to call the shared function instead of its private `jsxChildrenPropNodes`.

## Fixture

Two new conformance fixtures under `packages/adapter-tests/fixtures/`, `expectedHtml` generated from the Hono reference (no hand-authoring, no `SKIP_AUTO_UPDATE` entry needed since Hono is correct here):

- `jsx-element-prop-explicit-children` — a named JSX-element prop (`header`) and an explicit `children` prop given together resolve to two distinct payloads.
- `jsx-element-prop-no-children` — a named JSX-element prop alone (no children of any kind) must not backfill the unrelated children slot. Its `Card` component guards the children interpolation with `?? ''` (matching `component-with-jsx-children`'s existing convention) so the fixture stays scoped to this predicate rather than an unrelated, pre-existing CSR quirk where a genuinely-undefined `children` value renders as the literal text `"undefined"` in the client hydration template (`ir-to-client-js`, outside this PR's scope).

Regenerated `packages/adapter-tests/coverage-map.json`, `ui/compat.lock.json`, and `ui/support-matrix.lock.json` for the two new fixtures.

## Verification

Reproduction, before/after (`header={<span>HEADER</span>} children={<b>REAL-CHILDREN</b>}`):
- **Before**: go-template emitted `Header: ...HEADER...` **and** `Children: ...HEADER...` — `REAL-CHILDREN` dropped entirely. Twig/Hono already resolved `header` and `children` as two distinct payloads.
- **After**: go-template now matches Twig/Hono — `Header: ...HEADER...`, `Children: template.HTML("<b>REAL-CHILDREN</b>")`.

Test suites (this environment lacks a `go` binary, so go-template's own real-Go-compilation tests skip — noted below rather than asserted):

| Suite | Result |
|---|---|
| `bun test packages/jsx/src` | 3178 pass, 28 skip, 3 todo, 0 fail |
| `bun test packages/client/__tests__` | 668 pass, 0 fail |
| `bun test packages/adapter-tests/src` | 2236 pass, 4 skip, 0 fail |
| `bun test packages/compat` | 508 pass, 0 fail |
| `generate-expected-html.ts` then `git status --short` | empty after commit |
| `adapter-hono` (`bun test`) | 1618 pass, 19 skip, 0 fail |
| `adapter-go-template` (`bun test`) | 1838 pass, 24 skip, 0 fail (real Go rendering skipped — `go` not installed here) |
| `adapter-blade` (`bun test`) | 1545 pass, 22 skip, 0 fail |
| `adapter-erb` (`bun test`) | 1516 pass, 21 skip, 0 fail |
| `adapter-jinja` (`bun test`) | 1545 pass, 22 skip, 0 fail |
| `adapter-mojolicious` (`bun test`) | 1714 pass, 34 skip, 0 fail |
| `adapter-rust` (minijinja, `bun test`) | 1543 pass, 22 skip, 0 fail |
| `adapter-twig` (`bun test`) | 1553 pass, 22 skip, 0 fail |
| `adapter-xslate` (`bun test`) | 1545 pass, 22 skip, 0 fail |

Perturbation (reverted only the shared predicate back to the old "first prop of any name" logic, kept the fixtures): re-running `adapter-twig`'s real-PHP conformance suite reddened exactly the two new fixtures —

```
(fail) JSX Conformance Tests > [jsx-element-prop-explicit-children] ...
(fail) JSX Conformance Tests > [jsx-element-prop-no-children] ...
 1476 pass / 22 skip / 2 fail (was 1478 pass / 0 fail with the fix)
```

— and directly recompiling the original repro through go-template's IR-to-Go-template codegen reproduced the exact original bug (`Children` backfilled with the `header` payload). Restored the fix afterward and re-confirmed `adapter-twig` back to 0 fail.

## Test plan

- [x] Reproduction before/after confirmed via direct `compileJSX` scripts (Twig, Hono, go-template)
- [x] `bun test packages/jsx/src`, `packages/client/__tests__`, `packages/adapter-tests/src`, `packages/compat` all green
- [x] `generate-expected-html.ts` produces no diff after commit
- [x] All nine adapter conformance suites run sequentially, all green
- [x] Perturbation confirmed the new fixtures redden without the fix (via Twig's real PHP backend and go-template's codegen output)

---


kfly8:
## Independent verification of this PR

Re-measured everything myself rather than taking the implementation report's numbers. They hold.

**The fix, on the original reproduction** (`<Card header={<span>HEADER</span>} children={<b>REAL-CHILDREN</b>} />`):

```
Go:  Header:   template.HTML("<span bf-s=\"" + scopeID + "\">HEADER</span>"),
     Children: template.HTML("<b>REAL-CHILDREN</b>"),
```

Matches Twig and the Hono reference. Before this change `Children` carried the `HEADER` payload and `REAL-CHILDREN` was dropped entirely.

**Perturbation** — reverted only `resolveJsxChildrenProp` in `@barefootjs/shared` back to "first prop of any name", left the fixtures alone, ran the Twig suite:

```
(fail) [jsx-element-prop-explicit-children] A named JSX-element prop and an explicit children prop resolve to two distinct payloads
(fail) [jsx-element-prop-no-children]       A named JSX-element prop alone must not backfill the unrelated children slot
 1551 pass / 22 skip / 2 fail   (1575 across 7 files)
```

Exactly the two new fixtures went red and nothing else. They pin the defect. Restored afterwards; the predicate is back to matching on name.

## A gap this PR surfaced, not caused

**`update-expected-html` has never run on this PR, and cannot.** `.github/workflows/update-fixtures.yml` is gated on `pull_request: branches: [main]`, and this PR's base is `claude/barefootjs-ir-carries-representation`. `list_workflow_runs` for `update-fixtures.yml` on this branch returns `total_count: 0`.

That check is the one that caught the #2753 regression on `main`, and this PR adds two fixtures plus regenerated `coverage-map.json` / `compat.lock.json` / `support-matrix.lock.json` — precisely its subject matter. So I ran it by hand in a real checkout of this head:

```
bun run packages/adapter-tests/scripts/generate-expected-html.ts
git status --short   # empty
```

**Drift is 0.**

Worth stating plainly for anyone reading this stack later: every intermediate PR in a stack is invisible to this check. The stack is still covered at the point it reaches `main` — the PR whose base is `main` runs it over the accumulated diff — but an intermediate PR's own drift is not reported while it sits open. Not this PR's to fix; recording it so the absence of a green `update-expected-html` here is not mistaken for it having passed.

One measurement in the PR body remains unverified for the same environmental reason it states: go-template's real Go compilation, since no `go` binary exists in the sandbox. Its conformance suite runs everything up to actual Go rendering.

## Follow-up filed

While reviewing, the `props.children ?? ''` guard in `jsx-element-prop-no-children` turned out to be dodging a real, separate defect rather than a style choice: an unguarded `{props.children}` renders the literal text `undefined` on CSR mount when the caller passes no children, while SSR renders empty. Filed as #2775 with the measurement. `2660ef0b8` adds that URL to the fixture's docstring — the guard is a deliberate scope boundary around a filed defect, and the next person to touch it needs to be able to tell that from a cleanup opportunity.

---

PR #2781, opened 2026-08-30
Ungate generated-artifact drift jobs from branches: [main]

### LINKED ISSUE #2780: Generated-artifact drift jobs are gated to `branches: [main]`, so no stacked PR is ever checked
Four workflows are gated on `pull_request: branches: [main]`. A PR whose base is another branch never triggers them, and GitHub reports no failure — the jobs simply do not exist on that PR. Every other CI workflow in the repo is ungated and runs on stacked PRs normally.

This is not a hypothetical. It is the mechanism by which #2762's row-key regression reached `main`, later fixed in #2772.

## The gated set

| Workflow | Verifies | Gated since |
|---|---|---|
| `update-fixtures.yml` | `packages/adapter-tests/fixtures/` `expectedHtml` matches what the reference adapter emits | `1c169bb41`, 2026-02-26 |
| `update-meta.yml` | `ui/meta/*.json` and `llms.txt` match `ui/components/ui/` | `5ded788d4`, 2026-05-20 |
| `update-doc-snapshots.yml` | `packages/jsx/src/__tests__/__snapshots__/doc-examples.test.ts.snap` matches `docs/core/**` | `32f798590`, 2026-05-23 |
| `changeset-check.yml` | a changeset exists for published-package changes | `45349c758`, 2026-05-25 |

The first three are the generated-artifact drift family. They are precisely the checks that detect *silent* divergence — output that changed without any test noticing, because the test's expectation is itself a generated file. Losing them on a stacked PR loses the only signal for that failure class.

Surveyed with the trigger block of every file in `.github/workflows/`: 25 workflows define a `pull_request` trigger; 21 are ungated, these 4 are gated.

## Measurement

Branch `claude/barefootjs-datakey-ir` (#2762, based on `claude/barefootjs-bf044-recursive`, merged to `main` on 2026-08-29):

- `ci.yml` (ungated) — `total_count: 5`, all green
- `update-fixtures.yml` (gated) — `total_count: 0`

Same branch, same five pushes. The ungated job ran every time; the gated one never ran once. #2762 was reviewed as green and merged; the fixture drift it introduced was invisible until it landed on `main` and the job finally ran with `main` as the base.

The bottom-of-stack PRs from the same series, whose base *was* `main` (#2752, #2761, #2772), all triggered `update-expected-html` and were green — so the gate is the whole difference.

## Why "it was green" is not evidence today

On a stacked PR, the checks list shows a set of passing jobs and no failures. There is no "skipped" entry, no neutral conclusion, nothing to notice. A reviewer who checks that CI is green gets a true answer to the wrong question: the drift job is not red, because it does not exist.

`CLAUDE.md` currently documents the manual workaround (run `generate-expected-html.ts` in a real checkout and confirm `git status --short` is empty before the work reaches `main`). That is a human step guarding a machine-checkable property, which is the wrong shape — this issue is about removing the need for it.

## What a fix has to satisfy

1. **The drift jobs run on every PR regardless of base.** Dropping `branches: [main]` is the obvious move and probably correct for the three generated-artifact jobs.
2. **A stacked PR's result must be interpretable.** A stacked PR's head contains its base branch's commits, so a drift failure may originate below it. The job's failure message should say which files drifted (it already prints the diff) so the author can tell "mine" from "inherited". Consider whether the check should compare against the merge-base rather than the head.
3. **No new noise on the bottom-of-stack PR.** `concurrency` is keyed on `github.event.pull_request.number`, so stacked and base PRs already get separate groups.
4. **Decide `changeset-check.yml` separately.** A stacked PR's changeset requirement is genuinely different from a main-based one — the changeset may legitimately live in the base PR. This may want to stay gated, or want a different rule. Do not lump it in with the other three without deciding.

## Verification

The fix is verified by reproducing the measurement above on a new stack: open a two-deep stack touching `packages/adapter-tests/fixtures/`, deliberately leave one fixture stale on the upper PR, and confirm the drift job runs and fails there. A fix that only changes the YAML without that demonstration has not been shown to work — the failure mode of this issue is a job that silently does not run.

kfly8:
Fixes #2780.

## What changed

Three of the four workflows gated to `pull_request: branches: [main]` are the generated-artifact drift family: `update-fixtures.yml`, `update-meta.yml`, `update-doc-snapshots.yml`. Each of them checks out the PR head, regenerates the artifact, and diffs it against what's committed — a property of the head tree alone. The base branch is not an input to that question, so gating them to a `main` base only meant the check silently never ran on a stacked PR. That is the mechanism by which #2762's fixture-drift regression reached `main` undetected (see the issue for the measurement: `update-fixtures.yml` had `total_count: 0` across five pushes to that branch, while the ungated `ci.yml` ran every time).

This PR drops the `branches: [main]` line from those three workflows' `on: pull_request:` blocks, and adds a short comment to each recording why the filter isn't there.

`changeset-check.yml` is intentionally left alone. Unlike the three drift jobs, its check *is* base-relative — it diffs `base.sha...head.sha` for both published source and added changesets. On a stack where the base PR carries the changeset and the upper PR carries the source change, the added-changeset diff comes back empty while the source diff doesn't, which would be a false failure. Making that stack-correct is a separate design question (per the issue's design-note comment), so this PR only adds a comment above its `branches: [main]` line recording that the gate is deliberate and pointing back at #2780.

**Also:** `update-fixtures.yml` didn't list its own workflow file in `paths`, unlike its two siblings (`update-meta.yml`, `update-doc-snapshots.yml`, which do list themselves). That meant it was the one drift workflow not exercised by a PR that edits it — including this PR itself, before the fix. That's part of why the separate stacked demo PR below was necessary to prove the `branches: [main]` removal actually took effect, rather than this PR's own checks showing it directly. Added `.github/workflows/update-fixtures.yml` to its own `paths` list to close that gap too.

## Verification

A YAML diff alone doesn't prove a previously-non-existent job now runs. This is verified with a real stacked-PR demonstration: a second draft PR (branch `claude/barefootjs-drift-gate-demo`, #2782) is stacked on top of this branch, deliberately corrupting one fixture's `expectedHtml`. The measurement — whether `update-fixtures.yml`'s `update-expected-html` job exists and fails on that stacked PR — is posted as a comment on this PR.

### DIFF
--- a/.github/workflows/changeset-check.yml
+++ b/.github/workflows/changeset-check.yml
@@ -3,6 +3,17 @@ name: Changeset check
 on:
   pull_request:
     types: [opened, synchronize, reopened, labeled, unlabeled]
+    # Kept gated, unlike the three drift jobs (#2780): those diff the
+    # head tree against itself, so a stacked base is not an input to
+    # the question they ask. This check is different — it diffs
+    # `base.sha...head.sha` for both published source and added
+    # changesets. On a stack where the base PR carries the changeset
+    # and the upper PR carries the source change, `BASE_SHA` already
+    # contains that changeset, so the added-changeset diff comes back
+    # empty while the source diff doesn't — a false failure on a stack
+    # that is collectively fine. Making this stack-correct means
+    # deciding what "a stacked PR inherits its base's changeset"
+    # should mean, which is a separate design question. See #2780.
     branches: [main]
 
 permissions:
--- a/.github/workflows/update-doc-snapshots.yml
+++ b/.github/workflows/update-doc-snapshots.yml
@@ -15,10 +15,13 @@ name: Update Doc Snapshots
 # inspect the bot's diff in the same way they inspect `update-fixtures`
 # bot commits — an unintended snap drift is a compiler regression
 # disguised as a passing test, not a "fix forward" signal.
+#
+# No `branches` filter: a stacked PR's head can drift from the
+# doc-examples snapshot the same as a main-based one, and this is the
+# only check that would catch it. See #2780.
 
 on:
   pull_request:
-    branches: [main]
     paths:
       - 'docs/core/**'
       - 'packages/jsx/**'
--- a/.github/workflows/update-fixtures.yml
+++ b/.github/workflows/update-fixtures.yml
@@ -1,8 +1,10 @@
 name: Update Fixtures
 
+# No `branches` filter: a stacked PR's head can carry the same
+# undetected fixture drift as a main-based one, and this is the only
+# check that would catch it. See #2780.
 on:
   pull_request:
-    branches: [main]
     paths:
       - 'packages/jsx/**'
       - 'packages/adapter-hono/**'
--- a/.github/workflows/update-fixtures.yml
+++ b/.github/workflows/update-fixtures.yml
@@ -11,6 +11,7 @@ on:
       - 'packages/client/**'
       - 'packages/adapter-tests/src/**'
       - 'packages/adapter-tests/scripts/**'
+      - '.github/workflows/update-fixtures.yml'
       # Shared-component fixture sources (fixture-hydrate corpus inputs).
       # Editing a `*.tsx` here regenerates the `__snapshots__/<id>.{html,client.js}` pair.
       - 'integrations/shared/components/**'
--- a/.github/workflows/update-meta.yml
+++ b/.github/workflows/update-meta.yml
@@ -8,10 +8,13 @@ name: Update Meta
 # regenerations. Issue #1435 was the visible symptom — `bf add
 # checkbox` lost its `icon` dependency because `ui/meta/checkbox.json`
 # still said `dependencies.internal: []`.
+#
+# No `branches` filter: a stacked PR's head can drift from
+# `ui/components/ui/` the same as a main-based one, and this is the
+# only check that would catch it. See #2780.
 
 on:
   pull_request:
-    branches: [main]
     paths:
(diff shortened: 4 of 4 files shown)


kfly8:
## Verification measurement

Stacked demo PR #2782 (branch `claude/barefootjs-drift-gate-demo`, base `claude/barefootjs-unfate-drift-jobs`) deliberately corrupted `packages/adapter-tests/fixtures/adjacent-conditionals.ts`'s `expectedHtml`.

`update-fixtures.yml` run on that branch: https://github.com/piconic-ai/barefootjs/actions/runs/33291399226

- **Existence** — `total_count: 1` for `update-fixtures.yml` filtered to branch `claude/barefootjs-drift-gate-demo`. This is the direct contrast with the pre-fix measurement in #2780: the same query on stacked branch `claude/barefootjs-datakey-ir` returned `total_count: 0` across five pushes. The job now exists on a stacked PR.
- **Detection** — job `update-expected-html` concluded `failure`. Verified from the job log, not just the conclusion field: install, build, and `Generate expectedHtml` all completed successfully; the failure is the drift-check step itself, whose diff shows the committed `DELIBERATELY-WRONG-FOR-2780-DEMO` marker regenerated back to the correct `B`, followed by `##[error]fixture expectedHtml is out of date`. So the failure is the detector working as intended, not an earlier step dying for an unrelated reason.

Both halves required by the issue are demonstrated: the gate is gone, and the drift check still catches real drift once it runs.

#2782 is being closed without merging now that it has served its purpose.

---

Issue #2785, opened 2026-08-30
`support-matrix:lock` silently writes a lock with only the built adapters, deleting the rest


kfly8:
`bun run support-matrix:lock` discovers adapters from their built `dist`. If some adapter packages are not built, it does not warn, fail, or refuse — it writes a complete-looking lock file containing only the adapters it happened to find, and reports the reduced count as though that were the expected result.

Committing that output deletes the missing adapters' rows from `ui/support-matrix.lock.json`. Since the lock is the artifact the freshness test compares against, the deletion becomes the new expectation, and the drift check then passes on it.

## Measured

Hit while regenerating the lock for #2783. The worktree had only `@barefootjs/jsx`, `vite`, `client` and `hono` built (the build step copied from `update-fixtures.yml`, which is all that workflow needs):

```
$ bun run support-matrix:lock
Wrote support matrix → ui/support-matrix.lock.json (1 adapters, 16 kinds, 57 axes)
```

The resulting diff against the committed lock:

```
842 insertions(+), 8946 deletions(-)
removed keys: 1360
/adapters: ['hono','blade','erb','go-template','jinja','minijinja','mojolicious','twig','xslate'] -> ['hono']
```

Eight adapters gone. Nothing in the output distinguishes this from a legitimate regeneration — `(1 adapters, …)` is the only signal, and it reads as a status line, not a warning.

After building every adapter package and re-running, the same command produced the correct result — 9 adapters, 0 keys added, 0 keys removed, 122 value changes, all attributable to that branch's fixture edits.

## Why this is worth fixing rather than documenting

This is the repo's own silent-divergence shape, in the tooling rather than the compiler: the output is wrong, nothing says so, and the check that would catch a wrong lock is comparing against the wrong lock. It is also the exact hazard `update-fixtures.yml`'s comment already describes for its own artifact ("auto-committing makes this job the detector for its own subject") — the same trap, one artifact over, but here it can be triggered from a developer's shell rather than only from CI.

The blast radius is larger than a stale count. `ui/support-matrix.lock.json` is what the compatibility matrix is published from, so a truncated lock understates which adapters support what — the kind of claim that is worse than useless when the project is trying to move to being used by other people.

## Suggested fix

[...]

PR #2787, opened 2026-08-30
Recognize a renamed `children` destructure at the bare splice

### LINKED ISSUE #2786: A destructured-and-renamed `children` still renders the literal text `undefined` on CSR mount
#2775 fixed the bare `{children}` passthrough splice for the two shapes the recognizer matches. A third shape — `children` destructured under a different name — is not matched and still stringifies `undefined` on a pure-CSR mount when the caller passes no children.

## Measured

On #2783's head, compiled with `compileJSX` + `TestAdapter`, reading the emitted `hydrate(... template: ...)` line:

| Source | Emitted splice |
|---|---|
| `props.children` | `${markupOrEmpty(_p.children)}` — guarded |
| `{ children }` (destructured) | `${markupOrEmpty(_p.children)}` — guarded |
| `const { children: kids } = props` then `{kids}` | `${(_p.children)}` — **unguarded** |

Same divergence as #2775: SSR and SSR+hydration render nothing, a pure CSR mount renders the four-character string `undefined`.

## Why the recognizer misses it, and why that is fixable

`isChildrenPassthroughExpr` (`packages/jsx/src/ir-to-client-js/html-template.ts`) tests `node.expr`, the ORIGINAL source text. For this shape that text is `kids`, which matches nothing.

But the third row above shows the compiler has **already resolved the alias**: the emitted expression is `(_p.children)`, not `kids`. So the information needed to recognize it is present at the point of decision — `bareSpliceExpr` receives that resolved `valueExpr` as its second argument and currently ignores it for recognition purposes.

Testing the resolved `valueExpr` (tolerating the wrapping parens) instead of, or in addition to, `node.expr` would close this. That keeps the same false-positive profile already argued in `isChildrenPassthroughExpr`'s docstring: an unrelated `.children` member that resolves the same way is harmless, because the branch never escapes its value either way and a nullish one rendering `''` rather than `"undefined"` is an improvement regardless.

Worth confirming while fixing whether the resolved form is stable across all four `case 'expression'` builders, since `node.expr` was chosen originally precisely because it is the one form that does not vary between them.

## Current exposure

Nothing in the corpus exercises this shape today, which is why #2783 is green:

- No unguarded `${_p.children}` / `${(_p.children)}` remains in any regenerated fixture snapshot or in `doc-examples.test.ts.snap`.
- No component under `ui/components/` or `site/ui/components/` uses a renamed children destructure. (The one grep hit, `input-group-playground.tsx:53`, is an object literal `children: [...]` inside a JSX-tree data structure — a different construct.)

So this is a latent gap in what the compiler accepts, not a live regression, and closing it should produce no artifact churn beyond whatever fixture pins it.

## Owed

Per the three-piece rule this needs a conformance fixture using the renamed-destructure source with `expectedHtml` generated from Hono (correct here, as in #2775), landing with the fix as its regression test.

Not folded into #2783: that PR is green with five regenerated artifacts audited against criteria fixed in advance, and the shape is pre-existing rather than introduced there. Raised by a review comment on #2783 — the docstring called the gap out honestly, but this repo's convention is that a reproducible defect lands as a fixture, not a prose note, which is what this issue exists to correct.

kfly8:
Fixes #2786. Follow-up to #2783, which fixed two of the three spellings.

`const { children: kids } = props` followed by `{kids}` still rendered the literal text `undefined` on a pure-CSR mount when the caller passed no children, while SSR and SSR+hydration rendered nothing.

## Emitted template

```js
// before
template: (_p) => `<section class="panel"><h3 …>${escapeTextOrMarkup(_p.title)}…</h3>${(_p.children)}</section>`
// after
template: (_p) => `<section class="panel"><h3 …>${escapeTextOrMarkup(_p.title)}…</h3>${markupOrEmpty((_p.children))}</section>`
```

## Why this is a fix, not a documented limit

`isChildrenPassthroughExpr` tested `node.expr` alone — the ORIGINAL source text, which reads `kids` here, so nothing matched. But the `before` line above shows the compiler had **already resolved the alias**: the expression about to be spliced reads `(_p.children)`, not `kids`. The information needed was present at the point of decision and simply unused.

`bareSpliceExpr` now tests the resolved expression as well. `node.expr` stays the primary test, because it is the one form that does not vary between the four template builders — dropping it would trade this gap for a different one.

## A docstring claim that measurement contradicted

The recognizer's docstring argued that a false positive — an unrelated `.children` member, a tree node's own `children` array — was harmless. Measured, it cannot arrive at all: a reactive member expression is given a `slotId` and takes the escaped text-slot branch, so it never reaches the bare-splice fallthrough this gates.

```js
// props.node.children — a different branch entirely
`<div bf="s1"><!--bf:s0-->${escapeTextOrMarkup(_p.node.children)}<!--/--></div>`
```

The comment now says that, rather than defending against a case that does not occur. The harmlessness argument was true but aimed at nothing.

## Coverage

`children-passthrough-renamed` pins the shape — deliberately unguarded and deliberately renamed, so the alias path has its own coverage instead of sharing `children-passthrough-no-children`'s. `expectedHtml` generated from Hono unmodified.

## Verification

| Suite | Result |
|---|---|
| `packages/client` | 668 pass / 0 fail |
| `packages/jsx/src` | 3201 pass / 28 skip / 3 todo / 0 fail |
| `packages/adapter-tests/src` | 2248 pass / 4 skip / 0 fail |
| `packages/compat` | 508 pass / 0 fail |

All five generated artifacts regenerated with every adapter built first (per #2785), and audited against criteria fixed before running:

- **No existing snapshot moved at all** — `packages/adapter-tests/fixtures/__snapshots__/` and `doc-examples.test.ts.snap` are both untouched. Nothing in the corpus used the renamed shape, so the recognizer change alters no existing emission. That is the measurement behind calling this a latent gap rather than a live regression.
- **`coverage-map.json`** — exactly one fixture added, zero existing entries changed, `uncoveredKinds` unchanged at `['regex']`.
- **`ui/support-matrix.lock.json`** — 9 adapters (unchanged set); across all 657 cells, zero where `delta(pass) != delta(total)` and zero whose failure count grew.
- **`ui/compat.lock.json`** — one line, `totalFixtures` 352 → 353.

Raised by a review comment on #2783. That PR documented the gap in a docstring; per this repo's convention a reproducible defect lands as a fixture, not a prose note, which is what this PR delivers.

---


kfly8:
The nitpick was right on both halves, and the second half caught something I had no way to see from my own measurements. Fixed on #2788 by rewriting the body rather than appending a third correction to it.

`go-template` ↔ `minijinja` wasn't quite a swap — it was two independent errors that happened to land on those two rows. The first table claimed go-template passes because the local sweep hid `Skipping […]: go command not found` behind a `3 pass / 0 fail` summary. minijinja was missing for a different reason: my sweep looped over `packages/adapter-<name>`, and minijinja has no such package — it is served by `packages/adapter-rust`, so the loop **structurally could not see it**. Eight of nine adapters measured, and nothing in the output said so.

Settled it from the repo's own artifacts instead of another sweep, since the fixture's arrival raises each kind's `total` by one and an adapter that renders it also gains a `pass`:

| adapter | `identifier` pass/total (main → head) | verdict |
|---|---|---|
| go-template | 309/330 → **309**/331 | fails |
| mojolicious | 312/330 → **312**/331 | fails |
| minijinja | 311/330 → 312/331 | passes |
| hono, blade, erb, jinja, twig, xslate | +1 | pass |

Seven of nine. Locally minijinja's case times out at exactly 5000 ms even warm, which is an environment artifact rather than a failure — worth stating because it is the third distinct way a local measurement misrepresented an adapter in this branch.

Nothing in this PR changes as a result; the pins were already correct for the two adapters that fail.


---

PR #2792, opened 2026-08-31
Escape a controlled textarea's value in the loop-row/branch builders

### LINKED ISSUE #2765: Loop-row builder does not escape a controlled textarea's child text, so a rebuilt row can break out of the element
`lowerFormControlValueSsr` (`packages/jsx/src/jsx-to-ir.ts`) lowers a controlled `<textarea value={…}>` into element content, and attaches **two** forms of the expression for exactly this reason. Its own docstring:

> The client-side registration template interpolates slotless expression children RAW (no text-slot `escapeText` wrapper), so a value containing `</textarea>` would break out of the element on CSR mount. Escape in the client-only template variant; `expr` stays clean for the SSR adapters, whose template engines (and hono/jsx) already escape text children natively.

So the IR carries `expr` (for SSR adapters, which escape natively) and `templateExpr: escapeText(<expr>)` (for client-side string building).

`generateCsrTemplate` uses `templateExpr` and is correct. **`irToHtmlTemplate`'s loop-row builder uses `expr`**, so it emits the value raw:

```js
__tpl.innerHTML = `<li …><textarea bf="s0">${String(val())}</textarea></li>`
```

A row whose value contains `</textarea>` therefore breaks out of the element — but **only when the row is rebuilt by the reconciler**. SSR is correct (engine-escaped), hydration reuses SSR markup, and the CSR mount template is correct. Only the rebuild path is affected, which is why the three-way contract does not catch it: all three legs can be correct at first render and diverge only after an interaction changes the row count.

## Family

Same family as #2756 / #2754: a client-side emitter not reading a decision the shared IR already made. #2756 was the attribute half of this exact lowering (the builder baked the `clientOnly` `value` attribute the IR had asked it to drop) and is fixed in #2764. This is the **child-text** half of the same lowering, left out of that PR because the fix is in a different position — the interpolation of expression children, not the attribute map.

## Suggested outcome

The loop-row builder should use `templateExpr` where the IR supplies one, the same way `generateCsrTemplate` does. Worth checking at the same time whether any other slotless expression child reaches a client-side string builder through `expr`, since the escaping obligation is a property of the destination (client string building), not of this one lowering.

A regression test needs to exercise the **rebuilt** row specifically — a fixture asserting first render will pass without the fix.

## Provenance

Found while implementing #2764; reported rather than fixed there to keep that PR's scope to the attribute-representation half. Not yet reproduced end-to-end in a browser — the mechanism is read from the emitted builder string and the IR's own docstring, so the "breaks out of the element" consequence is inferred, not observed. Confirming that is the first step of any fix.

kfly8:
Fixes #2765.

A keyed `.map()` row containing a childless controlled `<textarea value={...} />` broke out of the element when the reconciler rebuilt the row, if the value contained a literal `</textarea>`.

## Observed in a real browser — not inferred

#2765's own Provenance note says the break-out consequence was inferred from the emitted builder string and never seen. It happens, and the shape it takes defeats the obvious assertion.

Driven through `fixture-hydrate.playwright.ts` in Chromium: a keyed row's `<textarea>` reads back its full value correctly (the per-row binding repairs `.value` as a DOM property after the raw HTML parse), while the payload's `<b class="broke">X</b>` sits in the live DOM as a **sibling** of the closed-early element. Asserting the value alone passes; asserting the break-out's own artefact fails:

```
Error: expect(locator).toBeHidden() failed
Locator:  locator('.broke').first()
Expected: hidden
Received: visible
  - locator resolved to <b class="broke">X</b>
```

**The three-point oracle (SSR ≡ hydrated ≡ csr-mount) also misses this.** Hydrated and csr-mount both build the rebuilt row through the same loop-row builder, so they agree with each other on the wrong output — only SSR differs, and SSR has no rebuilt row to compare against. A structural limitation worth keeping in mind: comparison across legs that share an implementation can't see a defect in that shared part.

## Root cause

`lowerFormControlValueSsr` (`jsx-to-ir.ts`) already attaches two forms of the lowered value for exactly this reason: `expr` for SSR adapters (whose template engines escape text children natively) and `templateExpr` wrapped in `escapeText(...)` for client-side string building. The loop-row and conditional-branch builders in `ir-to-client-js/html-template.ts` (`irToHtmlTemplate` / `irToPlaceholderTemplate`) read `expr` — the SSR form — instead, so they spliced the value raw into an `innerHTML` string.

## Why the fix is a new IR flag, not `templateExpr` reuse

My first attempt used `templateExpr ?? expr` wholesale, matching how the whole-component template builder (`generateCsrTemplateWithOpts`) already resolves this expression. That broke `client-js-generation.test.ts`'s prop-defaulting pin: `templateExpr` also carries the `_p.`-prefixed **binding**, not just the escape, and this builder's output runs in **init scope**, where the correct reference is the init-scope local (`const config = _p.config`) rather than a bare `_p.config` re-read. Swapping the binding silently dropped a destructured prop's `?? {}` guard.

`generateCsrTemplateWithOpts` can safely take the whole `templateExpr` because its output is a module-scope template lambda with no init-scope locals to lose — same expression, different scope, not interchangeable.

The fix instead adds `escapeInClientTemplate?: boolean` to `IRExpression`, set only by `lowerFormControlValueSsr`, carrying exactly the one obligation that differs at this call site: escape the value, keep the binding.

## Fixture

`textarea-row-breakout` — a shared fixture with `interactions`, driven through both `fixture-hydrate.playwright.ts` and the oracle suite (`oracle.playwright.ts`). Asserts the value on both the adopted row and the rebuilt row, **and** the absence of the break-out's injected element — the assertion lesson from the false-negative above, recorded in the fixture's own docstring so it isn't rediscovered.

## An unrelated Go template adapter gap, pinned rather than worked around

The fixture seeds its payload from a module-level `const` (`createSignal(PAYLOAD)`), which is the natural way to keep the same literal shared between the signal's initial value and the retained-on-input handler. Go template's own real `go run` conformance test failed on this fixture: `convertInitialValue` (`value-lowering.ts`) has no resolution path for a signal seeded from a bare identifier referencing a module-level const — only a prop reference or a literal expression resolve — so it bakes to `nil` instead of the const's value, and the SSR-rendered `<textarea>` in the generated Go came back empty.

An earlier revision of this PR "fixed" the CI failure by inlining the literal into the fixture component instead of using a shared const. That was the wrong move — `packages/adapter-tests` fixtures are adapter-agnostic by design and should assert what SHOULD happen, not be shaped around one backend's current limitation. Reverted, and the actual gap is filed as #2794 and pinned in `packages/adapter-go-template/src/render-divergences.ts` (the single source both the conformance suite's `skipJsx` set and `ui/compat.lock.json`'s published `fixtureDivergences` derive from), same pattern as this file's existing entries.

## Verification

All adapters built first (per #2785's lesson), then every generated artifact regenerated and audited:

| Check | Result |
|---|---|
| Existing fixture `.client.js` snapshots changed (besides `textarea-row-breakout` itself) | **0** |
| `support-matrix.lock.json` | go-template's `pass` count for this fixture's cells drops by exactly 1 per cell (the new pin), `total` unchanged everywhere else |
| `compat.lock.json` | new `textarea-row-breakout` → `go-template` entry under `fixtureDivergences`, `totalFixtures` +1 |

| Suite | Result |
|---|---|
| `packages/jsx` (full) | 3199 pass / 2 fail (pre-existing, reproduces identically on `main` with no changes — confirmed a local-sandbox `BUN_OPTIONS=--smol` artifact, not present in real CI: `test (jsx-1)`/`test (jsx-2)` are both green on this exact base commit) |
| `packages/adapter-tests/src` (full) | 2254 pass / 0 fail |
| `packages/adapter-go-template` (full, real `go run`, run in isolation) | 1843 pass / 0 fail — `textarea-row-breakout` now cleanly skipped via the pin instead of failing |
| `packages/compat` (full) | 511 pass / 0 fail |
| Browser (`textarea-row-breakout`, 4 specs) | all pass, including the `.broke`-absence assertion |

Base includes the unrelated Workers Cache changes merged to `main` after this branch was created (`site/`/`integrations/` cache-control work) — merged in cleanly, no conflicts, no overlap with the compiler files this PR touches.

---


kfly8:
Confirmed and fixed in 365e866. `irToPlaceholderTemplate`'s `case 'expression'` bare-splice fallthrough now checks `escapeInClientTemplate` the same way `irToHtmlTemplate`'s `escapeForClient` does — verified the composite path was actually reachable and unescaped before the fix (reverted the change locally and inspected the generated row-construction template: `${value()}` raw vs. `${escapeText(value())}` after).

Added `textarea-row-breakout-composite` (a nested `<Tag>` component forces `useElementReconciliation`) as a shared fixture with the same `interactions`/browser coverage as the original, so both builders now carry regression armor. Full suites (`packages/jsx`, `packages/adapter-tests`, `packages/adapter-go-template` real `go run`, `packages/compat`) all green.

---

PR #2808, opened 2026-09-02
Fix #2763: fragment-bodied keyed .map() row loses its key

### LINKED ISSUE #2763: Fragment-bodied keyed `.map()`: the row key vanishes from every SSR adapter while the client template still bakes it
A `.map()` whose body returns a fragment, with `key={}` on an element inside it, emits no row-key attribute from any SSR adapter while the client row template bakes `data-key`. Same signature as #2753 Shape A, in a shape #2753 never measured.

## Reproduction

```tsx
"use client";
import { createSignal } from "@barefootjs/client";
export function KeyFrag() {
  const [items] = createSignal([{ id: 1, label: "a" }, { id: 2, label: "b" }]);
  return (
    <ul>
      {items().map(item => (
        <>
          <li key={item.id}>{item.label}</li>
          <li>x</li>
        </>
      ))}
    </ul>
  );
}
```

SSR (Hono reference, actual rendered output):

```html
<ul><li>a</li><li>x</li><li>b</li><li>x</li></ul>
```

Client row builder:

```js
__tpl.innerHTML = `<li>${escapeText(item().label)}</li><li>x</li>`
```

SSR puts nothing on the `<li>`; the client puts `data-key` on it. No diagnostic is emitted — no BF error, no warning.

## Mechanism

`extractLoopKey` (`jsx-to-ir.ts`) handles `element`, `component` and `conditional` but **not `fragment`**, so `key` resolves to `null`. Consequently:

- `loopKeyFn` is never set, so `mapArray` gets a null `getKey` and reconciles positionally;
- the call to `applyLoopKeyAttr` is skipped entirely by its `if (key !== null)` guard, so `IRElement.keyAttr` is never set and no adapter emits a row-key attribute.

The client template generator takes a different path: `html-template.ts`'s `a.name === 'key' ? keyAttrName(loopDepth) : toHtmlAttrName(a.name)` reads the raw `key` attribute directly and rewrites it to `data-key`. It never consults `extractLoopKey`'s result, so it bakes a key without knowing the key was not hoisted.

The raw `key` attribute does not rescue the SSR side either: Hono's JSX runtime deletes `key` from HTML output (see the comment in `renderComponentProps`, `hono-adapter.ts`), so it reaches no adapter's markup.

## Verified

- Emitter output is byte-identical before and after `claude/barefootjs-datakey-ir` (#2762, the family 2 fix). **This is not a regression from #2762 — it is a shape #2762 does not reach.**
- The pairwise corpus's `fragment-row-loop` axis is unkeyed, so the t=2 sweep never exercises this.

## Two defensible resolutions; the current silence is the only one ruled out

Either outcome satisfies sound-or-loud. What is not acceptable is the present behaviour, which is silent.

1. **Treat it as keyed.** Add a fragment case to `extractLoopKey` that takes the first element's `key`. `applyLoopKeyAttr` needs a fragment case too, applying to the row's first element — matching the "first element, not first node" rule `keyAttr`'s docstring already cites, and the position the client row template actually bakes. Note this **changes that `.map()` from positional to keyed reconciliation**, so it is a behaviour change, not a fill-in.
2. **Decide it is not keyed.** Treat a `key` on a non-root element of a multi-element row as an ambiguous statement of row identity and refuse it loudly with a BF diagnostic. Also stop the client template from baking `data-key`, since that is the side currently disagreeing with SSR.

Either way, `html-template.ts` reading the raw `key` independently of `extractLoopKey`'s decision needs to go. In the spirit of #2762 — which collapsed the row-key decision onto one IR field — the client template should read `IRElement.keyAttr` rather than interpreting the raw `key` attribute itself.

kfly8:
## Summary

Closes #2763. `extractLoopKey` (`jsx-to-ir.ts`) handled `element`, `component`, and `conditional` loop bodies but not `fragment`, so a `.map()` whose callback returns a fragment (`<><li>...</li><li>x</li>`) resolved its key to `null`. Every SSR adapter then silently omitted the row-key attribute and `mapArray` fell back to positional reconciliation, while `ir-to-client-js/html-template.ts`'s client row builders kept baking `data-key` by reading the raw `key` JSX attribute directly, independently of that decision — a silent SSR/CSR divergence with no diagnostic.

## The fix

`extractLoopKey` and its write-side twin `applyLoopKeyAttr` gained a `fragment` case that reads/stamps the key on the fragment's first ELEMENT child (skipping whitespace-only text), matching the "first element, not first node" rule `IRElement.keyAttr`'s docstring already documents for the render-root relay case (#2753). Four `html-template.ts` call sites (`irToHtmlTemplate`, `buildLoopSkeletonTemplate`, `irToPlaceholderTemplate`, `generateCsrTemplateWithOpts`) that read the raw `key` attribute independently now go through a new `resolvedKeyAttrName` helper that consults `IRElement.keyAttr` instead — in the spirit of #2762's consolidation — so a row the SSR side doesn't recognize as keyed no longer gets a `data-key` baked into it client-side either. (A fifth raw-`key`-reading site exists in `irToComponentTemplateWithOpts`, but it's dead code for any loop-containing tree — `canGenerateStaticTemplate`'s `case 'loop': return false` gates both of its callers away first — so it needed no change.)

## A latent bug this surfaced

Chasing the new fixture to a byte-exact `expectedHtml` caught an unrelated, pre-existing bug: both the Hono and go-template adapters emitted their multi-root Fragment per-item start marker (#1212) as `bfComment('bf-loop-i')` / `{{bfComment "bf-loop-i"}}`, even though `bfComment` already prepends `bf-` itself — doubling the prefix, unlike every other call site and the client runtime already agree on. No prior fixture combined a multi-root Fragment row with an SSR render on either adapter, so this went unexercised until now. Fixed on both adapters; `multi-root-loop-body.test.ts`'s pinned assertions (which had hardcoded the buggy string as expected) updated to match.

## Coverage

New fixture `fragment-body-keyed-loop-row` (`adapter-tests`) pins the SSR/CSR contract for this shape (its `expectedHtml` auto-regenerated from the fixed Hono reference). New unit test `fragment-body-loop-key.test.ts` (`jsx`) pins both the IR-level `keyAttr` resolution and the resulting `mapArray` keyFn directly.

## Verification

- `packages/jsx` full suite: 3166 pass / 0 fail
- `packages/adapter-tests` full suite: 2297 pass / 0 fail
- `packages/adapter-go-template` full suite: 1861 pass / 0 fail (real `go run` steps skipped in this sandbox — no `go` binary; CI covers them)
- `packages/compat` full suite: 514 pass / 0 fail (after regenerating `ui/support-matrix.lock.json` and `ui/compat.lock.json`)
- `generate-expected-html.ts` / `coverage-map.ts` re-run, only the new fixture's entries changed
- Changesets added: `@barefootjs/jsx` (patch), `@barefootjs/hono` + `@barefootjs/go-template` (patch)

---

### DIFF
--- a/.changeset/fragment-body-loop-key-2763.md
+++ b/.changeset/fragment-body-loop-key-2763.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix #2763: a `.map()` whose callback body returns a fragment (`<><li key={..}/><li/></>`) now resolves its row key instead of silently losing it. `extractLoopKey` and its write-side twin `applyLoopKeyAttr` gained a `fragment` case that reads/stamps the key on the fragment's first ELEMENT child (skipping whitespace-only text), matching the "first element, not first node" rule `IRElement.keyAttr`'s docstring already documents for the render-root relay case. Previously `extractLoopKey` returned `null` for a fragment body, so every SSR adapter omitted the row-key attribute and `mapArray` reconciled positionally — while `ir-to-client-js/html-template.ts`'s client row builders kept baking `data-key` from the raw `key` JSX attribute independently of that decision, a silent SSR/CSR divergence with no diagnostic. Those builders now read `IRElement.keyAttr` (via a new `resolvedKeyAttrName` helper) instead of the raw attribute, so a row the SSR side doesn't recognize as keyed no longer gets a `data-key` baked into it client-side either.
--- a/.changeset/loop-item-marker-double-prefix.md
+++ b/.changeset/loop-item-marker-double-prefix.md
@@ -1,0 +1,6 @@
+---
+"@barefootjs/hono": patch
+"@barefootjs/go-template": patch
+---
+
+Fix a latent SSR-emission bug in the per-item start marker for multi-root Fragment loop bodies (#1212): both adapters called `bfComment('bf-loop-i')` / `` {{bfComment "bf-loop-i"}} `` even though `bfComment` itself already prepends `bf-`, doubling the prefix to `<!--bf-bf-loop-i-->` instead of the correct `<!--bf-loop-i-->` the client runtime and every other adapter's whole-item-conditional anchor (`bf-loop-i:KEY`) already use. No prior fixture exercised `bodyIsMultiRoot` on either adapter, so this went unexercised until #2763's fragment-bodied-keyed-loop fixture was the first to combine a multi-root Fragment row with an SSR render on these adapters, surfacing it as a byte-for-byte `expectedHtml` mismatch.
--- a/packages/adapter-go-template/src/adapter/go-template-adapter.ts
+++ b/packages/adapter-go-template/src/adapter/go-template-adapter.ts
@@ -7319,7 +7319,10 @@ export class GoTemplateAdapter extends BaseAdapter implements ParsedExprEmitter,
    * `mapArrayAnchored` can hydrate items that render no element.
    */
   private loopItemMarker(loop: { bodyIsMultiRoot?: boolean; bodyIsItemConditional?: boolean; key?: string | null }): string {
-    if (loop.bodyIsMultiRoot) return `{{bfComment "bf-loop-i"}}`
+    // `bfComment` prepends `bf-`, so the argument is `"loop-i"`, not
+    // `"bf-loop-i"` (which would double the prefix to `<!--bf-bf-loop-i-->`
+    // — the same latent bug #2763's fixture caught on the Hono adapter).
+    if (loop.bodyIsMultiRoot) return `{{bfComment "loop-i"}}`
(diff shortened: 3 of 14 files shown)


kfly8:
Thanks for the review. Addressed both notes:

- `applyLoopKeyAttr`'s docstring updated (93855e5) — it was still describing the pre-fix behavior it had just replaced.
- PR description corrected to "four" sites and added a note on why the fifth (`irToComponentTemplateWithOpts`) needed no change — confirmed it's dead code for any loop-containing tree since `canGenerateStaticTemplate`'s `case 'loop': return false` gates both of its callers away first.

Left the missing e2e coverage for the marker double-prefix fix as-is — it's a real gap but pre-existing and orthogonal to this PR's scope (SSR byte-exact + unit-level client-JS assertions already pin the fix).

---

---

Issue #2737, opened 2026-08-27
alias-props mutant leaves a bare `props` reference in a SolidJS-style child's CSR template


kfly8:
## Problem

Found while fixing #2723 (whose two actual defects — the effect-wrap decision and the rest-spread name check not following a `const x__alias = x` hop — are now fixed). After that fix, `reactive-props × alias-props` still fails the `three-point` oracle, but with a **different** shape than #2723's: the **hydrated** leg renders `""` (empty), while **csr-mount** renders the full correct DOM.

## Root cause (confirmed via direct compiler output)

`ReactiveChild(props: ChildProps)` — a SolidJS-style (non-destructured) component — has its whole `props` parameter aliased by the `alias-props` mutation (`const props__alias = props`, every body reference rewritten to `props__alias`). Compiling the mutated source (both `TestAdapter` and `HonoAdapter` — the client JS emission is adapter-agnostic) produces:

```js
hydrate('ReactiveChild__bc9226be', {
  init: initReactiveChild,
  template: (_p) => `<div class="reactive-child">...${escapeTextOrMarkup((props).label)}...${escapeTextOrMarkup((props).value)}...</div>`,
  name: 'ReactiveChild',
})
```

The **template** string references the bare identifier `props` — not `_p.label` / `_p.value` — even though the template closure's only parameter is `_p`. `props` is not in scope there, so building this component from scratch (CSR-mount, `createComponent()`) throws a `ReferenceError`, and hydration for the surrounding component tree ends up producing empty markup on the hydrated leg (both legs read this same broken client JS, but the empty/full split needs more digging — noted here as observed, not fully traced end-to-end).

The `init` function itself is fine (`const props__alias = _p` is correctly declared, and every `init`-scope read of `props__alias.xxx` resolves fine there).

For comparison, a *sibling* SolidJS-style component in the same source file, `PropsStyleChild(props: PropsStyleChildProps)`, does NOT hit this — its template correctly emits `(_p).label` / `(_p).value`. The two only differ in that `ReactiveChild` also has an event handler (`onClick={() => props.onIncrement()}`), which may route it through a different template-generation path (`generateCsrTemplate` vs `irToComponentTemplate`) with its own copy of the `propsObjectName` → `_p` access-prefix rewrite (`html-template.ts`'s `transformExpr`, "Normalize source-level props object access" step) that doesn't yet follow the same alias hop.

## Where to look

[...]

Issue #2723, opened 2026-08-26
createComponent renders empty when const statements precede the component body's original declarations


kfly8:
## Problem

Inserting semantically inert `const x__alias = x` hops at the TOP of a component body (before the original signal/derivation statements) makes `createComponent()`'s pure-CSR mount render **completely empty**, while SSR and hydration stay correct. Five mutants reproduce it (`button`, `input`, `kbd`, `label`, `textarea` × `alias-props`).

The CSR "build from scratch" codegen apparently assumes the leading statements of the body are the destructure/signal declarations it knows how to replay — a preceding unrelated statement derails template construction entirely. User impact: any component whose author writes an ordinary leading `const` (a computed label, a helper alias) silently breaks pure-CSR mounting while SSR keeps working.

Found by the #2481 mutation sweep (`alias-props` mutation) three-point oracle.

## Repro / pins

Quarantined in `packages/adapter-tests/e2e/mutation-quarantine.ts` (rot-checked; deterministic regeneration via `test:mutation`).

## Graduation

Make CSR template construction tolerate leading non-reactive statements (or classify precisely what it replays), confirm the five mutants pass, delete their ledger entries.

kfly8:
Measured before scheduling the fix, and **the mechanism in the body above is wrong** — it would send an implementer to the wrong file. Correcting it here; the original text is left intact above.

## What actually happens

The CSR template is **correct in every variant**. Nothing is wrong with template construction, and nothing "renders empty" because of it. What the alias hop destroys is `init`.

Five variants of the `label` shape, compiled through `TestAdapter`:

| variant | `initL` | `createEffect` | `applyRestAttrs` |
| --- | --- | --- | --- |
| A — base, no alias | full body | ✅ | ✅ |
| B — alias `className` + `children` | `function initL() {}` | ❌ | ❌ |
| C — alias `className` only | `function initL() {}` | ❌ | ❌ |
| D — alias, **no** rest spread | `function initL() {}` | ❌ | ❌ |
| E — alias **+ an event handler** | full body | ❌ | ✅ |

The compiled template for B is byte-identical in shape to A's, `bf="s0"` marker and all:

```js
hydrate('Label', { init: initLabel, template: (_p) => `<label data-slot="label" ${…class…} bf="s0">${(_p.children)}</label>` })
function initLabel() {}
```

## The real mechanism

[...]

PR #2812, opened 2026-09-02
Fix #2806: direct rest-bag read in JSX leaks unbound `rest` into CSR template

### LINKED ISSUE #2806: Client JS: a component reading its rest-bag spread (`{ children, ...rest }`) in its own body emits an undeclared `rest` reference
Surfaced by `jsx-element-prop-rest-bag-dynamic` (`packages/adapter-tests/fixtures/jsx-element-prop-rest-bag-dynamic.ts`, added in #2804 as a go-template regression fixture — the SSR side of this fixture is intentional and correct on every adapter except go-template, which loudly refuses per #2805).

## The gap

A component that destructures its rest-bag spread and reads it in its own JSX body —

```tsx
function Card({ children, ...rest }: { children?: any; [key: string]: any }) {
  return <header>{rest.header}</header>
}
```

— compiles to client JS that references `rest` without ever declaring/binding it in scope. The client-JS scope gate (`packages/adapter-tests/src/__tests__/client-js-scope.test.ts`) catches this mechanically: TS diagnostic 2304 ("Cannot find name 'rest'") on the emitted CSR template. It is a guaranteed `ReferenceError` the moment that code path runs on a real page.

This is the same root class already tracked for other rest-bag shapes (`jsx-spread-props-object` #1407, `rest-spread-child-attrs` #2131, `stateless-rest-spread-forward` #2754): `applyRestAttrs`/the rest bag is not modeled at CSR template-eval scope. This fixture's specific manifestation — reading the rest bag directly in JSX rather than spreading it onto an element's attrs — hadn't been exercised in the shared corpus before.

## Current state

- `jsx-element-prop-rest-bag-dynamic` is added to `CSR_SKIP_FIXTURES` (`packages/adapter-tests/src/csr-skip-set.ts`) and pinned in `KNOWN_UNDECLARED` (`packages/adapter-tests/src/__tests__/client-js-scope.test.ts`, `names: ['rest']`) pointing at this issue.
- SSR is correct on every adapter (Hono, ERB, Jinja, Mojolicious, MiniJinja, Twig, Xslate, Blade); go-template refuses loudly with `BF101` per #2805 (a separate, orthogonal gap — go-template has no delivery route for this shape at all yet, client-JS soundness is unrelated).

## Graduation

Fix the client-JS emitter to bind the rest-bag variable in scope wherever the component body reads it directly (not just where it's spread onto an element), regenerate, delete both pins.

Tier: `+ bug` (a silent divergence — the emission is simply wrong, not a documented refusal) per CLAUDE.md's known-limitation classification.

kfly8:
## Summary

Closes #2806. Stacked on #2811 (#2737) — same class of gap, same fix mechanism, next PR in the sequence.

A component reading its own rest-bag spread directly in JSX (`{rest.header}`, not spread onto an element's attrs via `{...rest}`) emitted a CSR template referencing the bare, unbound `rest` identifier — a guaranteed `ReferenceError` the moment that code path runs on a real page. The client-JS scope gate (`client-js-scope.test.ts`) caught this mechanically (TS diagnostic 2304).

- **Root cause**: the rest binding is `_p` at runtime — the init body already rewrites `rest.x` → `_p.x` via `rewritePropsObjectRef` (#2723), and `applyRestAttrs` excludes the consumed keys by name from that same object rather than constructing a narrower one. But the four CSR/static template builders in `html-template.ts` — the exact same sites #2737 just migrated onto `rewritePropsObjectRef` for `propsObjectName` — were still passing `null` for the rest name.
- **Fix**: thread `ctx.restPropsName` through as a new `TemplateOptions.restPropsName` field to all four sites. The static-template gate widens from "has a props object" to "has a props object OR a rest binding" — kept minimal (not made unconditional) so it doesn't change output for components with neither.
- **Not `BindingScope`**: this isn't a shadow-guard question (no "is this name bound here" check needed) — the shadow-guard case (a `.map()` row parameter literally named `rest`) already works via the `enclosingScope` option #2737 added to the same function. This fix is purely "thread an argument the function already accepts."
- **Fixture**: `jsx-element-prop-rest-bag-dynamic` (added in #2804) is the graduation fixture per the issue — no new fixture needed (not a subset extension; `coverage-map.json` already covers this shape). SSR is unaffected; regenerating `expectedHtml` produced zero diff.

Investigation and design for this fix were done by a separate Fable-model agent per this stack's process, verifying the root cause against the actual current codebase (compiled the issue's repro, confirmed the TS2304 diagnostic, traced both the working init-body path and the broken template path to their exact line numbers) before any code was written.

## Test plan

- [x] New compiler unit test: `packages/jsx/src/__tests__/issue-2806-rest-bag-template-rewrite.test.ts` (direct rest-bag text read, rest-bag read inside a `.map()` row, and the shadow-guard case where a loop parameter is literally named `rest`).
- [x] `client-js-scope.test.ts`: 358 pass, 0 fail (pin removed, gate no longer flags `rest` as undeclared).
- [x] `csr-conformance.test.ts`: green (one `toggle-shared` timeout under parallel load, reproduced as a pass in isolation — unrelated pre-existing flake, not caused by this change).
- [x] Full `packages/jsx/src/__tests__` suite: 3175 pass, 0 fail.
- [x] Full `packages/adapter-tests/src/__tests__` suite: 2303 pass, 0 fail.
- [x] `generate-expected-html.ts` regenerated — zero diff (SSR unaffected).
- [x] Changeset added (`@barefootjs/jsx` patch).

---

### DIFF
--- a/.changeset/rest-bag-template-rewrite-2806.md
+++ b/.changeset/rest-bag-template-rewrite-2806.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix #2806: a component reading its own rest-bag spread directly in JSX (`{rest.header}`, not spread onto an element's attrs via `{...rest}`) emitted a CSR template referencing the bare, unbound `rest` identifier — a guaranteed `ReferenceError`. The rest binding is `_p` at runtime (the init body already rewrites `rest.x` → `_p.x` via `rewritePropsObjectRef`, #2723, and `applyRestAttrs` excludes the consumed keys by name from that same object rather than constructing a narrower one), but the four CSR/static template builders in `html-template.ts` were calling `rewritePropsObjectRef` with `null` for the rest name — the same four sites #2737 just migrated onto this shared door for `propsObjectName`. They now thread `ctx.restPropsName` through as well, so a direct rest-bag read resolves the same way in the template as it always did in init.
--- a/packages/adapter-tests/src/__tests__/client-js-scope.test.ts
+++ b/packages/adapter-tests/src/__tests__/client-js-scope.test.ts
@@ -55,16 +55,12 @@ const KNOWN_UNDECLARED: Record<string, KnownHole> = {
   // map-index-handler, reactive-props, props-reactivity-comparison)
   // graduated with the memo/`templateExpr`/getter-elided-signal emission
   // fixes; see the issue for the closing PR.
-  'jsx-element-prop-rest-bag-dynamic': {
-    // #2806: a component reading its own rest-bag spread directly in JSX
-    // (`{rest.header}`) emits a CSR template referencing `rest`, which is
-    // never bound in that scope — the same `applyRestAttrs`-not-modeled
-    // class as the CSR_SKIP_FIXTURES entries above it, manifesting here as
-    // a genuine undeclared-identifier compile hole rather than a harness
-    // limitation.
-    names: ['rest'],
-    issue: 'https://github.com/piconic-ai/barefootjs/issues/2806',
-  },
+  // #2806 (a component reading its own rest-bag spread directly in JSX,
+  // `{rest.header}`) is FIXED — the four template builders in
+  // `html-template.ts` now thread `restPropsName` into the same
+  // `rewritePropsObjectRef` door the init body already used (#2723), so
+  // `rest.x` resolves to `_p.x` in the template the same way it always did
+  // in init.
 }
 
 /** One virtual .ts file per emitted client-JS artifact. */
--- a/packages/adapter-tests/src/csr-skip-set.ts
+++ b/packages/adapter-tests/src/csr-skip-set.ts
@@ -90,13 +90,4 @@ export const CSR_SKIP_FIXTURES: ReadonlySet<string> = new Set([
   // child needs no `$c` lookup at all (it IS `__scope`) — see
   // `ClientJsContext.commentScopeRootSlotId` and
   // `comment-wrapper-grandchild-slot-collision.test.ts`.
-  // #2806: same `applyRestAttrs`-not-modeled class as `jsx-spread-props-object`
-  // / `rest-spread-child-attrs` / `stateless-rest-spread-forward` above, but
-  // reading the rest bag directly in JSX (`{rest.header}`) rather than
-  // spreading it onto an element's attrs — the emitted CSR template
-  // references `rest` without ever binding it, a guaranteed
-  // `ReferenceError` (also pinned in `client-js-scope.test.ts`'s
-  // `KNOWN_UNDECLARED`). SSR is correct on every adapter; go-template's own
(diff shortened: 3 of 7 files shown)


kfly8:
Thanks for the detailed trace — the four `lazy-row-eligibility.ts`/`build-lazy-row.ts`/`stringify/lazy-row.ts` citations are accurate as far as they go, but the conclusion doesn't hold: the lazy-row path does **not** reproduce the #2806 bug, and I verified this empirically rather than by re-reading the same call sites.

The four gate/emitter facts above are all about how `lazyRowEligibility`/`classifyLazyBinding`/`stringify/lazy-row.ts` *classify and splice* a binding — but they aren't the pass that fixes #2806. That pass is a separate, pre-existing, whole-function-body rewrite: [`generate-init.ts:100`](https://github.com/piconic-ai/barefootjs/blob/9c7d7980e4efce4675bc426a3e6cd7e1dd66f144/packages/jsx/src/ir-to-client-js/generate-init.ts#L100), `rewritePropsObjectRef(lines.join('\n'), ctx.propsObjectName, ctx.restPropsName)`, which has walked the ENTIRE `init<Component>` body text (createRow/applyItem/applyOuter closures included) since #2723 — well before this PR. It runs regardless of which loop-row emitter produced the body, so a raw `rest.suffix` spliced by `stringify/lazy-row.ts` still gets rewritten to `_p.suffix` at this later stage. This PR's own fix only needed to patch the one line appended *after* that pass (`hydrate(... template: ...)` at line 101), which is the one spot the whole-body rewrite doesn't reach.

I confirmed this by actually compiling the repro shape from your comment (signal-backed array source, keyed single-root row, no nested component/ref/preamble, row body reading `rest.suffix`) on both this PR's branch and pre-#2806-fix `main`:

- It compiles through `mapArrayLazy` as predicted (lazy-eligible).
- The emitted `createRow`/`applyOuter` bodies read `_p.suffix`, **not** `rest.suffix`, on *both* `main` (before this PR) and this branch (after it) — the whole-init-body rewrite already produced correct output for the lazy path before this PR existed, and this PR doesn't touch it.

So the "opaque outer identifier" analysis in your comment is correct about the *eligibility gate's own classification* not knowing about `restPropsName` — but that only affects whether the loop is *classified* lazy-eligible, not whether the emitted reference is bound correctly once it lands in the init body.

I've added regression coverage for this to make it explicit going forward: two new test cases in `issue-2806-rest-bag-template-rewrite.test.ts` asserting `_p.suffix` (and no bare `rest.suffix`) in both the `mapArrayLazy` output (signal-backed source, no child component) and the `mapArray` output (same shape, forced eager via a child component in the row), plus an inline note on the existing 2nd test case explaining why it doesn't cover either path (its array source, `rest.items ?? []`, is classified `isStaticArray` and takes an unrelated `buildStaticLoopPlan` route). Pushed as `f07e41039`.

---

Issue #2813, opened 2026-09-02
SSR: a signal-getter alias as a `.map()` source (`const items__alias = items`) fails to render on every non-Hono adapter


kfly8:
## The gap

A `.map()` loop whose source is a local `const` alias of a signal getter (`const items__alias = items; items__alias().map(...)`) renders correctly on Hono (it's real JS — the getter call executes) but fails on every other adapter, which lower the loop source as a static field/data reference rather than executing an alias hop:

```tsx
"use client";
import { createSignal } from "@barefootjs/client";
export function AliasedLoopSource() {
  const [items] = createSignal([{ id: 1, label: "a" }, { id: 2, label: "b" }]);
  const items__alias = items;
  return <ul>{items__alias().map(it => <li key={it.id}>{it.label}</li>)}</ul>;
}
```

Measured against the `aliased-loop-source` fixture (`packages/adapter-tests/fixtures/aliased-loop-source.ts`, added alongside #2778's CSR-side fix):

- **ERB**: real Ruby render throws `undefined method 'length' for nil (NoMethodError)`.
- **go-template**: `template error: ... can't evaluate field Items__alias in type main.AliasedLoopSourceProps` — `queryHref`-style zero-arg-call-to-field lowering (`go-template-adapter.ts`'s call handling) maps `items__alias()` to a `.Items__alias` struct field that was never seeded, since the seeding logic only knows about `items`, the real signal name.
- **Mojolicious**: template render throws (empty loop body, undefined array).
- **Blade, Jinja (Twig-style), MiniJinja (Rust), Twig, Xslate**: all compile clean but SSR an empty `<ul>` — the loop body never executes because the seeded data under `items__alias` is absent (only `items` was seeded).

So this is not one adapter's bug — it's the SSR-side twin of #2778: every non-Hono adapter resolves a loop's array-source identifier by name lookup against seeded/struct data keyed by the signal's REAL name, with no alias-hop resolution. Hono's reference is unaffected because it's the only adapter that runs the actual expression as JS.

## Why this is a separate issue from #2778

#2778 is the CSR client-JS template's `(undefined)()` crash — fixed by resolving the alias at CSR-substitution time (`csr-substitute.ts`'s `resolveGetterAliases`, reusing the `resolveAliasOrigin` hop-walker `props-binding.ts` already had for the props/rest-alias case). That fix is purely about the client-JS emission path and does not touch SSR data-seeding at all.

[...]

Issue #2741, opened 2026-08-28
queryHref effect-update emit prop-rewrites object-literal keys, producing syntactically invalid client JS


kfly8:
## Symptom

Compiling a component whose attribute carries a `queryHref(base, { … })` call emits client JS whose `createEffect` update path prop-rewrites the object literal's **keys**, not just its values:

```js
createEffect(() => {
  if (_s0) {
    { const __v = queryHref(_p.base, { _p.tag: _p.tag, _p.page: _p.page ? _p.page : '' }); … }
  }
})
```

`{ _p.tag: _p.tag }` is a syntax error, so the whole client module fails to parse. The same compile also logs `[barefootjs] pruneUnusedPropExtractions: generated code did not parse; skipping prune` — the prune pass's parse failure is a secondary symptom of the same broken emit.

The `hydrate(...)` template lambda in the same output is **correct** (`{ tag: _p.tag, page: _p.page ? _p.page : '' }`); only the effect-update path is broken. No diagnostic is raised — the invalid module is written out silently, which puts this in the silent-divergence (`+ bug`) tier.

## Repro

Compile `packages/adapter-tests/fixtures/query-href.ts`'s source with the Hono adapter and inspect the `clientJs` output, or run CSR conformance for the `query-href` fixture — it throws at module parse.

## How it stayed hidden

`queryHref` shipped (#2042/#2057) with recognition tests only and **zero conformance fixtures** — the change-time coupling rule's plugin half had no mechanical backstop, and review didn't catch the missing fixture. The builtin-lowering coverage floor added the missing fixture, which surfaced this immediately.

## Pins (three-piece set)

- Conformance fixture `query-href` asserts the correct SSR output (auto-generated from the reference adapter, which is sound — no `SKIP_AUTO_UPDATE` needed).
- CSR conformance skip: `query-href` entry in `packages/adapter-tests/src/csr-skip-set.ts` pointing at this issue.

Graduation: fix the effect-path prop rewrite to leave object-literal keys alone, delete the skip entry — the fixture then stands as the regression test.

Issue #2468, opened 2026-08-02
CSR template lambdas reference init-scoped bindings — guaranteed ReferenceError on CSR mount (scope-gate inventory)


kfly8:
Surfaced mechanically by the new client-JS scope gate (`packages/adapter-tests/src/__tests__/client-js-scope.test.ts`), which type-checks every fixture's emitted client JS for out-of-scope identifier references (TS 2304/2552). Any hit is a guaranteed `ReferenceError` the moment the CSR `template:` lambda is evaluated — silent today because nothing checks the compiler's emitted output and `csr-conformance.test.ts` skips the affected fixtures.

## Inventory (undeclared identifiers per fixture)

| Fixture | Undeclared names |
|---|---|
| `button` | `className`, `size`, `variant` |
| `tooltip` | `className`, `size`, `variant` |
| `kbd` | `className` |
| `command` | `className`, `setOpen`, `size`, `variant` |
| `map-index-handler` | `setActive` |
| `reactive-props` | `value` |
| `props-reactivity-comparison` | `value` |

(Not listed here: the `search-params*` family — same mechanical symptom, already tracked as #2075's env-signal-getter contract — and `signal-early-return`, whose template-scope leak is one facet of #2463.)

## Cause

The `hydrate('X', { init, template })` registration emits `template` as a module-scope lambda, but the referenced bindings (destructured props like `className`/`size`/`variant`, signal setters like `setOpen`/`setActive`, locals like `value`) are declared inside `init` (or nowhere in the emitted module). The lambda can only work by accident of never being evaluated — SSR-hydration paths that reuse existing DOM never call it, which is why this stayed invisible.

## Expected

Either (a) the template lambda closes only over `_p` and its own parameters — destructured props are read as `_p.x`, handlers are not embedded in template output — or (b) shapes whose template cannot be made self-contained are refused loudly (BF diagnostic) instead of emitting a lambda that throws.

## Ledger

Each fixture above is pinned in the scope gate's `KNOWN_UNDECLARED` set with this issue's URL. The pin asserts the exact undeclared-name set, so a partial fix fails loudly and a full fix forces the pin's deletion — the set can only shrink.

PR #2814, opened 2026-09-02
Fix #2778: signal/memo getter alias resolves to undefined in CSR template

### LINKED ISSUE #2778: A local-const alias for a `.map()` source is substituted as `undefined` in the CSR template — `(undefined)()` throws on pure CSR mount, with no diagnostic
## The divergence

A component that aliases its loop source through a local const before iterating compiles clean, and the client template it emits calls `undefined` as a function. Pure CSR mount throws `TypeError` immediately. No BF diagnostic, no warning.

Source — the only difference from the working control is the alias hop:

```tsx
"use client";
import { createSignal } from "@barefootjs/client";
export function Aliased() {
  const [items] = createSignal([{ id: 1, label: "a" }, { id: 2, label: "b" }]);
  const items__alias = items;
  return <ul>{items__alias().map(it => <li key={it.id}>{it.label}</li>)}</ul>;
}
```

## Measured

Compiled at `e38d6c476` through `compileJSX` + `HonoAdapter`. Emitted `Aliased.client.js`:

```js
hydrate('Aliased', { init: initAliased, template: (_p) =>
  `<ul bf="s1"><!--bf-loop:l0-->${(undefined)().map((it) => `<li data-key="${escapeAttr(it.id)}">…</li>`).join('')}…` })
```

`errors` is `[]`.

The control, identical but iterating `items()` directly, substitutes correctly:

```js
`…${([{ id: 1, label: "a" }, { id: 2, label: "b" }]).map((it) => …)}…`
```

So the substitution machinery knows how to inline the signal's initializer; it just resolves the alias to `undefined` instead of following the hop.

`template:` is the pure-CSR-mount path, so this is not a latent inconsistency — it is a guaranteed `TypeError: undefined is not a function` the first time the component is mounted without SSR markup. SSR and hydration are unaffected.

## Why it escapes the existing guards

The scope gate added in #2468 catches an emitted reference to a name that is not in scope. `undefined` **is** in scope — it is a global — so a wrong substitution to `undefined` passes a gate designed to catch a missing substitution. That is the difference from #2751, which leaves the reference unsubstituted; this one substitutes, and substitutes wrongly.

## Family

This belongs with #2741 and #2737: **expression rewriting performed as string surgery, with one private copy per emit path, each copy holding a different hole.**

- #2741 — `rewriteDestructuredPropsInExpr` rewrites object-literal *keys*, producing syntactically invalid client JS (the compiler's own `pruneUnusedPropExtractions: generated code did not parse` fires).
- #2737 — `transformExpr`'s `\b${propsObjectName}\.` misses a `props__alias` hop and emits a bare `props`, out of scope in that closure.
- This issue — the same alias-hop blindness, on the loop-source position, resolving to `undefined` rather than leaving the name.

The canonical device already exists and is AST-based: `csrSubstitute`, whose own comment states it uses *"structural position checks instead of regex lookbehind … emit does no string transformation of its own (#1277)."* Every hole above is a hole in exactly one of the regex copies that coexist with it.

This is the third time this repo has grown the same shape: regex import parsing (now banned by CLAUDE.md in favour of a TS AST walk) and ad-hoc loop-scope tracking (six copies, collapsed onto `BindingScope` by #2482). The fix shape that worked both times was one device threaded everywhere plus a shrink-only ratchet enumerating the legacy call sites.

## Owed

Three-piece set per the repo convention: this issue, a conformance fixture asserting the correct substitution with the **aliased** source, and a csr-conformance pin carrying this URL until the emission is fixed. Only the issue exists so far.

## Not established

The runtime `TypeError` is inferred from the emitted source rather than observed in a browser — `(undefined)()` has one possible outcome, but no CSR-mount leg was executed for this repro. The alias-resolution site was not isolated to a specific function; #2737's `transformExpr` and #2724's prop-alias hop are the neighbours to check first.

Found while auditing the root cause of the #2753/#2762 defect family, not by the pairwise sweep — the corpus has no aliased-loop-source axis, which is why it has gone unreported.

kfly8:
## Summary

Closes #2778. Files #2813 as a follow-on. Stacked on #2812 (#2806) → #2811 (#2737) — third PR in the "alias resolution" sequence: #2737 fixed the props-object identifier, #2806 fixed the rest-bag identifier, this one fixes the signal/memo GETTER identifier — three different names feeding the same rewrite/substitution machinery, each previously blind to an alias hop.

A local `const` alias of a signal or memo getter (`const items__alias = items`) was substituted as the literal `undefined` in the CSR template instead of following the alias to the getter's initializer/computation:

```js
// before
`<ul>${(undefined)().map((it) => ...)}</ul>`
// after
`<ul>${([{ id: 1, label: "a" }, { id: 2, label: "b" }]).map((it) => ...)}</ul>`
```

`(undefined)().map(...)` is a guaranteed `TypeError` the moment a component using this shape pure-CSR-mounts. SSR and hydration were unaffected — only the module-scope `template:` lambda takes this path. No BF diagnostic fires today because `undefined` is a real in-scope global, so the #2468 scope gate (built to catch a MISSING substitution) can't see a WRONG one.

### Root cause & fix

The alias binding is not a value copy — `items__alias` IS `items` for CSR-substitution purposes. `resolveGetterAliases` (new, `csr-substitute.ts`) registers the alias as the **same call-kind substitution entry** its origin already has in `buildSignalMemoEnv`, so `items__alias()` is substituted by the exact mechanism that already substitutes `items()` correctly — through every hop of a multi-hop chain (`const a = items; const b = a`).

This reuses a generalized `resolveAliasOrigin` hop-walker extracted from `props-binding.ts`'s existing rest/props-alias resolver (`resolveRestSpreadOriginCore` is now a 3-line wrapper over it), rather than growing a third alias-chain walker — the exact "one decision, two implementations" anti-pattern this issue named #2741 and #2737 as prior instances of.

`compute-inlinability.ts`'s `classifyConstantInitial` and `populateCsrInlinable` both consult the same `getterAliases` set (computed once in `computeInlinability`), so a two-hop alias no longer trips a spurious BF061 diagnostic that a one-hop alias (caught by the pre-existing free-identifier check) didn't.

### A separate gap this surfaced: #2813

Investigating the fix, I measured the same `aliased-loop-source` shape against every adapter, not just Hono. It fails on **all eight** non-Hono adapters:
- **ERB**: `undefined method 'length' for nil`
- **go-template**: `can't evaluate field Items__alias in type main.AliasedLoopSourceProps`
- **Mojolicious**: dies inside `Mojo::Template::process`
- **Blade, Jinja, MiniJinja (Rust), Twig, Xslate**: SSR an empty `<ul>`

This is the SSR-side twin of #2778 — every non-Hono adapter resolves a loop's array-source identifier by name lookup against seeded/struct data keyed by the signal's REAL name, with no alias-hop resolution at all. It's a genuinely separate fix surface (SSR data-seeding / Go field-routing, not CSR template substitution) so it's filed as #2813 rather than folded in here. Each affected adapter gets a `render-divergences.ts` pin naming #2813 until it's fixed — the intended fix should reuse `resolveAliasOrigin`/`resolveGetterAliases`, not a fourth alias walker.

## Test plan

- [x] New compiler unit test: `packages/jsx/src/__tests__/issue-2778-getter-alias-csr-template.test.ts` (one-hop signal alias, one-hop memo alias, two-hop chain with no spurious diagnostic, shadow-guard for a loop param literally named after the alias).
- [x] New conformance fixture: `packages/adapter-tests/fixtures/aliased-loop-source.ts` — correct on Hono; `renderDivergences` pins added on `adapter-erb`, `adapter-go-template`, `adapter-blade`, `adapter-jinja`, `adapter-mojolicious`, `adapter-rust`, `adapter-twig`, `adapter-xslate` (all verified to actually diverge, each pin's failure text taken from the real error).
- [x] Full `packages/jsx/src/__tests__` suite: 3179 pass, 0 fail.
- [x] Full `packages/adapter-tests/src/__tests__` suite: 2309 pass, 0 fail.
- [x] `binding-scope-ratchet.test.ts`, `coverage-map.test.ts`, `csr-skip-rot.test.ts`: green.
- [x] `packages/compat` suite (support-matrix + compat-lock freshness): green after regenerating `ui/support-matrix.lock.json`.
- [x] `generate-expected-html.ts` regenerated — zero diff for the new fixture's `expectedHtml` beyond the fixture's own addition (Hono is correct, so no `SKIP_AUTO_UPDATE` needed).
- [x] Changeset added (`@barefootjs/jsx` patch).

Investigation and design for this fix were done by a separate Fable-model agent per this stack's process — it verified the root cause against the actual current codebase (reproduced the issue's repro on HEAD, traced the exact `populateCsrInlinable`/`isInlinableInTemplate` code path that freezes in the `undefined` fallback, and identified `resolveRestSpreadOriginCore` as the precedent to generalize) before any code was written.

---


kfly8:
Addressed the pullfrog nitpick — and while fixing it properly, found it was masking a real bug.

Rebuilt the "shadow guard" test with a genuine collision: a component-scope getter alias (`const items__alias = items`) alongside an unrelated `.map()` row parameter of the same name. That surfaced an actual defect: `csrSubstitute`'s post-substitution `freeIdentifiers` were computed by re-parsing the rewritten text in isolation, with no visibility into `enclosingScope` (the loop-row bindings introduced outside the expression's own AST). The substitution step correctly left the row-shadowed identifier alone, but the free-identifier check still reported it as free and tripped the outer alias's unsafe-name gate — silently baking `key={items__alias.id}` to `data-key="undefined"` for every row.

Fixed in `csr-substitute.ts`: `freeIdentifiers` now filters out `enclosingScope`-bound names too, so both checks agree on what's shadowed. Pushed `41ee0d1d9`.

Verified: full `packages/jsx` suite (3179 pass, 0 fail) and `packages/adapter-tests` suite (2309 pass, 0 fail), both matching the pre-fix baseline — no regressions.

---

PR #2816, opened 2026-09-02
Fix #2794 + #2815: Go signal seeded from a module-level const bakes to nil

### LINKED ISSUE #2794: Go Template: signal seeded from a module-level const bakes to nil instead of its literal value
## Summary

`convertInitialValue` (`packages/adapter-go-template/src/adapter/value/value-lowering.ts`) bakes a signal's initial value into the generated `New<Component>Props` Go constructor. It resolves two shapes: a direct prop reference (`createSignal(props.x)`) and a literal expression (`createSignal('foo')`, `createSignal([1])`). A signal seeded from a bare identifier that refers to a **module-level `const`** (`const PAYLOAD = '...'; createSignal(PAYLOAD)`) matches neither — the analyzer types the signal `{ kind: 'unknown' }` in this case, and the function falls through every typed branch to its final `return 'nil'`.

The practical effect: on real `go run` SSR, the field is emitted as `Value: nil` in the constructor, so any content rendered from that signal (e.g. a controlled `<textarea>`'s value, lowered into element content by `lowerFormControlValueSsr`) comes back empty instead of carrying the module const's value. Every other adapter (Hono is the reference) renders this correctly.

`resolveModuleStringConst` already exists on the Go template adapter (`go-template-adapter.ts`) and is used to resolve exactly this kind of module-const reference at other call sites (e.g. `template-interp.ts`), but it is not wired into `convertInitialValue`'s signal-baking path.

## Reproduction

```tsx

const PAYLOAD = 'hello'

export function Demo() {
  const [value] = createSignal(PAYLOAD)
  return <textarea value={value()} />
}
```

Compiling this with the Go Template adapter and rendering the generated `New DemoProps` constructor emits:

```go
Value: nil,
```

instead of

```go
Value: "hello",
```

Confirmed via a direct `compileJSX` + `adapter.generateTypes()` inspection and via a real `go run` render (see PR #2792, where a fixture using this exact shape initially failed the Go Template adapter's own `JSX Conformance Tests` suite until the fixture was reverted to avoid the gap and this issue was filed instead).

## Fix direction

Wire `ctx.resolveModuleStringConst` (and, if a numeric-const analogue exists, its counterpart) into `convertInitialValue`'s bare-identifier resolution, alongside the existing `propsParams` lookup — mirroring how `template-interp.ts` already resolves the same identifier shape for live template expressions.

## Scope note

`packages/adapter-tests` fixtures are adapter-agnostic by design — they assert what SHOULD happen, not what one backend's current limitation allows. This gap should be closed by fixing the Go Template adapter, not by avoiding the module-const pattern in shared fixtures. Until fixed, the `textarea-row-breakout` fixture is pinned via `packages/adapter-go-template/src/render-divergences.ts`.

### LINKED ISSUE #2815: Go Template: signal seeded from a module-level boolean const bakes to nil instead of its literal value
## Summary

Same family as #2794 (fixed for string and numeric module consts): a signal seeded from a bare identifier referencing a module-level **boolean** const (`const OPEN = true; createSignal(OPEN)`) still bakes to `nil` in the generated `New<Component>Props` constructor.

#2794's fix wired `ctx.resolveModuleStringConst` and `ctx.resolveModuleNumericConst` into `convertInitialValue`'s bare-identifier branch (`packages/adapter-go-template/src/adapter/value/value-lowering.ts`). There is no `resolveModuleBooleanConst` (or equivalent) anywhere in the Go Template adapter — this is a genuinely new gap, not an existing resolver that was simply left unwired.

## Reproduction

```tsx

const OPEN = true

export function Demo() {
  const [open] = createSignal(OPEN)
  return <div>{open() ? 'y' : 'n'}</div>
}
```

Go bakes `Open: nil`, and the ternary reads the wrong branch (`n`) where Hono (and every other adapter) renders `y`.

## Fix direction

Either:
- add a `resolveModuleBooleanConst` mirroring `resolveModuleNumericConst`'s shape (`packages/adapter-go-template/src/adapter/go-template-adapter.ts`, exposed on `GoEmitContext` the same way #2794 did for the other two), or
- generalize the two existing resolvers plus this one into a single structural resolver keyed off `ConstantInfo.parsed` (a `literal` node with `literalType: 'string' | 'number' | 'boolean'`) rather than growing a third near-identical regex/lookup function — CLAUDE.md's "one decision, two implementations" rule already applies with two; a third copy-pasted resolver would be the third instance of the same defect family this repo keeps producing.

Wire the result into `convertInitialValue`'s bare-identifier branch alongside the string/numeric checks #2794 added.

## Scope note

Per the same convention as #2794: fixtures are adapter-agnostic and assert correct behavior, not this backend's current limitation. Pin `packages/adapter-go-template/src/render-divergences.ts` on whatever fixture exercises this shape until fixed.

kfly8:
## Summary

Closes #2794. Closes #2815. Stacked on #2814 → #2812 → #2811 — first PR in the Cluster B (Go adapter) sequence.

A signal seeded from a bare identifier referencing a module-level const (`const PAYLOAD = 'hello'; createSignal(PAYLOAD)`) baked to `nil` in the generated `New<Component>Props` constructor instead of the const's literal value:

```go
// before
Value: nil,
// after
Value: "hello",
```

### Root cause

The analyzer types this signal `unknown` — it never chases a bare identifier to its declaration — so none of `convertInitialValue`'s typed branches (`value-lowering.ts`) ever saw it, and it fell through to the final `nil`. `resolveModuleStringConst` already existed on the adapter for exactly this resolution (used by `template-interp.ts` for live template expressions) but wasn't wired into the signal-baking path.

### Fix

`convertInitialValue`'s bare-identifier branch now checks `ctx.resolveModuleStringConst`, `ctx.resolveModuleNumericConst`, and `ctx.resolveModuleBooleanConst` — in that order, **after** the destructured-prop lookup so a same-named prop still shadows the const.

### #2815, filed and fixed in the same commit

Investigating this surfaced the identical resolver-not-wired gap on two more literal kinds:
- **Numeric**: `resolveModuleNumericConst` already existed but wasn't exposed on `GoEmitContext` — two lines to wire up.
- **Boolean**: no resolver existed at all (`resolveModuleBooleanConst` is new).

Rather than leave the boolean case as a follow-up, it's fixed here since it's the exact same code path and shape. The `.find()` lookup all three resolvers need is shared through one `findModuleConst` helper instead of three separate ones — `binding-scope-ratchet.test.ts`'s shrink-only floor for `go-template-adapter.ts` (currently 5 `localConstants.find(` sites) stays flat rather than growing to 6 or 7.

### Pin graduation

Deletes the `textarea-row-breakout` entry in `packages/adapter-go-template/src/render-divergences.ts` — the fixture already exercises this exact shape and now renders correctly on real Go.

## Test plan

- [x] New compiler unit tests in `go-template-adapter.test.ts`: string+numeric module-const seeding, boolean module-const seeding, and a destructured-prop-shadows-a-same-named-const regression pin.
- [x] `textarea-row-breakout` conformance case: was 4 pass / 1 skip (pinned), now 6 pass / 0 skip on real `go run` (`GOTOOLCHAIN=go1.25.6`).
- [x] Full `packages/adapter-go-template/src/__tests__` suite (real `go run` per fixture): 1870 pass, 0 fail (confirmed clean on a second run after an unrelated transient disk-space issue in this sandbox was resolved — `go clean -cache` freed 26GB from an accumulated build cache).
- [x] `binding-scope-ratchet.test.ts`: green — the shared `findModuleConst` helper keeps the file's `.find(` count at its existing floor of 5.
- [x] `packages/compat` suite (compat-lock + support-matrix freshness): green after rebuilding all 8 dist-resolved adapter packages and regenerating both lockfiles.
- [x] `generate-expected-html.ts` regenerated — zero diff (Hono reference unaffected; this is a Go-only fix).
- [x] Changeset added (`@barefootjs/go-template` patch).

Investigation and design for this fix were done by a separate Fable-model agent per this stack's process — it verified the root cause against the actual current codebase (reproduced the issue's repro via `generateTypes` + a real `go run` render, traced the analyzer's `unknown`-typing path to its exact line, and confirmed `resolveModuleNumericConst` already existed in the file but wasn't on the `GoEmitContext` seam) before any code was written.

---


kfly8:
Addressed the pullfrog nitpick — verified it against the actual compiler output first (the destructured-prop-shadows-const case emits `Value: in.PAYLOAD,`, matching the `Value: in.Value,` convention already used elsewhere in this file). Added that as a positive assertion alongside the existing negative one, so the test actually pins the shadowing behavior rather than just the bug's absence. Pushed `7faa5a577`.

Verified: full `go-template-adapter.test.ts` suite, 1813 pass, 0 fail.

---

PR #2819, opened 2026-09-02
Fix #2741: reactive-effect prop rewrite no longer touches object-literal keys

### LINKED ISSUE #2741: queryHref effect-update emit prop-rewrites object-literal keys, producing syntactically invalid client JS

kfly8:
## Summary

- `rewriteDestructuredPropsInExpr` (`emit-reactive.ts`) — the reactive `createEffect` update path's destructured-prop rewrite — ran a bare `\b\b` regex over the whole expression text with no structural distinction between a value reference and an object-literal key. So `queryHref(base, { tag: tag, page: page ? page : '' })` on a component destructuring `{ base, tag, page }` had the KEY rewritten too: `{ _p.tag: _p.tag, ... }` — not valid JS, so the whole client module failed to parse. No diagnostic was raised; the invalid module was silently written out (only surfacing as a `[barefootjs] pruneUnusedPropExtractions: generated code did not parse; skipping prune` warning with no indication of the actual cause).
- The hydrate template lambda in the same output was always correct — it already goes through `rewriteBarePropRefs` (`prop-rewrite.ts`), the one AST walk in the compiler that distinguishes a value reference from an object-literal key, a property-access name, a shorthand property, or a name shadowed by an inner binding (`items.map((tag) => tag.a)`). Per CLAUDE.md's "one decision, two implementations, no test comparing them" rule, `rewriteDestructuredPropsInExpr` now **delegates to that same function** instead of carrying a third, regex-based copy of the same decision.
- `applyScopedPropRefRewrite` and its unparseable-text fallback `applyRegexPropRefRewrite` (both in `prop-rewrite.ts`) gained an optional `replacementFor` override so the effect path's `(_p.x ?? <default>)` destructure-default wrapping stays call-site-specific — the walk itself, and its key/value distinction, is not duplicated.
- Graduates the `query-href` CSR-conformance skip entry (`csr-skip-set.ts`).
- Along the way, running the actual CSR-conformance test surfaced a **second, previously-masked** gap: the CSR test harness (`csr-render.ts`) strips real ES module imports from generated code and provides its own runtime-helper mirrors (`escapeAttr`, `date`, `formatDate`, `spreadAttrs`, …) — but had none for `queryHref`, so once the syntax error was fixed the fixture failed with `ReferenceError: queryHref is not defined` instead. Added a `queryHref` mirror matching the real implementation, in the same style as its neighbors.
- Deliberately does **not** touch a separate, pre-existing gap noted during investigation: a prop referenced only via object-literal **shorthand** (`{ page }`) is not currently discovered as a value reference by `collectAstPropRefs` at all (a Phase-1 exclusion, unaffected by this change) — filed as #2828 (pullfrog review, PR #2819), a distinct bug with its own fix and fixture needed.

## Test plan

- [x] New unit tests in `rewrite-destructured-props.test.ts`: explicit-key object literal keeps its keys and rewrites only values; a keyed prop with a destructure default still wraps only in value position; a prop name shadowed by an inner callback parameter is not rewritten inside that callback.
- [x] Full `packages/jsx` test suite: 3182 pass, 0 fail (two pre-existing, unrelated `pruneUnusedPropExtractions` warnings confirmed present on `main` too, via `git stash`).
- [x] `query-href` fixture: `csr-conformance.test.ts` now passes (was silently emitting invalid JS before, masked by the skip entry).
- [x] Full CSR conformance suite (`csr-conformance.test.ts`): 336 pass, 0 fail.
- [x] `csr-skip-rot.test.ts`: 23 pass, 0 fail (skip-set entry removal is consistent with the harness).
- [x] `generate-expected-html.ts` run — zero diff on tracked fixtures (this is a client-JS-only fix; SSR output is unaffected on every adapter).
- [x] `bun test packages/compat` → 521 pass, 1 timeout (re-ran standalone with a longer timeout: 8 pass, 0 fail — confirmed transient, not a regression).

Stacked on #2818 (`claude/fix-2700-go-signal-object-literal-loud-refusal`), which stacks on #2817 → #2816 → #2814 → #2812 → #2811, per the ongoing bug-fix marathon in piconic-ai/barefootjs.

Closes #2741.

---

### DIFF
--- a/.changeset/reactive-effect-prop-rewrite-object-keys-2741.md
+++ b/.changeset/reactive-effect-prop-rewrite-object-keys-2741.md
@@ -1,0 +1,9 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix #2741: `rewriteDestructuredPropsInExpr` (the reactive `createEffect` update path's destructured-prop rewrite) ran a bare `\b<prop>\b` regex over the whole expression text, so an object literal whose property KEY happened to share a destructured prop's name (`queryHref(base, { tag: tag, page: page ? page : '' })`) had the KEY rewritten too (`{ _p.tag: _p.tag, ... }`) — not valid JS, so the whole client module failed to parse. No diagnostic was raised; the invalid module was written out silently.
+
+`rewriteDestructuredPropsInExpr` now delegates to `rewriteBarePropRefs` (`prop-rewrite.ts`) — the same AST walk the hydrate template lambda already used, which distinguishes a value reference from an object-literal key, a property-access name, a shorthand property, or a name shadowed by an inner binding (`items.map((tag) => tag.a)`). `applyScopedPropRefRewrite` and `applyRegexPropRefRewrite` (its unparseable-text fallback) gained an optional `replacementFor` override so the effect path's `(_p.x ?? <default>)` destructure-default wrapping stays call-site-specific without a second copy of the walk.
+
+Graduates the `query-href` CSR-conformance skip entry (`packages/adapter-tests/src/csr-skip-set.ts`); the CSR test harness also gained a `queryHref` runtime mirror (`csr-render.ts`), a second, previously-masked gap the syntax error had hidden (the harness strips real imports and stubs each runtime helper a fixture's generated code calls, and `queryHref` had no stub yet).
--- a/packages/adapter-tests/src/csr-render.ts
+++ b/packages/adapter-tests/src/csr-render.ts
@@ -491,6 +491,24 @@ function formatDate(dateArg, pattern, timeZone = 'UTC', names = []) {
       : nameAt(31 + weekday),
   )
 }
+// Mirror @barefootjs/client's queryHref (#2741 graduation): a pure
+// base-path + query-params URL builder. Same stripped-imports reasoning
+// as \`date\`/\`formatDate\` above — the user's own
+// \`import { queryHref } from '@barefootjs/client'\` is stripped with
+// every other import, so a call in template/effect code needs this
+// mirror or the fixture fails with "queryHref is not defined".
+function queryHref(base, params) {
+  const u = new URLSearchParams()
+  for (const [key, value] of Object.entries(params)) {
+    if (Array.isArray(value)) {
+      for (const member of value) if (member) u.append(key, member)
+    } else if (value) {
+      u.set(key, value)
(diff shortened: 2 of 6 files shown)


kfly8:
Filed #2828 for the shorthand-prop-reference gap pullfrog flagged (previously only documented as PR-body prose, per CLAUDE.md's "a reproducible defect lands as a fixture, not a prose report" convention) and linked it from this PR's description. The fix itself and its own fixture are tracked there rather than folded into this PR.

---

PR #2825, opened 2026-09-03
Fix #2771: refuse loudly when a reactive primitive is called through an unresolved namespace import

kfly8:
## Summary

- `import * as bf from '@barefootjs/client'` followed by `bf.createSignal(0)` (or any other reactive primitive — `createMemo`/`createEffect`/`onMount`/`onCleanup`/`createSearchParams` — accessed off the namespace binding) compiled with zero diagnostics. `resolvePrimitiveKind`'s fast path only matches a bare identifier callee; its checker-based slow path, which DOES resolve a namespace-qualified call, only runs when a shared `ts.Program` is supplied via `CompileOptions.program` (which `@barefootjs/vite` always provides, but a bare `compileJSX` call — the adapter-conformance runner, `packages/compat`'s engine — does not). Without one, the declaration was silently dropped from `ctx.signals`/etc. while every reference to it survived in the emitted output, throwing `ReferenceError` at hydrate.
- New `BF013` diagnostic (`validateNamespaceQualifiedPrimitives`, `analyzer.ts`, wired into `validateContext` right after the existing BF110 pass) refuses loudly instead — gated on **non-recognition**, not on the namespace-import shape itself, so a compile that DOES supply a program is completely untouched (pinned by a dedicated checker-path test using the same `programFor` harness as `primitive-resolver-alias.test.ts`).
- False-positive guards: a namespace import of a different module, a component-local binding that shadows the namespace name, and a type-only namespace import are all confirmed unaffected.
- Suggests importing the primitive by name as the fix (`suggestion.escape: [{ kind: 'rewrite' }]` — a named import genuinely resolves this, unlike a `/* @client */` escape which wouldn't help here).
- Adds the `namespace-import-primitive` fixture (refused ahead of any adapter's `generate()`, so pinned identically across all nine adapters — the same reasoning as `jsx-element-prop-ternary`'s BF021) and its named-import escape twin `namespace-import-primitive-named-escape`.
- Docs: new `### BF013` section + quick-reference row in `docs/core/advanced/error-codes.md` (renamed the surrounding `## Signal Errors` heading to `BF011–BF013`), and a table row in `spec/compiler.md`.

## Test plan

[...]

PR #2824, opened 2026-09-03
Reclassify #2719 to bare known-limitation (ledger only, no code fix)

kfly8:
## Summary

- Re-measured #2719's `todo-app` oracle divergence against the current stack head (none of the intervening fixes — #2737/#2778/#2794/#2800/#2700/#2741/#2702 — touch `TodoApp.tsx`'s SSR markup or client JS, so the result is unaffected by exactly which commit is used).
- The "large structural divergence" is fully narrowed: `TodoApp.tsx` marks four regions `/* @client */` (the filtered keyed loop, both todo-count text expressions, the clear-completed conditional). Per `spec/callback-fidelity.md` §4, SSR renders those as empty placeholders BY CONTRACT and the client fills them in — that is exactly what `snap`/`three-point` are flagging, and it isn't a divergence from the compiler's contract.
- Verified: masking the four regions out of the hydrated tree makes it byte-identical to SSR; the three-point oracle's hydrated-vs-csr-mount leg agrees; `idempotence` passes; the marker-free twin `todo-app-ssr` passes all three oracles cleanly. Not #2714 (no SSR-present attribute is corrected) and not #2715 (no named-prop mirror) — different mechanism from both.
- This makes #2719 a gap in the ORACLE's expressiveness (no way to declare "this region is an intentional `@client` placeholder"), not a bug in the compiler or in `TodoApp.tsx` — per `CLAUDE.md`'s compatibility-policy tiers, that's an accepted permanent design position (bare `known-limitation`), not `+ bug`.
- Relabeled the GitHub issue (`bug` removed, `known-limitation` kept), retitled it to describe the narrowed mechanism, and updated `packages/adapter-tests/e2e/oracle-quarantine.ts`'s `todo-app` reason string (and added an explanatory comment) to match — no compiler/component/runtime code changed.

## Test plan

- [x] No code fix — this is a doc/ledger-only change. `oracle-quarantine.ts` loads and parses correctly (`bun -e "import(...)"` sanity check).
- [x] The empirical re-measurement (masked-tree diff, hydrated-vs-csr-mount comparison, `todo-app-ssr` control) was run directly against the actual Playwright oracle harness in this environment; full details in the linked issue comment.

Stacked on #2823 (`claude/fix-2777-aliased-import-child-registry`), which stacks on #2821 → #2820 → #2819 → #2818 → #2817 → #2816 → #2814 → #2812 → #2811, per the ongoing bug-fix marathon in piconic-ai/barefootjs.

[...]

PR #2826, opened 2026-09-03
Fix #2728: register a comment scope for a bare-mounted comment-wrapper

### LINKED ISSUE #2728: tabs: csr-mount doesn't compute the default-active TabsTrigger the same way as hydration
## Problem

Unmasked while fixing #2716 (a live `.value` DOM property expando that made `oracle.playwright.ts`'s `[three-point]` oracle fail its FIRST comparison — SSR vs hydrated — before ever reaching the second).

`runThreePointOracle` (`packages/adapter-tests/e2e/oracle-core.ts`) checks two things in order:
1. SSR vs hydrated (same as the `snap` oracle)
2. hydrated vs csr-mount

With #2716 fixed, (1) now passes for the `tabs` fixture — which lets (2) run for the first time and it fails:

- **hydrated**: the default-active `TabsTrigger` (Account) renders `aria-selected="true" data-state="active" data-value="account"`.
- **csr-mount**: the SAME trigger renders `aria-selected="false" data-state="inactive"` with `data-value` missing entirely.

So a from-scratch client-side mount of `Tabs` (no SSR input) does not compute/apply the default active tab the same way hydration does — a real CSR-vs-hydration behavioral divergence, not a DOM-property/attribute-hygiene issue like #2716.

## Repro / pins

Quarantined in `packages/adapter-tests/e2e/oracle-quarantine.ts` under `tabs` (`oracles: ['three-point', 'idempotence']`, rot-checked — see that file's docstring). `tabs`' `idempotence` failure (a distinct click-replay timeout) is tracked in the same ledger row; may or may not share a root cause with this one.

## Graduation

Find why the csr-mount code path doesn't select/apply the default `TabsTrigger` state the same way the hydrate path does (likely in the generated `initChild`/`createComponent` bootstrap for `Tabs`/`TabsTrigger`, or in how `activeTab`'s initial value reaches the csr-mount tree vs the hydrated one). Confirm the `three-point` fixture passes, delete the `three-point` entry from the ledger row (leave `idempotence` if still separately failing).

kfly8:
## Summary

- A "root is a child call" comment-wrapper component (`comment: true`, no `fragmentRoot` — the shape a `'use client'` component compiles to when its entire render is a single child-component call, e.g. `Tabs`' generated demo entry point wrapping `<Tabs><TabsList>...</TabsList>...</Tabs>`) mounted bare at the top level via `createComponent` (the CSR-mount path, no SSR markup to hydrate from) never registered a `<!--bf-scope:-->` boundary-comment pair for itself — unlike hydration's `hydrateCommentScope`, which registers one from the SSR-rendered comments.
- `$c()`'s dual-scope lookup (`getDualScopeIds`) then found nothing for any of the wrapper's own sibling child slots baked directly into its template (`TabsTrigger` ×2, `TabsContent` ×2 in the `Tabs` case) — they were never `initChild`'d at all, so their props and event handlers never ran even though their markup was present. This is why `tabs`' default-active trigger showed the wrong `aria-selected`/`data-state`/`data-value` on a from-scratch CSR mount, and why click replay (`idempotence`) diverged — one shared root cause behind both previously-quarantined failures, not two.
- `materializeComponent` (`packages/client/src/runtime/component.ts`) now emits the same boundary-comment pair for this wrapper shape that a genuine fragment root already gets (#2722), derived from the SAME `wrapperScopeId` #2757 already computes purely for prop-threading. The comment gate widens from `isFragmentRoot ? scopeId : null` to `isFragmentRoot ? scopeId : (isCommentWrapper ? wrapperScopeId : null)`; nothing else in the mount/connect/init pipeline needed to change, since everything downstream already keys off the generic `fragmentComments` value rather than `isFragmentRoot` directly.
- A bare top-level mount of this wrapper shape now returns a `DocumentFragment` instead of a bare `HTMLElement` (matching the existing fragment-root contract, and for the same reason: the boundary comments have no single element to travel attached to). Six existing tests assumed the old bare-element return — updated to append the result first, then re-acquire the element from the document (or query the fragment directly when not appending), per the same pattern `renderchild-noninlinable-props.test.ts` already used elsewhere in this file.
- Deleted the `tabs` quarantine row entirely (`packages/adapter-tests/e2e/oracle-quarantine.ts`) — both its `three-point` and `idempotence` failures share this one root cause and both are now fixed.

## Test plan

- [x] New `packages/client/__tests__/runtime/issue-2728-comment-wrapper-csr-mount-child-slots.test.ts`: a bare top-level mount of this wrapper shape returns a `DocumentFragment` carrying its own registered comment scope; every sibling child slot is actually initialized (not just present in markup) and stays reactive across a click after a bare CSR mount.
- [x] Full `packages/client` test suite: 671 pass, 0 fail (six pre-existing tests updated for the return-shape change, verified they still assert the same underlying behavior).
- [x] `packages/adapter-tests` CSR-conformance suite: 338 pass, 0 fail.
- [x] Real Chromium oracle run (`bun run --filter '@barefootjs/client' build` then `playwright test e2e/oracle.playwright.ts -g tabs`): `[snap]`, `[three-point]`, and `[idempotence]` for `tabs` all pass — confirmed both by removing the quarantine entry first (harness's own rot-check fired: "entry ... is stale — the fixture now passes") and by a clean re-run afterward.
- [x] Full oracle suite (`e2e/oracle.playwright.ts`, all fixtures): no regressions from this change. Incidentally surfaced two PRE-EXISTING, unrelated stale-quarantine entries (`accordion` and `command`, both `idempotence`, tracked under #2714) — not touched here; noted separately, out of scope for this PR.

Stacked on #2825 (`claude/fix-2771-namespace-import-primitive-loud-refusal`), which stacks on #2824 → #2823 → #2821 → #2820 → #2819 → #2818 → #2817 → #2816 → #2814 → #2812 → #2811, per the ongoing bug-fix marathon in piconic-ai/barefootjs.

Closes #2728.

---

### DIFF
--- a/.changeset/comment-wrapper-csr-mount-scope-2728.md
+++ b/.changeset/comment-wrapper-csr-mount-scope-2728.md
@@ -1,0 +1,9 @@
+---
+"@barefootjs/client": patch
+---
+
+Fix #2728: a "root is a child call" comment-wrapper component (`comment: true`, no `fragmentRoot` — the shape a `'use client'` component compiles to when its entire render is a single child-component call, e.g. a demo/story component wrapping `<Tabs>...</Tabs>`) mounted bare at the top level via `createComponent` (the CSR-mount path, no SSR markup to hydrate from) never registered a `<!--bf-scope:-->` boundary-comment pair for itself, unlike hydration's `hydrateCommentScope`, which does register one from the SSR-rendered comments. `$c()`'s dual-scope lookup then found nothing for any of the wrapper's own sibling child slots baked into its template — they were never `initChild`'d at all, so their props and event handlers never ran even though their markup was present (dead click handlers, un-applied template-only attributes).
+
+`materializeComponent` now emits the same boundary-comment pair for this wrapper shape that a genuine fragment root already gets (#2722), derived from the same id `#2757` already computed for thread-only purposes. A bare top-level mount of this shape now returns a `DocumentFragment` (matching the existing fragment-root contract) instead of a bare `HTMLElement` — six existing tests that assumed the old return shape are updated to append-then-requery rather than operate on the returned handle directly.
+
+Verified against the real oracle harness: `tabs`' `three-point` and `idempotence` oracles, previously quarantined (`packages/adapter-tests/e2e/oracle-quarantine.ts`) under this exact issue, both pass in real Chromium.
--- a/packages/adapter-tests/e2e/oracle-quarantine.ts
+++ b/packages/adapter-tests/e2e/oracle-quarantine.ts
@@ -41,13 +41,10 @@ export const ORACLE_QUARANTINE: Readonly<Record<string, QuarantineEntry>> = {
   // client effect pass CORRECTS it. The snap/three-point failures record
   // that pre-hydration state genuinely differs from post-hydration state;
   // the side that is wrong is the SSR markup, not hydration.
-  // `idempotence` graduated alongside #2728 (comment-wrapper CSR-mount
-  // scope registration): the second item's trigger is itself reached
-  // through a comment-wrapper composition, so its mount effects never ran
-  // on a from-scratch csr-mount before that fix — the [data-value] locator
-  // timeout was a symptom of the same missing-comment-scope bug, not a
-  // separate one. `snap`/`three-point` are unrelated (an SSR-literal-vs-
-  // hydration-correction divergence, see reason below) and stay quarantined.
+  // `idempotence` graduated alongside #2728 (same root cause: the second
+  // item's trigger sits inside the same comment-wrapper composition).
+  // `snap`/`three-point` are a separate, unrelated divergence (see reason
+  // below) and stay quarantined.
   accordion: {
     oracles: ['snap', 'three-point'],
     reason:
@@ -60,16 +57,10 @@ export const ORACLE_QUARANTINE: Readonly<Record<string, QuarantineEntry>> = {
       'Default-checked radio item SSRs the hard-coded aria-checked="false" literal; hydration corrects it to "true".',
     issue: 'https://github.com/piconic-ai/barefootjs/issues/2714',
   },
-  // `idempotence` moved to `IDEMPOTENCE_EXCLUDED` (#2827): unlike
-  // `accordion`'s idempotence row, `command`'s does NOT reliably graduate
-  // alongside #2728 — measured intermittent (2/5 and other repeat runs),
-  // with a genuine structural divergence (item selection / group-visibility
-  // state disagrees between the hydrated and csr-mount replay legs) some of
-  // the time and agreement the rest. That bimodality broke the quarantine
-  // ledger's "reliably fails" assumption in both directions on CI (the
-  // structural-divergence failure on one run, the stale-entry rot-check on
-  // another, same pair) — see `oracle.playwright.ts`'s `IDEMPOTENCE_EXCLUDED`
-  // for the skip and #2827 for the root-cause investigation.
+  // `idempotence` moved to `IDEMPOTENCE_EXCLUDED` (#2827) instead of
+  // graduating like `accordion`'s: measured bimodal (a genuine structural
+  // divergence some runs, agreement others), so the ledger's "reliably
+  // fails" assumption doesn't hold for this pair.
   command: {
     oracles: ['snap', 'three-point'],
     reason:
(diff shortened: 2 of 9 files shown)


kfly8:
`fixture-hydrate` CI failed on the previous commit: `oracle-quarantine.ts entry for [accordion]/'idempotence' is stale — the fixture now passes this oracle`. Root cause: this PR's fix (registering a comment scope for a bare comment-wrapper CSR mount) incidentally also fixes `accordion`'s and `command`'s idempotence divergence — the second accordion trigger and the command list items are themselves reached through the same comment-wrapper composition, so their mount effects never ran on a from-scratch csr-mount before this fix either.

Verified locally against the real oracle harness (Chromium): both fixtures' `idempotence` now consistently pass; `snap`/`three-point` remain genuinely failing for both (a separate, pre-existing SSR-literal-vs-hydration-correction divergence, unrelated to this PR) and stay quarantined. Removed the now-stale `idempotence` entries from both — pushed as `445e17cc5`. Full oracle suite (129 tests) is clean with 1 legitimately-still-quarantined pair skipped, 0 failures.

---

kfly8:
Correction to my previous comment: `command`'s `idempotence` graduation was wrong. Repeated local verification (5x each) shows `accordion` passes reliably (5/5) — that graduation stands — but `command` is genuinely intermittent (3/5 passed, 2/5 failed with a real structural divergence: item selection / group-visibility state disagrees between the hydrated and csr-mount replay legs, not a stale-rot-check false positive).

This flakiness pre-dates this branch: with `command`'s quarantine entry restored to its exact original (byte-identical to `main` before this PR), the rot-check *itself* still intermittently reports "stale" locally, because the underlying test occasionally passes by chance — the quarantine ledger's "must still fail" contract doesn't hold up against a genuinely flaky assertion. So this is a pre-existing hazard in the harness, unrelated to and not worsened by this PR's fix — `packages/adapter-tests/e2e/oracle.playwright.ts` already has a sanctioned pattern for exactly this shape (`IDEMPOTENCE_EXCLUDED`, currently used for `carousel`'s `drag` step), and `command` is a reasonable future candidate for it, but that's out of scope for a comment-wrapper CSR-mount fix.

Reverted `command`'s entry to its original content and pushed (`c7f529a36`). If CI still hits this specific flake on the current run, it's the same pre-existing intermittent condition, not a regression from this PR — I'll re-run rather than push further changes.

---

kfly8:
`fixture-hydrate` failed again on `c7f529a36`, but on the *other* face of the same flake: `command`'s `idempotence` test happened to pass this run, which trips the quarantine ledger's rot-check ("entry is stale — delete it"). Combined with the earlier run's genuine structural-divergence failure on the identical pair, this confirms the test is bimodal — not "reliably fails" (stay quarantined) nor "reliably passes" (graduate), so re-running alone can't converge; either outcome trips a different check.

Filed #2827 to track root-causing the actual race (a filtered-list reconciliation disagreement between the hydrated and csr-mount replay legs — unrelated to this PR's #2728 fix), and pushed `6b61faa8b`: moves `command` into `IDEMPOTENCE_EXCLUDED` (`oracle.playwright.ts`, same mechanism already used for `carousel`'s drag-step flake) and drops `'idempotence'` from its `oracle-quarantine.ts` entry (`snap`/`three-point` stay quarantined under #2714, unaffected).

Verified locally against the real Chromium harness: `--repeat-each=5` on `command` (idempotence reliably skips, snap/three-point still pass) and the full oracle suite (128 passed, 2 skipped: carousel drag + command idempotence, both intentional).

---

kfly8:
`test (jsx-2)` has now failed twice in a row on this exact head (`e5730400e`), both times on the identical assertion:

```
(fail) #2641 — reactive attribute bindings route catalogued calls through the runtime helper > non-@client reactive attribute, a catalogued call [~3ms]
```

Investigated before re-running a second time:
- This PR's diff touches only `packages/client/src/runtime/component.ts` and `packages/adapter-tests/e2e/oracle-quarantine.ts` — nothing in `packages/jsx` (the compiler package this test lives in), so there's no code-path connecting this PR's change to that test.
- Reproduced the exact CI shard locally — same file list and order as `bun scripts/ci/shard-test-files.ts packages/jsx 2 2` — with both bun 1.3.11 (local default) and bun 1.3.14 (CI's pinned version, downloaded directly from the GitHub release to match exactly): both runs passed 1797/1797 with 0 failures.
- The identical failure (same test, same ~3ms duration, same 1796/1797 pass count) also occurred on `#2824` and `#2825` in this same stack — both later passed cleanly on a re-run of the same job, at the same head commit, with no code change.

That rules out both this PR's diff and the bun version as the cause; it isn't reproducible outside the CI runner. Given `#2825`'s identical shard passed on retry while this one hasn't yet, I'm treating this as CI-runner-side noise (resource contention makes a load-sensitive path in the compiler's date-lowering caches behave differently under CI's parallelism than in an isolated local run) rather than re-running blindly a third time. Will keep watching and re-run again on the next push or check-in.

---

kfly8:
Update: this is now the third identical failure on this exact head (same test, same assertion, same ~3ms duration) — including one re-run I triggered after the comment above, which failed identically again. That's enough occurrences that it's not simple bad luck, but it's also not something I can pin on this PR's diff (`packages/client/src/runtime/component.ts` + the oracle ledger only — nothing in `packages/jsx`, where the failing test lives) or on the bun version (verified against CI's exact pinned 1.3.14 locally).

One structural note for whoever picks this up: the `test (jsx-2)` job's setup step also does `bun run --filter '@barefootjs/client' build` before running the jsx test shard (`.github/workflows/ci.yml`), so this PR's `component.ts` change is present in that shard's environment even though no `packages/jsx` test imports `@barefootjs/client` directly. I don't have a mechanism connecting that build to the specific failing assertion, but it's the one concrete difference between this PR's CI run and `#2825`'s (whose identical shard passed on retry).

Holding off on further re-runs per the one-retry guidance — already at three. Will pick this back up on the next natural trigger (a push, or if a maintainer re-runs it) rather than keep retrying blind.

---

Issue #2743, opened 2026-08-28
go-template: html/template's URL-context autoescape percent-encodes queryHref bases, diverging from the JS reference


kfly8:
## Symptom

For `<a href={queryHref(base, { … })}>`, the Go adapter's rendered `href` diverges from the JS reference when `base` contains non-URL-safe or multibyte characters:

| point | JS reference (Hono) | Go |
|---|---|---|
| `query-href:gen:base:markup` (`<b>&"'</b>`) | `&lt;b&gt;&amp;&quot;&#39;&lt;/b&gt;?tag=sale&amp;page=2` | `%3cb%3e&amp;%22%27%3c/b%3e?tag=sale&amp;page=2` |
| `query-href:gen:base:multibyte` (`日本語`) | `日本語?tag=sale&amp;page=2` | `%e6%97%a5%e6%9c%ac%e8%aa%9e?tag=sale&amp;page=2` |

## Mechanism

The runtime helper is not the source: `Query` (`packages/adapter-go-template/runtime/bf.go`) returns `base + "?k=v…"` with the base untouched, form-encoding only keys/values — matching the client `queryHref`. The percent-encoding comes from **`html/template`'s contextual autoescape**: a pipeline that lands in an `href` attribute passes through the URL-context escaper, which percent-encodes non-URL-safe bytes in the whole value. The JS reference only HTML-escapes. URL-structural characters (`?`, `=`, `&`) survive on both sides, which is why every previously existing fixture (URL-safe bases like `/items`) renders identically.

Not queryHref-specific in principle — any dynamic `href` interpolation on the Go adapter gets the same treatment; the `query-href` fixture's generated adversarial points (`gen:base:markup` / `gen:base:multibyte`, first landed in #2742) are simply the first to put such bytes into an `href`.

ERB and Jinja pass both points on their real backends, so the divergence is Go-side, not reference-side.

## Pins

`skipDataPoints` entries `query-href:gen:base:markup` and `query-href:gen:base:multibyte` in `packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts`, carrying this issue's URL.

## Resolution options

1. Neutralize the URL-context escaper for `bf_query` output (e.g. emit the helper call in a context html/template treats as opaque, or wrap as `template.URL`), restoring byte parity with the reference — note `template.URL` waives Go's sanitization, so the escape-analysis implications need a look.
[...]

PR #2841, opened 2026-09-05
Fix #2743: Go adapter's queryHref href no longer percent-encodes base

kfly8:
## Summary

`<a href={queryHref(base, {…})}>` on the Go adapter diverged from the JS reference (Hono) whenever `base` contained non-URL-safe or multibyte bytes, e.g. `日本語` rendered as `%e6%97%a5%e6%9c%ac%e8%aa%9e...` instead of passing through untouched. Root cause: `html/template` infers a URL context from the attribute name (`href`, `src`, `action`, `data-*`, …) and percent-encodes the whole pipeline result landing there; the reference only HTML-escapes. The `bf_query` runtime helper itself was already byte-faithful — the divergence was purely html/template's contextual auto-escaping bolted on top of the emitted attribute.

- Added a new `bf_attr(name, value) template.HTMLAttr` runtime helper (`packages/adapter-go-template/runtime/bf.go`) that emits a WHOLE `name="value"` attribute, HTML-escaping the value and nothing else — the same technique `bf_spread_attrs`/`bfHydrationAttrs` already use to bypass html/template's contextual auto-escaping (and `StyleObjectToCSS` uses for the analogous CSS-context case).
- Added `lowerRegisteredAttrCall` (`packages/adapter-go-template/src/adapter/expr/url-builder.ts`), which recognizes a `query` guard-list lowering node (the shape `queryHref` produces) landing in an attribute position — as a direct call OR inside a ternary branch (either side, recursing through right-folded chains) — and routes it through `bf_attr` instead of the generic `name="{{...}}"` emission. Wired into `elementAttrEmitter.emitExpression` in `go-template-adapter.ts`.
- The route is keyed on the neutral `helper === 'query'` fact, **not** the attribute name — so `src`, `title`, or any other attribute built from `queryHref` gets the same byte parity, not just `href` (Go's URL-context attribute table is an internal heuristic that shouldn't be re-implemented adapter-side).
- Confirmed against real Hono behavior: the reference adapter does no scheme filtering either (no `javascript:`/`data:` check) — it only HTML-escapes. Matching it exactly is the adapter contract, not a safety regression; a spread-props `href` already bypassed Go's scheme filter the same way before this change.

## Design

Reviewed and designed by an independent design pass (fable) before implementation: verified experimentally that `template.URL` does **not** fix the bug (it still percent-encodes; it only bypasses the scheme filter), and confirmed the whole-attribute `template.HTMLAttr` route matches Hono byte-for-byte (modulo the conformance harness's own HTML-entity-spelling normalization).

A follow-up round (pullfrog review) found the direct-call fix didn't cover a `queryHref` value reached through a ternary attribute branch with a real, non-`undefined` alternate (`cond ? queryHref(...) : '/fallback'`) — that shape already lowered both branches correctly via `bf_ternary`, but the whole pipeline was still wrapped in the plain `name="{{...}}"` form. `lowerRegisteredAttrCall` now also accepts a `conditional` node for this case.

## Test plan

- [x] `packages/adapter-go-template/runtime/bf_test.go`: `TestAttr` (unit) + `TestAttr_BypassesContextualURLEscaper` (end-to-end through a real `html/template` execution, asserting no percent-encoding / no `#ZgotmplZ`) + `TestScratchTernaryAttrBypass`-equivalent coverage for the ternary case
- [x] `packages/adapter-go-template/src/__tests__/query-href.test.ts`: shape pins — `href` and non-`href` (`title`) attributes route through `bf_attr`, text position unaffected, the ternary (non-undefined-alternate) shape routes through `bf_attr`, and the pre-existing `#2842` undefined-alternate gap is confirmed unaffected
- [x] `packages/adapter-go-template/src/__tests__/lowering-plugin.test.ts`: pin that a `helper-call` node (a different `LoweringNode` kind) is never routed through `bf_attr`
- [x] Removed the `query-href:gen:base:markup` / `query-href:gen:base:multibyte` `skipDataPoints` in `go-template-adapter.test.ts` — both now pass against the live Hono oracle
- [x] Added a declared `base-scheme-passthrough` data point to `packages/adapter-tests/fixtures/query-href.ts` pinning the no-scheme-filtering parity decision
- [x] Added `packages/adapter-tests/fixtures/query-href-src.ts` (queryHref in a `src` attribute, attribute-agnostic regression) and `query-href-ternary.ts` (queryHref in a ternary attribute branch, both branches), both registered in `fixtures/index.ts`
- [x] `query-href-ternary` compiles fine on Hono/Go but is refused (`BF101`) by all 7 other adapters (mojolicious/xslate/twig/erb/blade/jinja/minijinja) — a pre-existing, broader gap unrelated to this PR's escaping fix (their ternary-attribute lowering never consults the lowering registry for a branch). Filed as **#2843** and pinned per-adapter (with `unescapable` declarations, required by the escape-coverage floor test) so the shared conformance suite stays green.
- [x] Regenerated `expectedHtml`, `generated-data-points.json`, `coverage-map.json`, `ui/support-matrix.lock.json`, `ui/compat.lock.json` at each step — `git status` confirmed no unrelated drift each time
- [x] `packages/compat`'s full test suite (555 tests, including escape-coverage and both lock files' freshness tests) green
- [x] `bun test` green across all 9 adapters for `query-href`/`query-href-src`/`query-href-ternary`
- [x] `go test` green for the new `Attr` tests
- [x] `tsgo --emitDeclarationOnly` (type-check) clean
- [x] Full CI green (38/38 checks) — the one `fixture-hydrate` failure along the way was an unrelated `select`-component popover-positioning flake, confirmed by a clean re-run

## Out of scope (filed separately)

- **#2842** — `href={cond ? queryHref(base, {...}) : undefined}` (queryHref inside a ternary's `undefined`-alternate consequent) already emits invalid Go template syntax on `main`, independent of this bug (that path's consequent-only render doesn't consult the lowering registry at all). Confirmed unaffected by this PR's change.
- **#2843** — the 7 non-Go, non-Hono adapters refuse `queryHref` (or any lowering-registry call) inside a ternary attribute branch outright (`BF101`), a broader pre-existing gap surfaced while adding this PR's ternary fixture.

https://claude.ai/code/session_01P86NEtN4LGiyFYZWqESpQ6

---
_Generated by [Claude Code](https://claude.ai/code/session_01P86NEtN4LGiyFYZWqESpQ6)_

### DIFF
--- a/.changeset/go-query-href-attr-context-2743.md
+++ b/.changeset/go-query-href-attr-context-2743.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Fix #2743: `<a href={queryHref(base, {…})}>` on the Go adapter no longer diverges from the JS reference (Hono) when `base` contains non-URL-safe or multibyte bytes. `html/template` infers a URL context from the attribute name and percent-encodes the whole value there; the reference only HTML-escapes. A `queryHref`-produced (`query` guard-list) attribute value now emits the whole attribute via a new `bf_attr` runtime helper (`template.HTMLAttr`), the same technique `bf_spread_attrs`/`bfHydrationAttrs` already use to bypass html/template's contextual auto-escaping — keyed on the neutral `helper === 'query'` fact rather than the attribute name, so any URL-context attribute (`src`, `title`, …), not just `href`, gets the same byte parity.
--- a/packages/adapter-blade/src/conformance-pins.ts
+++ b/packages/adapter-blade/src/conformance-pins.ts
@@ -15,7 +15,7 @@ export const conformancePins: ConformancePins = {
   // rendering — it never consults the lowering registry for a branch the
   // way the direct-call attribute path does, so it falls to the generic
   // expression emitter, which refuses the params object literal.
-  'query-href-ternary': [{ code: 'BF101', severity: 'error', issue: 'https://github.com/piconic-ai/barefootjs/issues/2843' }],
+  'query-href-ternary': [{ code: 'BF101', severity: 'error', issue: 'https://github.com/piconic-ai/barefootjs/issues/2843', unescapable: { issue: 'https://github.com/piconic-ai/barefootjs/issues/2843' } }],
   'filter-typeof-predicate': [{ code: 'BF021', severity: 'error' }],
   'map-array-builder-body': [{ code: 'BF021', severity: 'error' }],
   'map-array-builder-escaping': [{ code: 'BF021', severity: 'error' }],
--- a/packages/adapter-blade/src/conformance-pins.ts
+++ b/packages/adapter-blade/src/conformance-pins.ts
@@ -10,6 +10,12 @@
 import type { ConformancePins } from '@barefootjs/jsx'
 
 export const conformancePins: ConformancePins = {
+  // #2843: a lowering-registry call (e.g. queryHref) inside a ternary
+  // attribute branch isn't recognized by this adapter's ternary-attribute
+  // rendering — it never consults the lowering registry for a branch the
+  // way the direct-call attribute path does, so it falls to the generic
+  // expression emitter, which refuses the params object literal.
+  'query-href-ternary': [{ code: 'BF101', severity: 'error', issue: 'https://github.com/piconic-ai/barefootjs/issues/2843' }],
   'filter-typeof-predicate': [{ code: 'BF021', severity: 'error' }],
   'map-array-builder-body': [{ code: 'BF021', severity: 'error' }],
   'map-array-builder-escaping': [{ code: 'BF021', severity: 'error' }],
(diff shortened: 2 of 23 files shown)


kfly8:
CI: the `fixture-hydrate` check failed on `ae5afe5`, but on an unrelated fixture — `[idempotence] select: replayed actions agree between hydrated and csr-mount`. The diff is in the Select popover's floating-position CSS (`top: -45.9062px` / `--radix-select-content-available-height: 761.9px` vs. `top: -691.906px` / `1407.9px` on the second measurement), not in any `href`/attribute output. This PR only touches Go-adapter attribute emission for `queryHref` values (`bf_attr`); it does not touch the Select component, popover positioning, or floating-ui-style layout math, so this looks like a pre-existing viewport/timing-dependent flake in that oracle test, not something introduced here.

Attempted to re-run the failed job to confirm, but GitHub returned "This workflow is already running" (other jobs in the same run were still in progress) — will re-run once the run finishes.

---

Issue #2828, opened 2026-09-03
collectAstPropRefs skips object-literal shorthand — a shorthand prop reference is never rewritten to _p.&lt;prop&gt;


kfly8:
## Summary

Found during #2741's fix (reactive-effect prop rewrite no longer touching object-literal keys), flagged explicitly by pullfrog review on PR #2819 as a gap that needed its own tracked issue rather than living only as PR-body prose.

`collectAstPropRefs` (`packages/jsx/src/prop-rewrite.ts:121`) explicitly skips a shorthand property assignment:

```ts
if (parent && ts.isShorthandPropertyAssignment(parent) && parent.name === n) return
```

So a prop referenced only via object-literal **shorthand** (`{ page }`, as opposed to `{ page: page }`) is never added to the discovered-refs set, and never rewritten to `_p.page`.

## Why this is a silent divergence, not a design choice

The rewrite exists specifically so a reactive-effect body's bare reference to a destructured/aliased prop name resolves to the live `_p.<prop>` read instead of a stale local. Shorthand (`{ page }`) and explicit (`{ page: page }`) are semantically identical JS — only the shorthand form is silently missed. The failure mode depends on what's in scope at the reference site:

- If a same-named local exists (e.g. the destructured parameter itself, or a variable from an enclosing scope), the rewrite silently reads that stale value instead of the live prop — a `ReferenceError` never fires, so this can render/execute with a wrong value and no diagnostic.
- If no such binding is in scope, it's a `ReferenceError` at the reactive-effect call site instead.

Either way it's the same "silently differs from the contract" class this repo's `known-limitation` `+ bug` tier is for.

## Repro shape

```tsx
"use client"
import { createSignal } from '@barefootjs/client'
export function Widget({ page }: { page: number }) {
  const [n] = createSignal(0)
  return <div onClick={() => console.log({ page })}>{n()}</div>
}
```

The effect/handler body's `{ page }` object-literal shorthand should read the live prop the same way `{ page: page }` does, but `collectAstPropRefs` never adds `page` to the discovered set for this call site, so no rewrite happens.

## Fix direction

[...]

Issue #2788, opened 2026-08-30
A renamed `children` destructure does not reach SSR on Mojo and Go


kfly8:
A component that destructures `children` under a different name compiles clean and then fails at SSR on two of the nine adapters. On Mojolicious it **dies at render time**: the emitted `.html.ep` interpolates the local alias, which the template stash never defines.

> **This body has been corrected twice and is now written from the repository's own artifacts rather than a local sweep.** The earlier per-adapter tables were wrong in both directions — see the History section at the bottom, kept because the way they were wrong is itself worth knowing.

## Measured

Source (the `children-passthrough-renamed` fixture added in #2787):

```tsx
export function Panel(props: { title?: unknown; children?: unknown }) {
  const { children: kids } = props
  return (
    <section class="panel">
      <h3>{props.title}</h3>
      {kids}
    </section>
  )
}
```

Emitted Mojo template:

```
<section class="panel" bf-s="<%= $bf->scope_attr %>" …><h3 bf="s1">…<%= $title %>…</h3><%= $kids %></section>
```

`$kids` is the **local** name. The stash carries `children`, so Perl dies inside `Mojo::Template::process`, by way of `BarefootJS::render_child`:

```
Traceback (most recent call first):
  File "/usr/local/share/perl/5.38.2/Mojo/Template.pm", line 163, in "Mojo::Template::process"
  File "render.pl", line 73, in "Mojo::Template::render"
  File "packages/adapter-perl/lib/BarefootJS.pm", line 351, in "main::__ANON__"
  File "template", line 3, in "BarefootJS::render_child"
```

## Per-adapter verdict

Derived from `ui/support-matrix.lock.json` on `main` versus #2787's head — the fixture's arrival raises each kind's `total` by one, so an adapter that renders it also gains a `pass`:

| adapter | `identifier` pass/total | `member` pass/total | verdict |
|---|---|---|---|
| hono | 327/330 → 328/331 | 186/189 → 187/190 | passes |
| blade | 311/330 → 312/331 | 170/189 → 171/190 | passes |
| erb | 312/330 → 313/331 | 171/189 → 172/190 | passes |
| **go-template** | 309/330 → **309**/331 | 168/189 → **168**/190 | **fails** |
| jinja | 311/330 → 312/331 | 170/189 → 171/190 | passes |
| minijinja | 311/330 → 312/331 | 170/189 → 171/190 | passes |
[...]

PR #2853, opened 2026-09-06
Fix #2828: collectAstPropRefs now discovers shorthand prop references

kfly8:
## Summary

- `collectAstPropRefs` (`packages/jsx/src/prop-rewrite.ts`) skipped a shorthand property assignment (`{ page }`) entirely when discovering which destructured props a reactive-attribute expression references. That means `rewriteBarePropRefs` found **zero** prop refs for text whose only prop reference was a shorthand one, and never attempted the rewrite at all — even though the emission side (`applyScopedPropRefRewrite`) already expanded a shorthand correctly (`{ page }` → `{ page: _p.page }`).
- Confirmed end-to-end: `queryHref(base, { tag, page })` inside a reactive `href` attribute emitted the bare shorthand names into the hydrate template lambda's module scope, where they don't exist as bindings — a `ReferenceError` at hydrate time.
- Fix: let a shorthand property's name participate in the discovered-refs set like any other value reference. The existing `isNonValuePosition` guard already keeps an explicit-key rewrite (`{ tag: tag }`, #2741) from touching the key, so no other behavior changes.

## Test plan

- [x] Added `query-href-shorthand-prop` conformance fixture (SSR + CSR) — the three-piece rule from `CLAUDE.md`.
- [x] Verified the fixture fails with the exact `ReferenceError: tag is not defined` before the fix (confirmed by reverting the fix locally and re-running).
- [x] `bun test packages/adapter-tests/src/__tests__/csr-conformance.test.ts -t query-href-shorthand-prop` passes.
- [x] `bun test packages/jsx/src/__tests__/` — no regressions (5 pre-existing, unrelated failures confirmed present on `main` too — a TS-checker-path resolver-migration issue unrelated to this change).
- [x] Regenerated `coverage-map.json` and `ui/support-matrix.lock.json` / `ui/compat.lock.json` for the new fixture; `git status` is clean after regeneration.
- [x] `biome check` clean on touched files.

Note: this PR is the base of a small stack — #2813 and #2788 fixes are stacked on top of it in separate PRs.

https://claude.ai/code/session_01ARqsMtqFeHTNdnxirGcQVd

---
_Generated by [Claude Code](https://claude.ai/code/session_01ARqsMtqFeHTNdnxirGcQVd)_


kfly8:
Thanks for the catch — verified and fixed in 1f5f7f4.

`collectBranchLocalPropRefsViaSubstitution` (`jsx-to-ir.ts`) dropped its own `isShorthand` exclusion, matching what this PR did for `collectAstPropRefs`. Confirmed with a repro: before the fix, a branch-local read from a prop and referenced via shorthand inside a reactive expression left a bare, unprefixed prop-name reference in the CSR hydrate template lambda (module scope — no such binding, `ReferenceError` at a pure CSR mount); after the fix it resolves through `_p.<prop>` like every other path. Pinned with a regression test in `rewrite-destructured-props.test.ts`.

While verifying this I found the full repro also hits a **separate, pre-existing** bug: `replaceBranchLocalRefs`'s regex-based substitution corrupts object-literal shorthand by dropping the implicit key (`{ local }` → `{ (_p.tag) }`, invalid JS) — reproduces identically with or without this PR's fix, so it's unrelated. Filed as #2856 rather than folding a larger AST-migration fix into this PR; the regression test above documents that it pins only the prop-dependency question, not that surrounding shape.

---

kfly8:
CI red on `test (rest)`: the `generated-data-points.json` freshness meta-test failed because this PR's `query-href-shorthand-prop` fixture wasn't reflected in that committed artifact (missed running `bun packages/adapter-tests/scripts/generate-data-points.ts` alongside the other regenerations). Fixed in 994f929 — purely additive (74 new lines for the new fixture's adversarial data points), verified `generated-data-points.test.ts` now passes (15/15). Rebased #2854 and #2855 on top so the fix propagates through the stack.

---

PR #2855, opened 2026-09-06
Fix #2788: renamed children destructure now reaches SSR on Mojo and Go

kfly8:
## Summary

Stacked on #2854 (base is that branch, not `main`).

A component using the bare-props form (`function Foo(props: Props)`) whose body destructures a prop under a different local name (`const { children: kids } = props`) compiled clean and died at SSR:

- Mojolicious: Perl strict-mode `Global symbol "$kids" requires explicit package name`.
- Go: template execution failure, `can't evaluate field Kids`.

**Root cause:** `propsParams` for the bare-props form comes from the TYPE annotation (`extractPropsFromTypeMembers`), which has no notion of a body-level rename — unlike parameter destructuring (`{ n: count }`), which already carries `ParamInfo.sourceName`. The analyzer already resolves such a destructuring statement into an ordinary `localConstants` entry whose `parsed` value is a plain `props.<key>` member read; recognizing that shape is enough to answer the alias question without a second AST walk of the destructuring pattern itself.

Added `resolveBodyDestructuredPropAliases` (`props-binding.ts`, exported alongside the existing `buildPropAliasMap`) and wired it into:

- `ssr-defaults.ts`: seeds the stash under the local name too, alongside the caller-facing key.
- `go-template-adapter.ts`: `rootFieldRef` also resolves through this alias map before capitalizing into a struct field.

Not children-specific — the same gap affects **any** renamed body-level prop destructure on any template-stash adapter, but the other seven already-passing adapters mask it for a no-children-passed test case (a wrong stash lookup and a genuinely absent value both render as "nothing").

Also fixes `buildChildDefaultsPerl` (`adapter-mojolicious/test-render.ts`), which had the same hand-enumeration gap fixed for root-level components in #2854's PR, this time for CHILD components — `children-passthrough-renamed`'s `Panel` is a child of `Page`.

## Real-interpreter verification

- Perl `Mojolicious` (already installed for #2854).
- Go 1.25.6 (already installed for #2854).

Deleted the `children-passthrough-renamed` `render-divergences.ts` pin on both Mojolicious and Go.

## Test plan

- [x] `children-passthrough-renamed` passes against real Mojolicious (Perl) and Go 1.25.
- [x] Regression-checked an unrelated already-passing adapter (ERB) for the same fixture — still passes.
- [x] Regenerated `coverage-map.json` (no change — no new fixture here), `ui/support-matrix.lock.json`, `ui/compat.lock.json` — `git status` clean after regeneration.
- [x] `biome check` clean on touched files.
- [x] Full-stack drift check (coverage-map + expected-html + support-matrix) on this branch's tip (all three PRs combined) — zero diff.

https://claude.ai/code/session_01ARqsMtqFeHTNdnxirGcQVd

---
_Generated by [Claude Code](https://claude.ai/code/session_01ARqsMtqFeHTNdnxirGcQVd)_

### DIFF
--- a/.changeset/children-passthrough-renamed-2788.md
+++ b/.changeset/children-passthrough-renamed-2788.md
@@ -1,0 +1,7 @@
+---
+"@barefootjs/jsx": patch
+"@barefootjs/go-template": patch
+"@barefootjs/mojolicious": patch
+---
+
+Fix #2788: a bare-props-form component (`function Foo(props: Props)`) whose body destructures a prop under a different local name (`const { children: kids } = props`) now reaches SSR correctly on Mojolicious and Go instead of dying/failing template execution. Added `resolveBodyDestructuredPropAliases` (`props-binding.ts`), which recognizes the analyzer's `props.<key>` member-read shape for such a destructure and seeds/routes the local name alongside the caller-facing key.
--- a/packages/adapter-go-template/src/adapter/go-template-adapter.ts
+++ b/packages/adapter-go-template/src/adapter/go-template-adapter.ts
@@ -82,6 +82,7 @@ import {
   buildImportAliasMap,
   resolveGetterAliases,
   collectAliasableGetterNames,
+  resolveBodyDestructuredPropAliases,
 } from '@barefootjs/jsx'
 import { findInterpolationEnd } from '@barefootjs/jsx/scanner'
 import { BF_REGION, escapeHtml, resolveJsxChildrenProp } from '@barefootjs/shared'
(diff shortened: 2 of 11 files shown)


kfly8:
Thanks for the review. On the filter-predicate alias-resolution bypass: deferring it rather than folding into this PR — it's pre-existing (not introduced by #2813 or #2788), unexercised by any fixture today, and fixing it properly means a conformance fixture + touching `renderFilterExprNode`'s `identifier`/`call` arms, which is more than this stack's scope.

Filed as #2857 with the two call sites, the required fix (use `rootFieldRef`'s resolved return value instead of re-deriving from the raw identifier), and the fixture needed to graduate it.

---

Issue #2189, opened 2026-07-09
Event handler referencing the .map() index param throws "ReferenceError: i is not defined" at runtime (event-delegation lowering drops the index)


kfly8:
## Summary

When a list item's event handler closes over the `.map()` callback's **index** parameter (as opposed to the item itself, or an item property), the compiled client bundle throws a `ReferenceError` at runtime the first time the event fires. Accessing item properties (e.g. `item.id`) in the same handler works fine — only the index variable is affected.

## Environment

- `@barefootjs/client` / `@barefootjs/cli` / `@barefootjs/hono`: `0.18.4`
- Adapter: Hono / Cloudflare Workers (`--adapter hono` scaffold)
- Node: v22.22.2

## Minimal repro

```tsx
'use client'

import { createSignal } from '@barefootjs/client'

interface Item {
  id: number
  label: string
}

export function IssueRepro() {
  const [items] = createSignal<Item[]>([
    { id: 1, label: 'a' },
    { id: 2, label: 'b' },
  ])

  const handle = (index: number) => {
    console.log('clicked index', index)
  }

  return (
    <ul>
      {items().map((item, i) => (
        <li key={item.id}>
          <button type="button" onClick={() => handle(i)}>
            {item.label}
          </button>
        </li>
      ))}
    </ul>
  )
}
```

Render this component, click any `<button>` in the browser. Expected: `console.log('clicked index', 0)` (or `1`). Actual:

```
Uncaught ReferenceError: i is not defined
```

## Root cause (from the compiled output)

`bf build` lowers the per-item click handler into a single delegated `click` listener on the list's container, which looks up the clicked item via `closest('[data-key]')` and re-finds it in the live array — but it only threads the **item** (as `item2`) through this delegation, not the **index** (`i`). The generated code (`components/IssueRepro.client.js`, formatted for readability) is:

```js
_s2 && _s2.addEventListener('click', __bfEvt => {
  const s1El = __bfEvt.target.closest('[bf="s1"]')
  if (s1El) {
    const li = s1El.closest('[data-key]')
    if (li) {
      const key = li.getAttribute('data-key')
      items().find(item2 => String(item2.id) === key) &&
        (() => handle(i))(__bfEvt)   // <-- `i` is not in scope here at all
    }
    return
  }
})
```

[...]

PR #2860, opened 2026-09-06
Fix #2859: keyed .map() row index staleness after a same-key reorder

### LINKED ISSUE #2859: Keyed .map() list: index-derived render output not refreshed for a row that moves position after a reorder
## What happened?

In a keyed `.map()` list backed by a signal (`items().map((item, i) => <li key={item.id}>...</li>)`), reordering the array (same keys, new order — e.g. moving one item to a new position) correctly moves each row's own item-derived content (text/attributes read from the item itself, or from a `key`-based signal lookup). However, any expression in the row's JSX that reads the `.map()` callback's **index parameter** (`i`) directly — not through a signal — keeps showing the value from wherever that row was *first created*, not its current position.

This happens for a plain text interpolation (`{String(i + 1)}`, no signal involved at all) exactly the same as for an expression that combines the index with a real signal (`selectedIndex() === i`) — in the latter case the signal read (`selectedIndex()`) does correctly retrigger the effect, but the comparison is made against the stale `i`, so the "selected" row visually stays attached to whichever DOM node happened to be created at that index, instead of following the item that is now actually at that index.

Root cause (from `packages/client/src/runtime/map-array.ts`): `mapArray`'s per-row `renderItem(itemAccessor, index, existing)` receives `index` as a plain JS number, captured once when the row is first created (`createItemScope`). On a same-key update (the reorder-only case), `mapArray` only calls `existing.setItem(item)` to push the new item value into the row's own per-item signal — it never re-invokes `renderItem`, and nothing else exists to refresh whatever effect closure captured the original `index`. So any binding built from that raw `index` — rather than from a reactive re-derivation of "this key's current position" — is permanently stale for a row that survives a reorder by key.

For contrast: I also checked whether an event handler that closes over the same index (`onClick={() => handle(i)}`) has the same problem, since #2189/#2191 previously addressed a related index-in-handler issue. It does not — `bf build`'s delegated click dispatch re-derives the current index at click time via `arr.findIndex(...)` keyed off `data-key`, so a moved row's click handler already fires with the correct, current index. The bug reported here is specific to index-derived **render output** (text/attribute/class bindings evaluated in the loop body), not event handlers.

## Component source that reproduces it

```tsx
"use client"

type Item = { id: string; label: string }

export function ReorderRepro() {
  const [items, setItems] = createSignal<Item[]>([
    { id: 'a', label: 'Alpha' },
    { id: 'b', label: 'Beta' },
    { id: 'c', label: 'Gamma' },
    { id: 'd', label: 'Delta' },
    { id: 'e', label: 'Epsilon' },
  ])
  const [selectedIndex] = createSignal(2)

  // Moves item 'a' (index 0) so it lands between 'c' and 'd' —
  // same keys persist, only their order changes: a,b,c,d,e -> b,c,a,d,e
  const reorder = () => {
    setItems(([a, b, c, d, e]) => [b, c, a, d, e])
  }

  return (
    <div>
      <button id="reorder-btn" onClick={reorder}>Reorder</button>
      <ul>
        {items().map((item, i) => (
          <li key={item.id} class={selectedIndex() === i ? 'row selected' : 'row'} data-item-id={item.id}>
            <span class="badge">{String(i + 1)}</span>
            <span class="label">{item.label}</span>
          </li>
        ))}
      </ul>
    </div>
  )
}
```

## Steps to reproduce

1. Compile and run the component above with `bf build` / `@barefootjs/vite` (CSR adapter) and load it in a browser.
2. Note the initial render: rows in key order `a,b,c,d,e`, badges `1,2,3,4,5`, and row `c` (index 2) has the `selected` class.
3. Click "Reorder". This calls `setItems(([a,b,c,d,e]) => [b,c,a,d,e])` — same 5 keys, new order.
4. Inspect the rendered `<li>` elements' key, badge text, and `selected` class.

## Expected vs. actual

**Expected**, after the reorder produces the new key order `b,c,a,d,e`:
- Rows render in DOM order `b,c,a,d,e` (this part already works correctly).
- Badges read `1,2,3,4,5` in that new order (row `a`, now at index 2, shows `3`; row `d`, still at index 3, keeps showing `4`; etc.).
- The `selected` class is on whichever row is now truly at index 2 — row `a` (since `selectedIndex()` is still `2`).

**Actual** (verified in a real browser against the compiled CSR output):
- DOM order is correctly `b,c,a,d,e`.
- Badges remain `2,3,1,4,5` — every row keeps the badge from its *original* index at creation time, not its current position. Only rows `d` and `e`, whose position didn't change, show the correct value.
- The `selected` class stays on row `c` — the row that was originally created at index 2 — instead of moving to row `a`, which is now actually at index 2.

## Why this matters

This is not just a cosmetic display bug. Anything computed from the `.map()` callback's raw index parameter inside the loop body — a row number, a "same index as X" comparison, a lookup into an array/object keyed by position (`someArray[i]`) — silently keeps pointing at the row's original position for as long as that row's key survives, even though the row's own item content updates correctly. For any UI with a reorderable (drag-and-drop, sortable) keyed list, this means index-derived visuals go out of sync with the actual data the moment a row moves, without any error or warning — the kind of divergence that's easy to miss in testing (since a full remount, e.g. changing the array's length, "fixes" it) but shows up quietly in production once users start reordering things.

## Environment

- `@barefootjs/client` / `@barefootjs/jsx` / `@barefootjs/shared` / `@barefootjs/vite`: `0.33.6` (exact resolved version, via `^0.33.6` in package.json)
- Adapter: CSR (`@barefootjs/client/csr-adapter`), built with `@barefootjs/vite` + Vite 6
- Runtime: verified in a real browser (Chromium via Playwright) against the `vite build` output, served with `vite preview`
- OS: macOS (Darwin)

kfly8:
## Summary

Fixes #2859.

In a keyed `.map()` list, reordering the array by key (same keys, new order) correctly moves each row's item-derived content, but any expression that reads the `.map()` callback's raw **index** parameter directly (a text/attribute/class binding, or a direct event handler) kept showing the value from wherever the row was first created — it never updated to reflect the row's new position.

### Root cause

- **Runtime** (`packages/client/src/runtime/map-array.ts`): `mapArray`/`mapArrayAnchored` already give each row a per-item `createSignal` for its *item* value, and push new values through it (`setItem`) on every same-key reconcile — that's what keeps item-derived content live across a reorder. The row's *index*, by contrast, was a plain JS number captured once at row creation, with no equivalent signal and no update path.
- **Compiler** (`packages/jsx/src/ir-to-client-js`): `wrapLoopParamAsAccessor` already rewrites every reference to the loop's *item* parameter into a call through the item accessor (`item` → `item()`) wherever it builds reactive attrs/texts, preamble, refs, and component props — which is what makes item-derived output reactive in the first place. It never did the same for the *index* parameter, so even with a live index signal, nothing in the emitted code would have read it.

### Fix

- `ItemScope`/`AnchorScope` in `map-array.ts` gain a second per-row signal for the row's current index; `renderItem` now receives an **index accessor** (mirroring the existing item accessor) instead of a plain number, and the same-key reconcile path pushes the row's current position through it (batched with `setItem` so a row whose effect reads both doesn't run twice).
- `wrapLoopParamAsAccessor` is extended to also rewrite the loop's index parameter into an accessor call, and every loop-plan builder (plain, composite, component, branch, nested/inner, conditional-arm) threads the loop's `index` field through it — one shared implementation change covering every loop shape uniformly, rather than a one-off patch on the reported repro.
- Static (non-keyed) `.map()`-to-`forEach` loops and the lazy-row fast path are unaffected: static loops never scope index through a signal at all, and lazy-row eligibility already refuses any loop whose reactive bindings reference the index parameter.

### A related bug found while auditing similar cases

A **direct** (non-delegated) event listener on a composite/inner/conditional-arm loop row that closes over the raw index (`onClick={() => handle(i)}`) had the identical staleness bug — these loop shapes attach listeners once via `addEventListener` rather than going through the top-level delegated click dispatcher (`build-event-delegation.ts`, #2189/#2191), which already re-derives the current index at click time via `data-key`/`findIndex` and was already correct. Fixed the same way, by extending the same wrap.

### Known remaining gaps (not in scope here, worth a follow-up issue)

- An index-only expression with **no signal read and no function call at all** (e.g. bare `{i}` text, or `class={i % 2 === 0 ? 'a' : 'b'}`) is still classified as fully static by Phase 2 reactivity analysis and never gets a `createEffect` wired at all — this is a pre-existing gap (confirmed unaffected/unchanged by this PR, see `nested-loop-index-param.test.ts`'s "template text interpolation" case), not a new regression, but the same class of problem as #2859 for a narrower shape of expression.
- A nested loop's body referencing the *outer* loop's own index (as opposed to its own) isn't covered here either.

I'll file a follow-up issue for both once this lands.

## Design

The design for this fix was done in consultation with Fable before implementation — investigating the runtime/compiler split above, confirming the direct-listener event-handler bug as the same root cause, and confirming the lazy-row exclusion stays sound.

## Test plan

- [x] New runtime unit tests in `packages/client/__tests__/runtime/map-array.test.ts`: index accessor tracks a row's current position across a reorder, a stationary row's index effect doesn't re-run on an unrelated reorder, hydration seeds the index accessor from SSR ordinal position.
- [x] Updated compiler unit tests whose expectations pinned the old (stale) behavior as correct, now asserting the fixed `i()` accessor-call form: `loop-child-reactive-attr-const-shadow.test.ts`, `loop-item-conditional-codegen.test.ts`, `nested-loop-index-param.test.ts`, `nested-loop-reactive-attrs.test.ts`.
- [x] Full `packages/client` test suite: 701 pass, 0 fail.
- [x] Full `packages/adapter-tests` suite: 2376 pass, 9 fail — all 9 failures are a pre-existing environment issue (`@barefootjs/client` package not built in this fresh checkout), unrelated to this change; no SSR/HTML fixture drift, confirming this change doesn't touch adapter output.
- [x] Full `packages/jsx` test suite, minus the 5 pre-existing failures confirmed present on `main` before this change (unrelated resolver-alias tests).
- [x] `bun run <package>` typecheck (`tsgo --noEmit`) clean for both `packages/client` and `packages/jsx`.

Added a changeset (`patch` for `@barefootjs/client` and `@barefootjs/jsx`).

---


kfly8:
`update-expected-html` was red on the first commit, and it caught a real bug rather than a stale snapshot: `select-native-shared`'s row uses `mapArrayLazy`, and the first commit's index-accessor wrap was applied unconditionally to a plain/branch loop's row template and preamble — but that same string is reused verbatim by the lazy row plan when the loop turns out lazy-eligible, and `mapArrayLazy`'s `createRow(entry, index)` still hands the row a plain index **number**, never an accessor (`map-array-lazy.ts` is intentionally untouched by this fix). So the wrap baked `i()` into a template lazy would later interpolate with a plain `i`, which throws at runtime. This also explains why `e2e-site-ui`'s `field-arrays.spec.ts` (add/remove field) and `controlled-input-focus.spec.ts` (field-arrays case) failed on the same commit — `field-arrays-demo`'s row is lazy-eligible and uses the index in its `placeholder` text.

Fixed in 5ba7a04: build the row template/preamble with only the *item* wrapped first (safe either way — lazy-row eligibility is decided from that), and apply the index wrap as a second pass only once the loop is confirmed **not** going lazy.

Verified locally:
- `bun run packages/adapter-tests/scripts/generate-expected-html.ts` now produces a zero diff (confirmed with `@barefootjs/client` actually built, so the previously-masked `select-native-shared` case was included).
- Re-compiled `field-arrays-demo.tsx` directly and confirmed its row template now emits the correct, unwrapped `` `Email ${index + 1}` `` (lazy-safe) instead of the broken `` `Email ${index() + 1}` ``.
- Full `packages/client` and `packages/jsx` suites still green (same pre-existing 5 failures on `main`, unrelated to this change).

The `toast-queue.spec.ts` failure on the first commit's `e2e-site-ui (4)` shard looked unrelated to this change (a different component); will re-check once this run finishes.

---

PR #2872, opened 2026-09-07
fix(jsx): dedup reactive attribute writes against their previous value

### LINKED ISSUE #2869: Consolidated per-row loop effect has no previous-value guard — untrack can't prevent reassignment when a sibling binding retriggers the shared effect
## What happened

A keyed `.map()` row whose bindings span multiple elements gets compiled to **one shared `createEffect` per row** (`emitConsolidatedRowEffect`, `packages/jsx/src/ir-to-client-js/control-flow/stringify/reactive-effects.ts:242-315`, dispatched from `stringifyReactiveEffects()` at lines 111-145; introduced by `e71e19d86` — "perf(jsx): one effect per loop row" — per `spec/slot-unification.md` §3(c)/§8). That consolidation is a reasonable perf trade-off on its own.

The problem is what happens *inside* that shared effect: every attribute write goes through `emitAttrUpdate()`'s generic fallback (`packages/jsx/src/ir-to-client-js/emit-reactive.ts:123-125`):

```js
{ const __v = <expr>; if (__v != null) el.setAttribute('name', String(__v)); else el.removeAttribute('name') }
```

with **no comparison against the attribute's current value** — `style` (99-103), `class` (104-108) and `dangerouslySetInnerHTML` (90-98) are the same shape. Only the form-`value` case (`emitValueUpdateStatements`, 54-59) checks the previous value first. `srcdoc` isn't special-cased either — `classifyDOMProp()` (`packages/shared/src/dom-prop.ts:209-223`) falls through to `{ kind: 'attr' }` for it, so it goes through the same unguarded generic branch.

Put those two facts together and `untrack()` stops being a reliable "compute this once" tool the moment a binding shares its row's fused effect with **any** other genuinely-reactive sibling binding — which, in a row with more than one or two dynamic bindings, is close to unavoidable. `untrack` only suppresses *new dependency registration* for the read inside it (`packages/client/src/reactive.ts:620-628` just nulls the tracking `Listener` around `fn()`) — it does nothing to stop the *surrounding* effect from re-running for an unrelated reason and then unconditionally re-executing + re-writing the "untracked" expression's result anyway.

I hit this trying to freeze an `<iframe srcdoc>` binding inside a keyed `.map()` row (`srcdoc={untrack(() => expensiveDoc())}`) so it would load exactly once — reassigning `srcdoc`, even to a byte-identical string, forces a real navigation/reload (verified separately: reassigning an identical string 3 times in a row fires 3 `load` events). A completely unrelated sibling binding in the same row, reading the row's own item accessor directly (e.g. `data-key={item.key}`, or in my real case a `title` built from a per-key index signal), legitimately needs to react when the row's content changes — which reruns the whole fused effect and reassigns the "frozen" `srcdoc` anyway, producing a visible reload on every edit to that row's own data, despite the `untrack`.

## Why `untrack` alone can't fix this — and the docs don't say so

`docs/core/reactivity/untrack.md` (27-34) documents `untrack` purely as "won't trigger *this* effect [for the untracked read]" — there's no example or caveat for the case where the *surrounding* effect reruns for an unrelated dependency and re-executes the untracked expression anyway. I initially assumed (reasonably, given the docs) that wrapping the whole expression in `untrack` was sufficient, and only discovered otherwise via `bf debug graph`/`bf debug loops` (which correctly showed the unrelated bindings sharing one slot ID) plus reading the compiled output directly. This is a real documentation gap on top of the underlying behavior gap.

## You've already built the fix — just on the other loop strategy

`packages/client/src/runtime/claim-slots.ts:81-92` documents "Dedup, no trust-first-run" for `'markup'` writes, and the *lazy-row* loop strategy's `dedupGuard()` (`control-flow/stringify/lazy-row.ts:453-455`, `Object.is`-based) already applies per-binding dedup to attribute writes on that path. `spec/slot-unification.md` §9 ("Lazy row graph") documents this dedup as an explicit design goal — but only for lazy-row. The *plain*-loop consolidated-row-effect path (`emitConsolidatedRowEffect`) — which a loop falls back to whenever `lazy-row-eligibility.ts` excludes it (one such exclusion, at line 288, is any binding referencing the loop's index parameter — an extremely common pattern, and the same mechanism #2861 discusses in a different context) — has no equivalent guard at all.

## Component source that reproduces it

```tsx
"use client"

import { createSignal, createEffect, untrack } from '@barefootjs/client'

type Item = { key: string; text: string }

const indexSignals = new Map<string, [() => number, (v: number) => void]>()
function indexSignal(key: string): [() => number, (v: number) => void] {
  let s = indexSignals.get(key)
  if (!s) { s = createSignal(0); indexSignals.set(key, s) }
  return s
}

const fragmentSignals = new Map<string, [() => string, (v: string) => void]>()
function fragmentSignal(key: string): [() => string, (v: string) => void] {
  let s = fragmentSignals.get(key)
  if (!s) { s = createSignal(''); fragmentSignals.set(key, s) }
  return s
}

export function RealisticFusionRepro() {
  const [items, setItems] = createSignal<Item[]>([
    { key: 'a', text: 'hello' },
    { key: 'b', text: 'world' },
  ])

  createEffect(() => {
    items().forEach((item, i) => {
      const [ig, is] = indexSignal(item.key)
      if (ig() !== i) is(i)
      const [fg, fs] = fragmentSignal(item.key)
      if (fg() !== item.text) fs(item.text)
    })
  })

  const edit = () => setItems(prev => [{ key: prev[0].key, text: prev[0].text + 'x' }, prev[1]])

  return (
    <div>
      <button id="edit-btn" onClick={edit}>Edit</button>
      <ul>
        {items().map(item => (
          <li key={item.key} data-key={item.key}>
            <span class="idx">{String(indexSignal(item.key)[0]() + 1)}</span>
            <div
              class="frozen"
              data-frozen={untrack(() => {
                const w = window as unknown as { __untrackRuns?: number }
                w.__untrackRuns = (w.__untrackRuns ?? 0) + 1
                return fragmentSignal(item.key)[0]()
              })}
            />
          </li>
        ))}
      </ul>
    </div>
  )
}
```

## Steps to reproduce

1. Compile and run the component above with `@barefootjs/vite` (CSR adapter) and load it in a browser.
2. Note the compiled output: `data-key`, the `idx` text slot, and the `untrack`-wrapped `data-frozen` all live inside **one** `createEffect(...)` for the row (confirmed by inspecting the built JS — the same `S(() => { if (l) {...data-key...}; if (c) {...data-frozen via untrack...}; w("s1", ...idx...) })` block).
3. Click "Edit" repeatedly. Each click gives row `a` a new item object (same key), and the driving `createEffect` above pushes the new `text` into that key's `fragmentSignal`.
4. Inspect `window.__untrackRuns` (incremented *inside* the `untrack` callback) and the `frozen` div's `data-frozen` attribute after each click.

## Expected vs. actual

**Expected**: since the `data-frozen` binding's only signal read is wrapped in `untrack`, it should never re-register as a dependency and — critically — its value should never need to change after the row's first render (that's the whole point of wrapping it). `window.__untrackRuns` should stay at whatever it was after the initial mount.

**Actual** (verified against the compiled CSR output in a real browser):
- `idx` correctly keeps showing `1` (row `a`'s position never changes).
- `window.__untrackRuns` increments by exactly 1 on **every** click (5 clicks → 5 more runs), and `data-frozen` is reassigned every time (`hellox`, `helloxx`, `helloxxx`, ... tracking the live, current text) — even though the only signal read inside the `untrack` callback was supposed to make this binding inert after mount.

## Suggested fix shape

1. **Highest priority, cheapest, works regardless of row-effect granularity**: give `emitAttrUpdate()`'s generic fallback (and the `class`/`style`/`dangerouslySetInnerHTML` branches) the same previous-value guard the lazy-row path's `dedupGuard()` already has (`Object.is`-based, `control-flow/stringify/lazy-row.ts:453-455`) — this alone would have prevented the `srcdoc` reload regardless of the `untrack` interaction, since a stable expression would simply stop writing once its value settles.
2. Document `untrack`'s actual contract precisely in `docs/core/reactivity/untrack.md`: it suppresses dependency *registration* for the wrapped read, not *re-execution* of the surrounding effect when something else in the same effect legitimately changes. As written today the docs read as if wrapping an expression in `untrack` is sufficient to freeze it, which is only true if that expression is the effect's *only* binding.
3. Lower priority / open design question: whether it's worth extending `lazy-row-eligibility.ts` so that having an index-parameter-derived binding doesn't unconditionally disqualify a loop from the (already-guarded) lazy-row path, rather than only fixing the plain-loop path in isolation.

The only reliable workaround I found (short of (1)) was moving the write out of JSX reactivity entirely into a `ref` callback — confirmed empirically that `ref` fires exactly once at row creation and is never re-invoked by the row's shared effect, even after 5+ content updates to that row.

## Environment

- `@barefootjs/client` / `@barefootjs/jsx` / `@barefootjs/shared` / `@barefootjs/vite`: `0.35.0`
- Adapter: CSR (`@barefootjs/client/csr-adapter`), built with `@barefootjs/vite` + Vite 6
- Runtime: verified in a real browser (Chromium via Playwright) against the `vite build` output
- OS: macOS (Darwin)

## Related

- #2859 / #2860 / #2861 — the same `packages/jsx/src/ir-to-client-js/control-flow/` loop-compilation area, also involving per-row index handling.

kfly8:
## Summary

Fixes #2869.

An eager reactive-attribute write (`emitAttrUpdate`'s generic/`class`/`style`/`dangerouslySetInnerHTML` branches) rewrote the DOM unconditionally on every `createEffect` rerun, no matter which binding in that effect actually changed. This bites hardest in a keyed `.map()` row, where multiple bindings are fused into ONE shared `createEffect` for perf (`emitConsolidatedRowEffect`): a sibling binding's legitimate change reruns the whole row effect and rewrites every attribute in it, including ones whose own computed value never changed.

`untrack()` cannot prevent this on its own — it only suppresses dependency *registration* for the wrapped read, not *re-execution* of the surrounding effect when a sibling binding legitimately changes. So an `<iframe srcdoc={untrack(() => doc)}>` sharing a row's fused effect with an unrelated reactive sibling (e.g. `data-key={item.key}`) kept reloading on every unrelated update to that row.

The lazy row graph (`control-flow/stringify/lazy-row.ts`) already had exactly this guard — an `Object.is` comparison against a per-row last-value store — just not applied to the plain-loop consolidated-row-effect path or to the ordinary (non-loop) reactive-attribute effects.

- Extracts the guard into a shared `emitDedupedAttrUpdate` / `dedupGuard` / `DEDUP_STORE_DECL` in `emit-reactive.ts`.
- Routes **every** eager reactive-attribute emission site through it (there were 8, not just the loop path): `emitReactiveAttributeUpdates`, `emitReactiveChildProps`, `emitAttrSlotsGranular` (profile mode), `emitConsolidatedRowEffect`, `stringifyBranchReactiveAttrs`, the branch-arm `insert()` path, the inner-loop path, and the static-loop materialize path.
- `lazy-row.ts`'s own `emitAttrBinding` now calls the same shared helper instead of duplicating the guard logic — the only remaining substrate-specific piece is *where* its `__l` store comes from (the reconciler's `entry.last`, since `applyItem`/`applyOuter` are separate invocations with no shared closure of their own).
- Documents `untrack`'s actual contract in `docs/core/reactivity/untrack.md` (stops dependency registration for the wrapped read, not re-execution of the surrounding effect).

Only client JS *behavior* changes (write frequency inside an already-running effect) — SSR/CSR template HTML is byte-identical before and after, confirmed by running `generate-expected-html.ts` and diffing (only `.client.js` snapshots changed, no `.html` files).

## Design

The fix shape (which sites to guard, the shared-helper structure, the `untrack.md` doc patch, and the test plan) was designed by Fable (`claude-fable-5-1`) before implementation, per this repo's convention of separating design from implementation.

## Test plan

- [x] `packages/client/__tests__/runtime/issue-2869-attr-dedup-guard.test.ts` — new behavioral regression test. Compiles real repro components (a fused keyed-row effect and a plain non-loop effect, both with an `untrack`-wrapped sibling binding) and hydrates them against `@happy-dom/global-registrator`, asserting `setAttribute` call counts before/after triggering a sibling update. Fails on `main`, passes with this fix.
- [x] `packages/jsx/src/__tests__/issue-2869-attr-dedup-guard.test.ts` — new codegen-shape pins, plus a shrink-only ledger asserting no eager emission site calls `emitAttrUpdate` directly outside `emit-reactive.ts`.
- [x] Updated 3 pre-existing codegen-shape test pins (`style-css-var-reactive.test.ts`, `child-prop-fallback-wrap.test.ts`, `nested-loop-reactive-attrs.test.ts`) that asserted the old unguarded write shape — behavior unchanged, only the intermediate temp-variable shape changed.
- [x] Regenerated `packages/jsx/src/__tests__/__snapshots__/doc-examples.test.ts.snap` and reviewed the diff (every hunk is either the new guard wrapper around an unchanged write, or the new `untrack.md` example).
- [x] Ran `packages/adapter-tests/scripts/generate-expected-html.ts` and confirmed only `.client.js` fixture snapshots changed (no `.html` diffs) — the fixture-drift check CLAUDE.md requires before merging.
- [x] Full suite: `bun test packages/jsx packages/client packages/test packages/adapter-tests packages/router` — 6553 pass, 0 fail.
- [x] `bun test ui/components/` — 1240 pass, 0 fail.

https://claude.ai/code/session_01GGEdZGGtKh3hy1DfMcn59A

---
_Generated by [Claude Code](https://claude.ai/code/session_01GGEdZGGtKh3hy1DfMcn59A)_

### DIFF
--- a/.changeset/attr-dedup-guard-2869.md
+++ b/.changeset/attr-dedup-guard-2869.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix eager reactive-attribute writes rewriting the DOM unconditionally on every `createEffect` rerun, regardless of which binding in that effect actually changed. This mattered most for a keyed `.map()` row's fused row effect, where a sibling binding's legitimate change reran the whole effect and rewrote every attribute in it — including one wrapped in `untrack()`, which only suppresses dependency registration, not re-execution of the surrounding effect. Every eager reactive-attribute write is now guarded against its previous value (`Object.is`), matching the guard the lazy row graph already had.
--- a/docs/core/reactivity/untrack.md
+++ b/docs/core/reactivity/untrack.md
@@ -67,3 +67,43 @@ createEffect(() => {
   setResult(a + b)
 })
 ```
+
+
+## What `untrack` Does Not Do
+
+`untrack` only stops the wrapped read from **registering a dependency**. It does not stop the surrounding effect from **re-running** — and when the effect re-runs for any other reason, the code inside `untrack` runs again too.
+
+This matters because BarefootJS groups several bindings into one effect for performance: every reactive attribute on one element shares an effect, and a keyed `.map()` row's attributes and text share a single row effect. When one binding in that group changes, the whole effect re-runs, and every expression in it — untracked or not — is evaluated again.
+
+```tsx
+'use client'
+import { createSignal, untrack } from '@barefootjs/client'
+
+function renderPreview(id: number): string {
+  return `<p>Preview for item ${id}</p>`
+}
+
+export function Gallery() {
+  const [items, setItems] = createSignal([{ id: 1, label: 'First' }, { id: 2, label: 'Second' }])
+  const rename = () => setItems(items().map(item => ({ id: item.id, label: item.label + '!' })))
+  return (
+    <div>
+      <button onClick={rename}>Rename all</button>
+      <ul>
+        {items().map(item => (
+          <li key={item.id} title={item.label} data-preview={untrack(() => renderPreview(item.id))}>
+            {item.label}
+          </li>
+        ))}
+      </ul>
+    </div>
+  )
+}
+```
+
+Clicking "Rename all" changes `title` on every row, so every row effect re-runs and `renderPreview` is called again for each row even though it is wrapped in `untrack`. Two guarantees hold regardless:
+
+- The untracked read still does not subscribe: a signal read only inside `untrack` never triggers the effect on its own.
+- A DOM write is skipped when the binding's new value is the same as the last one it computed (compared with `Object.is`). `data-preview` above is recomputed on every run but not rewritten, so an attribute whose write is expensive to apply — `srcdoc` on an `<iframe>`, which reloads the frame on every assignment — stays untouched until its value actually changes.
+
+If the **computation** itself is what you need to avoid repeating, `untrack` is the wrong tool: put it in a `createMemo` so it only re-runs when its own inputs change.
--- a/packages/adapter-tests/fixtures/__snapshots__/accordion.client.js
+++ b/packages/adapter-tests/fixtures/__snapshots__/accordion.client.js
@@ -39,11 +39,16 @@ export function initCheckIcon(__scope, _p = {}) {
 
   const [_s0] = $(__scope, 's0')
 
+  { const __l = []
   createEffect(() => {
     if (_s0) {
-      { const __v = `shrink-0 ${(_p.className ?? '')}`; if (__v != null) _s0.setAttribute('class', String(__v)); else _s0.removeAttribute('class') }
+      { const __x = `shrink-0 ${(_p.className ?? '')}`
+      if (!(0 in __l) || !Object.is(__l[0], __x)) {
+        { const __v = __x; if (__v != null) _s0.setAttribute('class', String(__v)); else _s0.removeAttribute('class') }
+      }
+      __l[0] = __x }
     }
-  })
+  }) }
 
   if (_s0) applyRestAttrs(_s0, _p, ["size","className","xmlns","viewBox","fill","stroke","stroke-width","stroke-linecap","stroke-linejoin","class","aria-hidden"])
 
@@ -144,11 +149,16 @@ export function initChevronDownIcon(__scope, _p = {}) {
 
   const [_s0] = $(__scope, 's0')
 
+  { const __l = []
   createEffect(() => {
     if (_s0) {
-      { const __v = `shrink-0 ${(_p.className ?? '')}`; if (__v != null) _s0.setAttribute('class', String(__v)); else _s0.removeAttribute('class') }
+      { const __x = `shrink-0 ${(_p.className ?? '')}`
+      if (!(0 in __l) || !Object.is(__l[0], __x)) {
+        { const __v = __x; if (__v != null) _s0.setAttribute('class', String(__v)); else _s0.removeAttribute('class') }
+      }
+      __l[0] = __x }
     }
-  })
+  }) }
 
   if (_s0) applyRestAttrs(_s0, _p, ["size","className","xmlns","viewBox","fill","stroke","stroke-width","stroke-linecap","stroke-linejoin","class","aria-hidden"])
 
@@ -249,11 +259,16 @@ export function initChevronUpIcon(__scope, _p = {}) {
 
   const [_s0] = $(__scope, 's0')
 
+  { const __l = []
   createEffect(() => {
     if (_s0) {
-      { const __v = `shrink-0 ${(_p.className ?? '')}`; if (__v != null) _s0.setAttribute('class', String(__v)); else _s0.removeAttribute('class') }
+      { const __x = `shrink-0 ${(_p.className ?? '')}`
+      if (!(0 in __l) || !Object.is(__l[0], __x)) {
+        { const __v = __x; if (__v != null) _s0.setAttribute('class', String(__v)); else _s0.removeAttribute('class') }
+      }
+      __l[0] = __x }
     }
-  })
+  }) }
 
   if (_s0) applyRestAttrs(_s0, _p, ["size","className","xmlns","viewBox","fill","stroke","stroke-width","stroke-linecap","stroke-linejoin","class","aria-hidden"])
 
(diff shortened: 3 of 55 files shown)


kfly8:
The `check` job failed because this PR touches published `@barefootjs/jsx` source with no changeset. Pushed `.changeset/attr-dedup-guard-2869.md` (patch) to fix it.

---

PR #2875, opened 2026-09-07
fix: re-run the .map() preamble inside loop-row branch arms

kfly8:
## Summary

Found while discussing (post-#2872) whether `insert.ts`'s branch-arm reactive-attr handling and `loop-child-arm.ts`'s `stringifyBranchReactiveAttrs` should share one implementation. Tracing the types turned up a real bug, not just an asymmetry:

`ReactiveAttrEffect` (the type `stringifyBranchReactiveAttrs` consumes) carries a `readsPreamble?: boolean` flag — set by `reactivity.ts`'s `collectLoopChildReactiveAttrs` (called for the branch-arm case with `stopAtReactiveConditionals: true`) for exactly the shape its own comment documents: `const cls = row.done ? … ; class={cls}`. But `stringifyLoopChildArm` / `stringifyBranchReactiveAttrs` never received a `mapPreambleWrapped` string to re-run before reading `cls` — unlike every other `readsPreamble` consumer (`emitAttrSlotsGranular`, `emitConsolidatedRowEffect`, and the outer conditional's own condition getter in `emitOuterConditional`).

**Confirmed via a happy-dom hydration probe**: a same-key row update refreshes a plain `{row().field}` text sibling correctly, but a class/attribute derived from a `.map()` preamble local inside the same row's branch conditional stayed frozen at whatever it was on the row's last full render — a silent divergence from a sibling binding reading the underlying data directly.

An audit (design by Fable, `claude-fable-5-1`) found the gap had **three** structural pieces, not one:
1. Plan layer: `buildArmAttrsPlan` dropped `readsPreamble` on the way from `LoopChildReactiveAttr` to the arm-attrs plan.
2. Stringify layer: `stringifyLoopChildArm` → `stringifyBranchReactiveAttrs` never threaded `mapPreambleWrapped` down to where `readsPreamble`-flagged attrs are emitted.
3. A **second, independent** instance of the same class of bug: `LoopChildConditionalPlan` (a *nested* conditional's own condition, one level inside a branch arm) was missing `readsPreamble` entirely — so a nested conditional's condition expression reading a preamble local had the identical staleness bug.

## What changed

- `LoopChildConditionalPlan` gains `readsPreamble?: boolean` (mirrors `NestedConditionalPlan.readsPreamble` one level up); `buildLoopChildConditionalsPlan`/`buildArmAttrsPlan` now forward the flag instead of dropping it.
- `stringifyLoopChildArm` / `stringifyBranchReactiveAttrs` / `stringifyLoopChildConditional` all gained a `mapPreambleWrapped: string | undefined` parameter, threaded through to where a `readsPreamble`-flagged attr effect or condition getter needs it.
- New shared helper `conditionGetterExpr()` in `loop-child-arm.ts`, used by both `emitOuterConditional` (outer, row-level conditional) and `stringifyLoopChildConditional` (nested, branch-scoped conditional) to build the `() => cond` getter, with or without a preamble re-run — so the two paths can't drift apart again on this logic (per the repo's "one decision, one implementation" convention).
- Left untouched, by design: branch-scoped inner-loop (`stringifyBranchInnerLoops`) and top-level inner-loop (`stringifyInnerLoops`) preambles — their preamble isn't a single re-runnable string like the outer row's, so there's nothing correct to thread there yet; both call sites now carry an explicit comment saying so.

**Relation to #2871** (already merged, in this PR's base): unrelated bug, same neighborhood. #2871 fixes `classifyReactivity` never recognizing an index-*only* expression as reactive at all (missing classification). This bug is different: the attribute/condition *is* correctly classified reactive, and does get an effect, but the effect never re-ran the preamble computation the expression depends on — a wiring gap downstream of correct classification.

## Test plan

- [x] `packages/client/__tests__/runtime/preamble-branch-attr-staleness.test.ts` — happy-dom hydration test; drives a same-key row update and asserts the class now updates (`'tier-silver'`), previously stuck at `'tier-gold'`.
- [x] `packages/jsx/src/__tests__/preamble-branch-arm-rerun.test.ts` — codegen-shape pins for (a) a reactive attribute inside a branch arm, (b) a nested conditional's own condition, and (c) a negative control (no preamble local in play — nothing injected).
- [x] `bun run packages/adapter-tests/scripts/generate-expected-html.ts` — zero diff; this is a client-JS-behavioral fix only, no SSR/CSR template output changes.
- [x] `.changeset/preamble-branch-arm-rerun-2596.md` added (`@barefootjs/jsx` patch).

https://claude.ai/code/session_01GGEdZGGtKh3hy1DfMcn59A

### DIFF
--- a/.changeset/preamble-branch-arm-rerun-2596.md
+++ b/.changeset/preamble-branch-arm-rerun-2596.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix a `.map()` callback preamble local (e.g. `const cls = row.done ? 'a' : 'b'`) going stale inside a loop row's branch conditional: a reactive attribute reading the local, or a nested conditional's own condition reading it, used to freeze at whatever value the local had on the row's last full render instead of refreshing on a same-key data update — silently diverging from a sibling binding that read the underlying data directly. Both are now fixed by re-running the row's preamble inside the branch's own effect / condition getter, matching how this was already handled everywhere else (`emitAttrSlotsGranular`, `emitConsolidatedRowEffect`, the outer conditional's own condition getter).
--- a/packages/client/__tests__/runtime/preamble-branch-attr-staleness.test.ts
+++ b/packages/client/__tests__/runtime/preamble-branch-attr-staleness.test.ts
@@ -1,0 +1,147 @@
+/**
+ * Regression probe, NOT yet filed as a GitHub issue: found while discussing
+ * whether `insert.ts`'s branch-arm reactive-attr handling and
+ * `loop-child-arm.ts`'s `stringifyBranchReactiveAttrs` should share one
+ * implementation (#2869 follow-up conversation). They turned out to answer
+ * DIFFERENT questions — `ReactiveAttrEffect` (the type `stringifyBranchReactiveAttrs`
+ * consumes) carries a `readsPreamble?: boolean` field that `ArmReactiveAttr`
+ * (`insert.ts`'s equivalent, for top-level non-loop conditionals) structurally
+ * cannot have, because there is no `.map()` preamble outside a loop.
+ *
+ * Tracing where `readsPreamble` is actually set (`reactivity.ts`'s
+ * `collectBranchReactiveAttrs`, guarding the exact "#2447" scenario this test
+ * reproduces: `const cls = row.done ? … ; class={cls}`) proved it CAN be true
+ * for an attribute inside a loop-row's branch conditional. But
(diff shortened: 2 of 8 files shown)


kfly8:
@pullfrog[bot] Thanks for the review. Updated the PR title and description to reflect the actual fix (previously stale from when this was test-only) — title is now "fix: re-run the .map() preamble inside loop-row branch arms" and the body describes the three structural pieces of the fix rather than posing it as an open question.

Re the nitpick: agreed the inner-loop preamble gap is now well-documented rather than merely suspected — will consider filing a `known-limitation` issue for it as a separate follow-up; it's intentionally out of scope for this PR.

---

Issue #2867, opened 2026-09-07
map-body-no-silent-divergence.test.ts: two subtests reported flaky under full-suite `bun test` (BF_ASSERT_NO_JSX_IN_GETJS leak?), pass in isolation


kfly8:
## What's happening

`packages/jsx/src/__tests__/map-body-no-silent-divergence.test.ts` — the "no-silent-divergence trichotomy" harness described in the repo's `CLAUDE.md` ("every `.map()` body shape must be sound-or-loud"; its `KNOWN_HOLES` set "is empty and may only shrink") — has two subtests that are **flaky specifically under a full `packages/jsx` suite run**, but pass cleanly when the file is run in isolation:

- `.map() body trichotomy — sound bundle XOR loud error, never silence > signal-array-builder (JS-runtime)`
- `.map() body trichotomy — sound bundle XOR loud error, never silence > flatmap-expression-body-signal-array (JS-runtime)`

Reported to me, and confirmed on a clean checkout of `main` at commit `8cff173` (zero code changes):

- `cd packages/jsx && bun test src/__tests__/map-body-no-silent-divergence.test.ts` → **29 pass, 0 fail**.
- `cd packages/jsx && bun test` (the full suite, all 265 test files) → **2 fail**, both inside this file, exactly the two subtests above.

Both fail with the same **uncaught exception** (not a normal `expect()` assertion failure):

```
error: getJS() called on a JSX-bearing node — raw JSX must never be spliced into emitted output. Carry mixed content as structured segments (MapCallbackPreamble / FlatMapCallback) instead.
      at getJS (/home/user/barefootjs/packages/jsx/src/analyzer-context.ts:329:9)
      at shouldAutoDeferReactiveBrand (/home/user/barefootjs/packages/jsx/src/jsx-to-ir.ts:8216:35)
      at transformExpressionInner (/home/user/barefootjs/packages/jsx/src/jsx-to-ir.ts:2432:26)
      at transformChildren (/home/user/barefootjs/packages/jsx/src/jsx-to-ir.ts:2244:25)
      at transformHtmlElement (/home/user/barefootjs/packages/jsx/src/jsx-to-ir.ts:1662:20)
      at buildIRRoot (/home/user/barefootjs/packages/jsx/src/jsx-to-ir.ts:1144:16)
      at jsxToIR (/home/user/barefootjs/packages/jsx/src/jsx-to-ir.ts:1085:16)
[...]

Issue #2861, opened 2026-09-06
A signal-free, function-call-free .map() index expression never gets a reactive effect at all


kfly8:
## What happened

Follow-up from #2859 (fixed in #2860), which made a keyed `.map()` row's index-derived output stay live across a same-key reorder — but only for expressions that already get a `createEffect` wired for some other reason (a signal read elsewhere in the expression, or a function call that trips the AST-flag fallback, e.g. `String(i + 1)`).

A row-index expression with **no signal read and no function call at all** — e.g. bare `{i}` text, or `class={i % 2 === 0 ? 'a' : 'b'}` — is still classified as fully static by Phase 2 reactivity analysis (`classifyReactivity` in `packages/jsx/src/ir-to-client-js/reactivity.ts`) and never gets a `createEffect` wired. It's baked once into the row's template at creation time and never updates — not even when the array itself grows and a brand-new row is created at a position derived from a *stale* baked value elsewhere, and certainly not across a same-key reorder.

This is distinct from #2859's root cause: Phase 1 (`jsx-to-ir.ts`'s `hasReactiveAttributes`/`referencesLoopParam`, via `BindingScope.valueBoundNames()`) already grants such an expression a slot (so it's *patchable*), but Phase 2 (`classifyReactivity`) has no equivalent to the `{ kind: 'loop-param' }` case it already has for the *item* parameter — there's no `{ kind: 'loop-index' }` — so no effect is emitted at all when nothing else in the expression is reactive.

## Reproduction

```tsx
"use client"
import { createSignal } from '@barefootjs/client'

export function Repro() {
  const [items, setItems] = createSignal([{ id: 'a' }, { id: 'b' }, { id: 'c' }])
  return (
    <ul>
      {items().map((item, i) => (
        <li key={item.id}>{i}</li>
      ))}
    </ul>
  )
}
```

After `setItems(([a, b, c]) => [b, c, a])` (same keys, reordered), every row's text should read its new position (`0, 1, 2`) but instead keeps showing whatever `i` it was created with.

This is confirmed as pre-existing, unchanged behavior — see `packages/jsx/src/__tests__/nested-loop-index-param.test.ts`'s `'index referenced in a template text interpolation is bound and rendered'` test, which pins the bare-`{i}` case as baked into the CSR clone template with no effect, both before and after #2860.

## Why it needs a separate fix

[...]

PR #2876, opened 2026-09-07
Fix #2868: reactive conditional branch no longer corrupts a colliding tag name

### LINKED ISSUE #2868: Reactive conditional branch template corrupts an `<i>` (or other 1-letter-tag) element when the loop declares a same-named index param
## Summary

A reactive conditional (`{cond ? <A/> : <B/>}`) inside a keyed `.map()` row whose `.map()` callback declares an index parameter named `i` (`.map((item, i) => ...)`) corrupts the **branch's own re-render template** when a branch's root element happens to be an `<i>` tag — the tag name gets mangled to `<i()>`. The condition itself does not need to reference `i` at all; the bug fires purely from the tag name and the index parameter's name colliding textually.

## Reproduction

```tsx
'use client'

export function Repro() {
  const [items] = createSignal([{ id: 'a' }, { id: 'b' }])
  const [flag] = createSignal(true)
  return (
    <ul>
      {items().map((item, i) => (
        <li key={item.id} ref={(el: HTMLElement | null) => {}}>
          {flag() ? <b>even</b> : <i>odd</i>}
        </li>
      ))}
    </ul>
  )
}
```

Compiled client JS (verbatim, relevant excerpt):

```js
insert(__el, 's0', () => flag(), {
  template: () => { const __slots = []; return { html: `<b bf-c="s0">even</b>`, slots: __slots } },
  bindEvents: (__branchScope, { isFirstRun: __bfFirstRun = false } = {}) => {}
}, {
  template: () => { const __slots = []; return { html: `<!--bf-cond-start:s0--><i()>odd</i><!--bf-cond-end:s0-->`, slots: __slots } },
  bindEvents: (__branchScope, { isFirstRun: __bfFirstRun = false } = {}) => {}
})
```

The `false` branch's own live re-render template has `<i>` rewritten to `<i()>` — broken HTML that would fail to parse as the intended element when `insert()` re-renders this branch at runtime. Note:

- The condition (`flag()`) never references `i`.
- The **initial** SSR/clone template (`__tpl.innerHTML = ...<i bf-c="s0">odd</i>...` earlier in the same `renderItem`) and the top-level `hydrate(...)` template call are both **correct** — only the branch arm's own `template()` builder (used when `insert()` later swaps the branch) is affected.
- Confirmed present on `main` — this is not caused by any in-flight change; it reproduces with a plain signal-driven conditional and requires nothing beyond an index-named-`i` `.map()` and an `<i>` (or presumably any other short tag name that happens to match the index/item parameter identifier) branch root.

## Likely root cause

Whatever builds the branch arm's own template string appears to apply a word-boundary regex-based identifier rewrite (`wrapLoopParamAsAccessor` or a sibling helper, used elsewhere to turn a bare loop parameter reference like `i` into an accessor call `i()`) directly over the **rendered HTML string** for this branch, rather than only over parsed JS expression positions. A word-boundary regex (`\bi\b`) matches the `i` inside `<i>...</i>` just as readily as a real identifier reference, because `<` and `>` are non-word characters. Any short tag name coinciding with a loop's item/index parameter name (or a destructured binding name) is presumably at risk — `<i>`, and potentially single-letter or otherwise-colliding custom element names.

## Why this needs its own fix

This is a **silent divergence**: the branch compiles clean, but the emitted re-render template for that branch is broken markup, not a refusal — bit-for-bit different from the intended output and unrelated to whether the loop index is ever referenced by that branch. It's a distinct, more general bug than #2859/#2860/#2861 (the index-reactivity-freezing family) — it isn't about reactivity at all, it's a raw text-substitution bug that just happens to also involve a loop's index/item parameter name.

## Suggested fix direction

Whatever emits the branch's `template()` HTML for a reactive conditional inside a loop row should not run identifier-rewriting regexes over the already-rendered HTML string. Per this repo's own architecture rule (`CLAUDE.md`, "never splice raw source text into emitted artifacts untyped"), the loop-param accessor rewrite for a branch's dynamic parts should happen on the **JS expression** positions (attribute/text bindings, event handlers) captured structurally during IR→template lowering — not via a blanket string-level regex pass over the finished HTML markup that can also match plain tag names.

## Scope note

Not fixed as part of #2861 (loop index reactivity) — found incidentally while writing that fix's tests (an index-only reactive conditional with an `<i>` branch triggered this and was mistaken at first for a regression, until reproduced independently on `main` with a plain signal-based conditional with no index involvement at all).

kfly8:
Fixes #2868.

## Root cause

A reactive conditional branch (`{cond ? <A/> : <B/>}`) inside a keyed `.map()` row's own re-render `template()` builder was built by re-wrapping the branch's already-**assembled HTML string** with `wrapLoopParamAsAccessor` — a word-boundary regex meant to turn a bare loop-param/index reference into an accessor call (`i` → `i()`). That regex can't tell a real identifier reference apart from a tag name spelled the same way: `<i>odd</i>` became `<i()>odd</i>`, broken markup, whenever a `.map()` declared an item or index parameter matching a branch's own root tag name.

## Fix

Every loop-param-wrapped HTML string is now rendered against the full loop-param chain **structurally, at IR-render time** (`irToHtmlTemplate`'s `loopParams`), instead of being wrapped a second time afterward as a blanket string-level pass — exactly the architecture `CLAUDE.md` already calls for ("never splice raw source text into emitted artifacts untyped").

- Removed the post-hoc `wrapLoopParamAsAccessor(cond.whenTrueHtml/whenFalseHtml, ...)` calls in `build-reactive-effects.ts`, `build-loop-child-arm.ts`, and `build-lazy-row.ts` — the HTML is already correctly wrapped by the time it reaches these builders.
- `collectLoopChildConditionals` (`collect-elements.ts`) gained an optional `loopParams` parameter carrying the **full ancestor loop-param chain**, not just the immediate loop's own — fixing an adjacent gap where a conditional inside an inner `.map()` couldn't reference an *outer* loop's item/index in its branch HTML at all (it silently stayed a bare, un-called accessor reference).
- `build-plain-row.ts`'s eager row template gets the same treatment via a new `TopLevelLoop`/`BranchLoop.templateIndexed` field — a second structured render with the index wrapped, computed alongside the existing item-only `template` in `collect-elements.ts` — replacing its own post-hoc `wrapIndexParamAsAccessor(loop.template, loop.index)` call. (The lazy row plan still needs the index-*unwrapped* `template`/`mapPreambleWrapped`, since `mapArrayLazy` hands `createRow` a plain number, never an accessor — unchanged.)

## Tests

- New `packages/jsx/src/__tests__/issue-2868-cond-arm-tag-collision.test.ts`: 5 compiler-unit tests covering index-as-tag (eager, forced via `ref`), item-as-tag, a nested conditional inside a branch arm, the row-template (Step C) path, and a nested `.map()` conditional arm reading an *outer* loop's own index (confirming it's correctly called as `gi()`, not passed as a bare function reference — a real, distinct bug caught while writing this fix).
- New real-browser conformance fixture `cond-arm-tag-collision` (`CondArmTagCollision.tsx`): an eager row (`<b>`/`<i>` branches, item param literally named `b`) and a lazy row, toggled via a button — verified across all adapters and Playwright-snapshotted.
- `binding-scope-ratchet.test.ts`'s shrink-only ledger updated: this fix legitimately extends the already-established, permanently-allowlisted "shape 3" accessor-rewrite-payload mechanism (`loopParams: ReadonlyArray<string | LoopParamSpec>`) to a few more call sites — not a new ad-hoc scope device.

## Verification

- Confirmed the exact issue repro (`<i>` inside a branch, `.map((item, i) => ...)`) corrupts on pre-fix code and compiles clean after, via a standalone script.
- Confirmed 3 additional variant repros (item-as-tag, row-template index-as-tag, nested-loop outer-index arm) all reproduce pre-fix and are fixed.
- Full `packages/jsx` suite: 3342 pass, 0 fail (the only 2 failures are the pre-existing `map-body-no-silent-divergence.test.ts` flake tracked in #2867/#2830, already fixed on a separate PR not yet merged into this branch's base).
- `bun test packages/compat`: 524 pass, 0 fail (`ui/compat.lock.json`/`ui/support-matrix.lock.json` regenerated for the new fixture).
- `bunx tsgo --noEmit`: clean.

---


kfly8:
Pushed `ab7036e`, on top of `972a65a` (merge of `main`, needed since `main` moved past #2872's attr-dedup-guard after this branch was created).

`ab7036e` fixes a **real regression this PR's own main fix introduced**, caught by `site/ui`'s `DashboardBuilderDemo` e2e suite (`e2e-site-ui`), not by any fixture in this PR: removing the post-hoc `wrapLoopParamAsAccessor` pass over branch-arm HTML (this PR's fix for the `<i>` tag-name-collision bug) also removed a pass that had been accidentally papering over a separate, pre-existing gap — `irToHtmlTemplate`'s `case 'component':` branch never wrapped a child component's prop expressions through `wrapExpr` at all. A child component rendered inside a reactive conditional branch, with a prop reading the row's own loop item/index (e.g. `{item.active ? <Badge text={item.label} /> : null}`), emitted a `renderChild(...)` call reading the bare loop param instead of its live accessor — so the prop froze at its row-creation-time value instead of updating.

Fixed by wrapping that expression the same way every other JS-expression position in `irToHtmlTemplate` already does. Added a compiler-unit regression test (verified: fails without the fix, passes with it) and updated the changeset.

Verification on the new head:
- Full `packages/jsx` + `packages/adapter-tests` suite: 5765 pass, 0 fail.
- `ui/compat.lock.json` / `ui/support-matrix.lock.json`: no drift.
- `bunx tsgo --noEmit`: clean.
- `site/ui/e2e/dashboard-builder.spec.ts`: 23/23 passing (was 15/23 before this fix).

Will keep watching CI on the new head.

---

kfly8:
Marked ready for review. Also resolved a real merge conflict along the way (this branch and #2874 had independently extended the same functions — `collectInnerLoops` and `collectLoopChildConditionals` in `collect-elements.ts`, and the conditional-arm plan builder in `build-reactive-effects.ts` — for different reasons):

- `collectInnerLoops`'s `outerLoopParam` type (`string | LoopParamSpec`, from this PR) and #2861's new `outerLoopParamBindings`/`outerLoopIndex` trailing params auto-merged into a body that silently wrapped an already-a-`LoopParamSpec` value inside another spec object. Fixed by normalizing `outerLoopParam` into one `LoopParamSpec` shape up front and dropping the now-redundant trailing params — both callers already pass a full spec.
- `collectLoopChildConditionals` gained two independent new trailing params from the two PRs (this PR's `loopParams`, #2861's `parentScope`) — kept both, they serve different concerns (template-rendering ancestor chain vs. reactivity-scope stacking).
- `build-reactive-effects.ts`'s conditional-arm plan builder: kept this PR's "don't re-wrap `whenTrueHtml`/`whenFalseHtml`" fix while also picking up #2861's new `loopIndex` argument to `buildOuterArm` (dropping it would have shifted every argument after it by one position).

Verified: `bunx tsgo --noEmit` clean, full `packages/jsx` + `packages/adapter-tests` suite (5786 pass / 2 fail, both fixed — a stale `coverage-map.json` and a shrink-only ratchet allowlist count that legitimately grew from the merge), `ui/compat.lock.json`/`ui/support-matrix.lock.json` regenerated with no unexpected drift.

---

kfly8:
Fixed the `update-expected-html` CI failure in `ee54661`: `cond-arm-tag-collision`'s committed snapshot predated main's `createDisposableEffect`/`__disposers` cleanup wiring for branch-arm reactive text effects — pure snapshot regen (`generate-expected-html.ts`), no code change. Verified `git status --short` was empty after regen except that one file, and the fixture's own test (`cond-arm-tag-collision`) passes.

---

Issue #2335, opened 2026-07-19
Go adapter: unify conditional lowering on pipeline bf_ternary — deletes the fragment special-casing, and fixes a silent branch-discard in condition position


kfly8:
Follow-up to #2331, which added `bf_ternary` (pipeline-position conditional, `runtime/bf.go`) for helper-call args because Go `text/template` is the only backend language without an expression-level conditional. The rest of the Go adapter still works around that gap with `{{if}}…{{else}}…{{end}}` **action fragments** plus guards to avoid double-wrapping them. A survey of `packages/adapter-go-template/src/` found one high-value refactor, one latent correctness bug, one low-value cleanup — and three areas that look like candidates but are **not** (documented here so nobody re-investigates).

## 1. HIGH — `conditional()` emits fragments; five guard sites exist only to accommodate that

- `go-template-adapter.ts:3792-3802` — the ParsedExpr `conditional()` emitter produces `{{if t}}c{{else}}a{{end}}` fragments (text-position only).
- `go-template-adapter.ts:4770-4781` — `renderConditionalBranch`, used ONLY by `conditional()` (bare-text string branches, nested `{{if}}` chains).
- `go-template-adapter.ts:3186-3188` — `isTemplateFragment` (`{{`-leading OR template-literal), consumed at `:3118` (renderExpression), `:3815` (templateLiteral), `:6106`/`:6144` (attr paths), `:6331`/`:6337` (substituteJsInterpolations) purely to avoid re-wrapping conditional fragments.

If `conditional()` instead emits `(bf_ternary <test> <c> <a>)` — exactly the form `renderLoweringNode`'s `lowerArg` already produces (`expr/url-builder.ts:123-129`) — conditionals stop being `{{`-leading: `renderConditionalBranch` dies, `isTemplateFragment` collapses to `kind === 'template-literal'`, the five guard sites lose their conditional case, and the two parallel conditional-lowering paths unify.

**Not a free swap — two gates:**
- **Branch quoting**: the fragment path renders string branches as bare unquoted text; `bf_ternary` args must be quoted values (as `lowerArg` does). Behavior for string branches moves from raw-text to escaped-value — safe for the value-position shapes `conditional()` is reached for, and precisely why this must NOT be extended to the text-position `IRConditional` path (`renderConditional`, `:5065`), whose branches are arbitrary markup.
[...]

Issue #2236, opened 2026-07-12
Go: convertExpressionToGo's bare-identifier fast path inlines a local const at a loop-shadowed name (sibling of #2221)


kfly8:
Found during the #2221 sweep (PR #2234). The Go template adapter has its own independent instance of the #2221 bug class.

## Summary

`convertExpressionToGo`'s bare-identifier fast path (`packages/adapter-go-template/src/adapter/go-template-adapter.ts`, around lines 4541-4548) resolves a bare identifier against local constants with **no loop-shadowing guard** — unlike its sibling paths `resolveModuleStringConst` / `resolveModuleNumericConst` (~lines 3190-3223), which already check `loopParamStack` / `loopVarRefCount` / `isOuterLoopParam`.

## Repro (verified)

```tsx
const count = 7
return <ul>{[2, 5].map((count) => <li key={count}>{count}</li>)}</ul>
```

compiles to `data-key="{{7}}"` — the outer const's literal — instead of the loop variable, for every iteration.

## Fix direction

Apply the same shadow guards the sibling resolvers already use (`loopParamStack` et al.) to the bare-identifier fast path. Note part of the observed output may also come from the shared IR-level attribute fold (#2235) — verify both layers with the repro after fixing either.

## Relationship

- #2221 (PR #2234): the Twig-family equivalent (fixed there; Go deliberately out of scope as its architecture differs).
- #2235: shared jsx-to-ir attribute fold, which can mask/compound this.

PR #2877, opened 2026-09-07
Fix #2863: Go adapter ternary/operand template-literal branch emits invalid syntax

### LINKED ISSUE #2863: Go Template: conditional template literal returned by a local helper emits invalid bf_ternary syntax
## What happened?

When a local helper used inside a keyed `.map()` returns a conditional expression whose consequent is a template literal, the Go Template adapter emits nested `{{ ... }}` actions as an operand of `bf_ternary`.

The compiler/build completes successfully, but the generated template is invalid and `html/template` fails to parse it at application startup with `unexpected "{" in operand`.

This still reproduces on the current `main` branch at commit `bccca0481f835a359a52ce27a62092bf0aa9c4f3`.

## Component source that reproduces it (if handy)

```tsx
'use client'

type Event = {
  id: string
  time: string
  endTime: string
}

type Dashboard = {
  events: Event[]
}

const empty: Dashboard = { events: [] }

export function Schedule() {
  const [data] = createSignal<Dashboard>(empty)

  const eventTime = (event: Event) =>
    event.endTime ? `${event.time}–${event.endTime}` : event.time

  return (
    <ol>
      {data().events.map((event) => (
        <li key={event.id}>
          <time>{eventTime(event)}</time>
        </li>
      ))}
    </ol>
  )
}
```

## Expected vs. actual (if known)

Expected:

The template-literal branch should be lowered to one valid Go-template operand, for example using `printf`, or an equivalent supported expression:

```gotemplate
{{(bf_ternary (bf_truthy .EndTime) (printf "%s–%s" .Time .EndTime) .Time)}}
```

Actual:

```gotemplate
{{(bf_ternary (bf_truthy .EndTime) {{.Time}}–{{.EndTime}} .Time)}}
```

The nested actions make the generated template invalid. Parsing it with Go's `html/template` produces an error similar to:

```text
template: Schedule.tmpl:...: unexpected "{" in operand
```

The build itself exits successfully, so the problem is only detected when the Go application loads the generated templates.

## Environment (if relevant)

- BarefootJS version: `@barefootjs/go-template 0.34.0`, `@barefootjs/jsx 0.34.0`
- Also reproduced on latest main: `bccca0481f835a359a52ce27a62092bf0aa9c4f3`
- Adapter: Go Template
- Runtime: Bun 1.4.0 for the upstream reproduction
- Go: 1.27.1
- OS: macOS arm64

kfly8:
Fixes #2863.

## Root cause

A ternary (or a `+`/`||`/`??` operand, or a `queryHref`-style helper argument) whose branch was a **multi-part** template literal (mixed literal text and 2+ interpolations, e.g. `` event.endTime ? `${event.time}–${event.endTime}` : event.time ``) lowered through the Go adapter's TEXT-position `templateLiteral()` emitter — mixed literal text plus `{{…}}` action wraps, meant for direct markup embedding. That output was then used as-is inside another action's pipeline argument list:

```
{{(bf_ternary (bf_truthy .EndTime) {{.Time}}–{{.EndTime}} .Time)}}
```

`bf build` succeeds, but the generated `.tmpl` file fails `html/template.Parse` at Go application startup with `unexpected "{" in operand` — nested `{{`/`}}` delimiters aren't legal inside another action's operand list.

The bug is broader than the reported ternary shape: the same leak affects a `+`/`||`/`??` operand and a registered-lowering (`queryHref`) argument that is a template literal, and a pre-existing (already-guarded-against-single-part-only) gap at the per-row child-component prop site (#2445).

## Fix

- New `lowerTemplateLiteralValue` + `lowerValueOperand` (`expr/url-builder.ts`) — the shared door for "this `ParsedExpr` sits in Go pipeline ARGUMENT position." A template literal now folds to a left-folded `bf_concat_str` chain (the same runtime helper already used for JS string-concatenation `+`, #2168) instead of leaking the TEXT-position form. Wired into `lowerTernary`'s branches/test, `lowerUrlGuard`, and `renderLoweringNode`'s helper-call/guard-list arguments.
- `binary()`/`logical()`/`unary()` only special-case the `template-literal` kind (via a narrow `emitOperand` wrapper) and otherwise keep using the existing `emit()` dispatcher unchanged — routing every operand through the generic value path would also pick up `convertExpressionToGo`'s own bare-identifier-const-inlining fast path, a real behavioral divergence the existing `#2236` regression pins caught during verification (see below).
- The per-row child-component prop site (#2445) is unified onto the same door: a multi-part template-literal prop, previously refused outright with `BF101`, now lowers faithfully instead of being refused.

## Tests

- `packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts`: new tests in the `#2335` `bf_ternary value-position lowering` describe block (multi-part branch, single-part reduction, static-text escaping, nested ternary inside a branch, the issue's exact local-helper-in-`.map()` repro) plus a new `#2863` describe block for `+`/`||` operands. Updated the pre-existing `#2445` test to assert the new (correct) `bf_concat_str` lowering instead of the old BF101 refusal.
- New real-backend cross-adapter conformance fixture `packages/adapter-tests/fixtures/ternary-template-literal-branch.ts` (attribute-position ternary+template-literal, with an HTML-special-character prop to pin escaping parity, plus an alternate-branch data point).

## Verification

- `bunx tsgo --noEmit`: clean.
- `bun test packages/adapter-go-template`: 2002 pass, 0 fail.
- `bun test packages/adapter-tests`: green, including the new fixture.
- `bun test packages/compat` (`support-matrix.test.ts` / `coverage-map.test.ts`): green after regenerating `ui/compat.lock.json`, `ui/support-matrix.lock.json`, and `packages/adapter-tests/coverage-map.json`.
- **Real Go execution** (not just string assertions): confirmed with `go1.25.6` (`GOTOOLCHAIN=go1.25.6`) that the fixed template actually parses via `html/template.Parse` and executes correctly against the real runtime `FuncMap()`, including an HTML-escaping-parity check. (Note for reviewers: this sandbox's default `go` resolves to 1.24.7, and `test-render.ts`'s `isGoAvailable()` check forces `GOTOOLCHAIN=local`, so real Go rendering is silently skipped locally unless `GOTOOLCHAIN` is pinned to `go1.25.6\` as above — CI's `setup-go` uses 1.25 so this isn't an issue there.)

## Process note

Design for this fix was delegated to a Fable agent given the root-cause investigation had already been done; I implemented and verified its recommendation directly, including catching and correcting a real regression its recommended `binary`/`logical`/`unary` refactor would have introduced (surfaced by the pre-existing `#2236` loop-param-shadowing regression suite, not by the design review itself).

---

### DIFF
--- a/.changeset/go-ternary-template-literal-2863.md
+++ b/.changeset/go-ternary-template-literal-2863.md
@@ -1,0 +1,7 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Fix #2863: a ternary (or `+`/`||`/`??` operand, or a `queryHref`-style helper argument) whose branch was a multi-part template literal (e.g. `` event.endTime ? `${event.time}–${event.endTime}` : event.time ``) emitted invalid Go template syntax — nested `{{…}}` action delimiters inside another action's pipeline argument list (`(bf_ternary … {{.Time}}–{{.EndTime}} …)`). `bf build` succeeded, but the generated `.tmpl` file failed `html/template.Parse` at Go application startup with `unexpected "{" in operand`.
+
+The Go adapter's ternary/operand lowering now folds a template literal used in this VALUE position into a single Go pipeline value via a left-folded `bf_concat_str` chain (the same runtime helper already used for JS string-concatenation `+`), instead of routing it through the TEXT-position `templateLiteral()` emitter (mixed literal text plus `{{…}}` action wraps), which is only safe when spliced directly into markup. A per-row child-component prop reading a multi-part template literal — previously refused outright with `BF101` — now lowers faithfully through the same path instead.
--- a/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
+++ b/packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts
@@ -376,6 +376,101 @@ describe('GoTemplateAdapter - bf_ternary value-position lowering (#2335)', () =>
   test('emits no {{if}} action fragment for a value-position ternary', () => {
     expect(render("flag ? 'a' : 'b'")).not.toContain('{{if')
   })
+
+  // #2863: a multi-part template literal (mixed literal text and MULTIPLE
+  // interpolations) as a ternary BRANCH used to lower through the TEXT-
+  // position `templateLiteral()` path (`{{.Start}}-{{.End}}`), nesting raw
+  // `{{`/`}}` delimiters inside `bf_ternary`'s own argument list — a
+  // `html/template` parse error ("unexpected \"{\" in operand") at Go
+  // application startup, even though `bf build` itself succeeded. It must
+  // instead fold to one pipeline value, chained through the same
+  // `bf_concat_str` runtime helper JS string-concat `+` already uses (#2168).
+  test('a multi-part template-literal branch folds to a bf_concat_str chain, not a raw fragment (#2863)', () => {
+    expect(render('end ? `${start}-${end}` : start')).toBe(
+      '{{(bf_ternary (bf_truthy .End) (bf_concat_str (bf_concat_str .Start "-") .End) .Start)}}',
+    )
+  })
+
+  test('a template-literal branch never leaks a raw {{ or }} into the surrounding action (#2863)', () => {
+    const out = render('end ? `${start}-${end}` : start')
+    const inner = out.slice(2, -2) // strip the outer {{ }} the whole expression is wrapped in
+    expect(inner).not.toContain('{{')
+    expect(inner).not.toContain('}}')
+  })
+
+  test('a single-part template-literal branch reduces to the bare value, no bf_concat_str wrap (#2863)', () => {
+    expect(render('end ? `${start}` : end')).toBe('{{(bf_ternary (bf_truthy .End) .Start .End)}}')
+  })
+
+  test('static text in a template-literal branch is an escaped Go string literal (#2863)', () => {
+    expect(render('end ? `say "hi" ${start}` : end')).toBe(
+      '{{(bf_ternary (bf_truthy .End) (bf_concat_str "say \\"hi\\" " .Start) .End)}}',
+    )
+  })
+
+  test('a nested ternary inside a template-literal branch still lowers to bf_ternary, not {{if}} (#2863)', () => {
+    const out = render('end ? `x ${flag ? \'a\' : \'b\'}` : end')
+    expect(out).toBe(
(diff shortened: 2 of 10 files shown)


kfly8:
Fixed the `test (rest)` CI failure in `082bc06`: the new `ternary-template-literal-branch` fixture's prop types weren't yet reflected in the committed `packages/adapter-tests/generated-data-points.json` (the auto-generated adversarial-data-points catalog), failing `generated-data-points.test.ts`'s freshness check. Regenerated via `bun packages/adapter-tests/scripts/generate-data-points.ts` — purely additive (58 new lines for the new fixture's generated points: empty/markup/multibyte variants of `a`/`b`, and the `ok:false` branch).

Verified: `generated-data-points.test.ts` passes, and re-ran the full `packages/adapter-tests` suite (2433 pass, 0 fail) to confirm none of the newly generated adversarial points (markup/multibyte/empty-string variants) surface a new divergence on any adapter.

---

kfly8:
Good catch — verified all three by actually reproducing each with `compileJSX` before fixing (confirmed the identical invalid nested `{{...}}` output), then fixed and re-verified, including real Go execution (`html/template.Parse` + `Execute` against the real `FuncMap()` via a throwaway `go test`). Went with option (a): closing the bug class for real rather than filing a follow-up.

Fixed in `a372cd5`:
- **`renderConditionExpr`'s `template-literal` case** — this is a separate walker from the generic `emit()` dispatcher (it threads its own `preamble` string), so I mirrored the adjacent `conditional` case's existing pattern (which already delegates to `lowerTernary` the same way) rather than routing through `lowerValueOperand` directly — keeps preamble handling consistent with that established precedent.
- **`arrayMethod()`** — rather than patching only `join`/`includes`/`indexOf`/`lastIndexOf`/`at`/`concat`/`slice` as named, I shadowed the whole function's `emit` parameter with `emitOperand` once at the top, closing every argument position in the ~30-case switch at once (`toFixed`, `split`, `startsWith`/`endsWith`, `replace`, `padStart`/`padEnd`, etc. all shared the identical unfixed pattern, not just the ones named).
- **`indexAccess()`** — both the object and the index operand now route through `emitOperand`.

Also confirmed `renderFilterExpr` is NOT affected, as you noted — no `template-literal`/`conditional` case there, so it structurally refuses instead.

New regression tests for all three in `go-template-adapter.test.ts`. Full `packages/adapter-go-template` suite: 2012 pass, 0 fail. `ui/compat.lock.json`: no drift.

---

PR #2890, opened 2026-09-08
Fix #2883: Mojolicious filter predicate against a renamed prop uses numeric comparison

### LINKED ISSUE #2883: mojolicious: MojoFilterEmitter.identifier() doesn't resolve a renamed prop destructure, unlike the top-level splice path
## The gap

`MojoFilterEmitter.identifier()` (`packages/adapter-mojolicious/src/adapter/expr/emitters.ts`) resolves an identifier reaching a `.filter()`/`.find()`/etc. predicate against only two things:

```ts
identifier(name: string): string {
  if (name === this.param) return `$${this.param}`
  const signal = this.localVarMap.get(name)
  if (signal) return `$${signal}`
  return `$${name}`
}
```

Anything else falls through to a bare `$<name>` — a Perl variable spelled exactly as the local JS name. For a bare-props-form body-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) referenced directly inside a filter predicate, this emits `$skipLabel` — a Perl variable that was never declared anywhere in the generated `.html.ep` (only `$fallbackLabel`, the caller-facing name, is seeded/declared). Real Mojolicious rendering evaluates the filter guard against this undeclared variable, and every row is filtered out (the generated `<ul>` renders empty) instead of the correct subset.

This is the same alias family #2788 originally covered (a renamed body-destructured prop not reaching its correct Perl variable) — but #2788's fix apparently only covers the main stash-interpolation/splice path (`packages/adapter-mojolicious/src/render-divergences.ts` is currently empty, so that shape is fixed), not `MojoFilterEmitter`'s own separate `identifier()` implementation, which was never taught the same alias resolution.

## Repro

```tsx
'use client'

type Todo = { id: number; label: string }

export function FilterPredicateRenamedProp(props: { fallbackLabel: string }) {
  const { fallbackLabel: skipLabel } = props
  const [todos] = createSignal<Todo[]>([
    { id: 1, label: 'Alpha' },
    { id: 2, label: 'Beta' },
  ])
  return (
    <ul>
      {todos().filter(t => t.label !== skipLabel).map(t => (
        <li key={t.id}>{t.label}</li>
      ))}
    </ul>
  )
}
```
With `fallbackLabel: 'Alpha'`, Hono (and every other adapter) renders only the `Beta` row. Real Mojolicious rendering produces an empty `<ul></ul>` — every row filtered out.

## How this was found

Surfaced by the `filter-predicate-renamed-prop` conformance fixture added for #2857's fix (a Go Template adapter bug in the analogous `renderFilterExprNode` alias-resolution path). That fixture is a shared cross-adapter fixture and is not Go-specific, so it also exercises Mojolicious — where it fails for this unrelated, pre-existing reason. Pinned as a render divergence (`packages/adapter-mojolicious/src/render-divergences.ts`) rather than fixed inline, since Mojo's adapter code is out of scope for #2857's Go-only fix.

## Suggested fix

Teach `MojoFilterEmitter.identifier()` the same alias resolution the top-level path already has for renamed prop destructures (and ideally the #2813 getter-alias family too, for parity — though the `filter-predicate-signal-alias` fixture's getter-alias case already passes on Mojo today, so that family may already be covered via `localVarMap` seeding upstream). A conformance-fixture-driven fix (the existing `filter-predicate-renamed-prop` fixture, unpinned) should land alongside it, per this repo's fixture-before-fix convention.

Tier: `+ bug` (silent divergence — real rows silently vanish instead of a documented refusal).

kfly8:
## Summary

Stacked on #2889 (independent fix, same bug-ticket sweep).

A `.filter()`/`.find()` predicate comparing against a bare-props-form BODY-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) lowered to Perl's numeric `!=`/`==` instead of string `ne`/`eq`. `'Alpha' != 'Alpha'` and `'Beta' != 'Alpha'` both numify to `0 != 0`, so every row was silently filtered out regardless of the real string value.

The issue's own diagnosis pointed at `MojoFilterEmitter.identifier()` never resolving the alias — investigation showed that's actually already correct (Mojo's stash derivation resolves `$skipLabel` for both emitters). The real gap: `collectStringValueNames`'s string-typed-name witness (which decides `eq`/`ne` vs `==`/`!=`) had no notion of this alias family at all, so the comparison fell through to numeric regardless of which variable name it saw.

`collectStringValueNames` now also resolves this alias family via the shared `resolveBodyDestructuredPropAliases` helper (already used by the #2788 top-level splice path), so both `MojoTopLevelEmitter` and `MojoFilterEmitter` read the same, now-correct witness. Un-pins the existing `filter-predicate-renamed-prop` conformance fixture's Mojo render divergence (`render-divergences.ts`).

Fixes https://github.com/piconic-ai/barefootjs/issues/2883.

## Test plan

- [x] `bun test packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts -t 2883` — new regression test asserting `ne`/not `!=`.
- [x] `bun test packages/adapter-mojolicious` — 1871 pass / 40 skip (real Perl execution skipped, no Mojolicious in this sandbox) / 1 fail + 1 error, both confirmed pre-existing and unrelated (`vite.test.ts`'s `@barefootjs/vite` module-resolution gap, reproduces identically on unmodified `main`).
- [x] `bunx tsgo --noEmit -p packages/adapter-mojolicious/tsconfig.json` — clean.

---

### DIFF
--- a/.changeset/mojo-filter-renamed-prop-2883.md
+++ b/.changeset/mojo-filter-renamed-prop-2883.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/mojolicious": patch
+---
+
+Fix #2883: a `.filter()`/`.find()` predicate comparing against a bare-props-form BODY-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) lowered to Perl's numeric `!=`/`==` instead of string `ne`/`eq`, because `collectStringValueNames`'s string-typed-name witness had no notion of this alias family — only `MojoTopLevelEmitter`'s stash-derived `$skipLabel` variable (correctly resolved, contrary to this issue's original diagnosis of `MojoFilterEmitter.identifier()`) ever reached the comparison. `'Alpha' != 'Alpha'` and `'Beta' != 'Alpha'` both numify to `0 != 0`, so every row was silently filtered out regardless of the real string value. `collectStringValueNames` now also resolves this alias family via the shared `resolveBodyDestructuredPropAliases` helper (already used by the #2788 top-level splice path), so both emitters read the same, now-correct witness.
--- a/packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts
+++ b/packages/adapter-mojolicious/src/__tests__/mojo-adapter.test.ts
@@ -733,6 +733,40 @@ export function L() {
     expect(result.template).not.toContain('$sel eq $t->{n}')
   })
 
+  test('compares a filter predicate against a renamed body-destructured prop with Perl `ne`, not numeric `!=` (#2883)', () => {
+    // `MojoFilterEmitter.identifier()` correctly resolves `skipLabel` to
+    // `$skipLabel` (Mojo seeds/derives that stash key for the #2788 alias
+    // family, same as the top-level splice path) — the actual bug was
+    // `collectStringValueNames`'s witness for `eq`/`ne` vs `==`/`!=` not
+    // knowing about body-destructured RENAMED props, so this comparison
+    // fell through to numeric `!=`. `'Alpha' != 'Alpha'` and
+    // `'Beta' != 'Alpha'` both numify to `0 != 0` — every row was
+    // filtered out regardless of the real string value.
+    const result = compileAndGenerate(`
+"use client"
+import { createSignal } from "@barefootjs/client"
+
+type Todo = { id: number; label: string }
+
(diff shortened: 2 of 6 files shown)


kfly8:
Thanks for the review. Filed the go-template/erb sibling gap as a follow-up: https://github.com/piconic-ai/barefootjs/issues/2894 — agreed it's out of scope for this PR (pre-existing, untouched by this diff).

---

PR #2891, opened 2026-09-08
Fix #2798: nested .map() over plain elements inside a static outer row is unwired

### LINKED ISSUE #2798: A `ref` callback on an element inside a static (non-signal) array's `.map()` is never called
Found while building the #2750 regression fixture (`packages/adapter-tests/fixtures/nested-loop-ref-const.ts`) — a `code-review` pass on that PR caught that the fixture, as first written, used a plain literal array (`const items = [...]`) for the looped data, and empirically that fixture passed the scope gate whether or not #2750's fix was present. Investigating why: a literal (non-signal) array routes through the hoisted static-array fast path (`collect-elements.ts`'s `isStaticArray` branch), which pre-renders the loop to a static HTML template and never splices in a ref call site at all — `emitRefCall` is only wired up by the dynamic `mapArray`/row-construction codegen paths (`stringify/loop.ts` / `stringify/insert.ts` / `phases/ref-callbacks.ts`), none of which run for the static fast path.

## Repro

```tsx
'use client'
export function StaticRefCase() {
  const items = [{ id: 1 }, { id: 2 }]
  const trackMount = (el: Element) => {
    el.setAttribute('data-tracked', '1')
  }
  return (
    <ul>
      {items.map(item => (
        <li key={item.id} ref={trackMount}>{item.id}</li>
      ))}
    </ul>
  )
}
```

`trackMount` is declared (no `ReferenceError` — this is NOT #2750's dead-declaration bug, the const *is* reachable and emitted), but it is never called: the static path renders `<li>` markup once from the literal array and moves on, with no per-row `ref` invocation at any point in the emitted `initX`.

Not depth-dependent — this happens at the TOP level of a static-array loop too, not just when nested.

## Why this was invisible until now

No existing fixture pairs a `ref` on a loop row with a literal (non-signal) array — every existing `ref`-in-loop fixture/case uses a signal-backed array, which takes the dynamic path and correctly invokes the ref. A user writing a static list with a `ref` today gets no error and no ref invocation — a silent capability gap, not a loud refusal.

## Follow-up

Needs the three-piece set once triaged: this issue, a conformance/CSR fixture pinning the correct behavior (once decided — see next), and pins on the broken side. Undecided: should a static-array loop's `ref` be (a) faithfully invoked once at construction (matching what a signal-backed array does on its one-time initial render), or (b) loudly refused at compile time since the static fast path's whole premise is "nothing here ever needs a live DOM handle"? That's a design question for whoever picks this up — not answered here.

kfly8:
## Summary

Stacked on #2890 (independent fix, same bug-ticket sweep).

A static (non-signal) array's nested `.map()` over PLAIN ELEMENTS (not child components) inside a static outer row wired up nothing at all — no `ref` invocation, no reactive text or attribute effect — even though the outer row's own bindings, and a depth-N CHILD COMPONENT inner loop, were already faithfully wired. `buildStaticArrayChildInitsPlan`'s `elem.innerLoops` pass only ever ran when it found a matching nested CHILD COMPONENT (`innerComps.length === 0` skipped the loop entirely), so a plain-element inner loop's row stayed frozen at its SSR-baked value: neither a `ref` callback nor a signal-derived text/attr inside it ever updated.

The issue's own repro (a TOP-LEVEL static loop with a `ref`) turned out to already work correctly — that's pinned here as regression armor against the issue's own misdiagnosis being reintroduced. The real, still-open gap was one level deeper.

The existing `inner-loop-nested` double-`forEach` shape (`plan/build-static-array-child-init.ts`, `stringify/static-array-child-init.ts`) now also wires the inner loop's own reactive attrs/texts/refs, scoped to depth-1 nesting under a plain-element-rooted outer item — `NestedLoop` has no parent link to thread a second offset through for deeper nesting, so that stays a documented known limitation (`nested-loop-ref-const.ts`'s docstring updated to describe the current, narrower gap accurately).

Fixes https://github.com/piconic-ai/barefootjs/issues/2798.

## Test plan

- [x] `bun test packages/jsx/src/__tests__/issue-2798-static-nested-loop-bindings.test.ts` — new unit tests: top-level static ref (regression armor), nested-static ref invocation, nested-static reactive text, nested-static reactive attr.
- [x] `bun test packages/jsx/src/__tests__/nested-loop-plain.test.ts packages/jsx/src/__tests__/nested-loop-index-param.test.ts packages/jsx/src/__tests__/compiler-stress-1244.test.ts` — 104 pass / 3 todo / 0 fail, no regressions.
- [x] New conformance fixture `static-nested-loop-ref` added and registered; `expectedHtml` regenerated from the Hono reference via `generate-expected-html.ts` (unaffected by the fix — SSR never runs a `ref`, matching `nested-loop-ref-const.ts`'s precedent). `git status --short` confirmed clean after regeneration (no unrelated fixture drift).
- [x] `bun test packages/jsx` — full suite, 3367 pass / 5 fail, all 5 failures confirmed pre-existing and unrelated (see #2889's PR description).
- [x] `bunx tsgo --noEmit -p packages/jsx/tsconfig.json` — clean.

---

### DIFF
--- a/.changeset/static-nested-loop-bindings-2798.md
+++ b/.changeset/static-nested-loop-bindings-2798.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+Fix #2798: a static (non-signal) array's nested `.map()` over PLAIN ELEMENTS (not child components) inside a static outer row wired up nothing at all — no `ref` invocation, no reactive text or attribute effect — even though the outer row's own bindings, and a depth-N CHILD COMPONENT inner loop, were already faithfully wired. `buildStaticArrayChildInitsPlan`'s `elem.innerLoops` pass only ever ran for a matching nested child component (`innerComps.length === 0` skipped the loop entirely), so a plain-element inner loop's row stayed frozen at its SSR-baked value. The same `inner-loop-nested` double-`forEach` shape now also wires the inner loop's own reactive attrs/texts/refs, scoped to depth-1 nesting under a plain-element-rooted outer item (deeper nesting remains a documented known limitation — `NestedLoop` has no parent link to thread a second offset through).
--- a/packages/adapter-go-template/src/conformance-pins.ts
+++ b/packages/adapter-go-template/src/conformance-pins.ts
@@ -101,4 +101,19 @@ export const conformancePins: ConformancePins = {
   // resolves the primitive normally and never reaches this refusal (formerly
   // tracked as #2771, closed) — no open issue tracks further work.
   'namespace-import-primitive': [{ code: 'BF013', severity: 'error' }],
+  // #2893: `analyzeBakeableStaticElementLoop`'s static-array bake (Go's only
+  // path to bind a static/non-signal array as a loop source — `html/template`
+  // has no slice literal syntax) bails the WHOLE loop when the body contains
+  // a nested `loop` node (module docstring, `static-element-loop-bake.ts`).
+  // This fixture's outer static array's row has a nested `.map()` over
+  // `item.children`, so it falls through to the generic "computed loop
+  // array" BF101 refusal. Unrelated to #2798's own fix (a client-JS-only
+  // ref/reactive-binding gap) — this is a pre-existing Go-adapter SSR
+  // static-loop-baking scope boundary the new fixture happened to be the
+  // first to exercise.
+  'static-nested-loop-ref': [{
+    code: 'BF101',
+    severity: 'error',
+    issue: 'https://github.com/piconic-ai/barefootjs/issues/2893',
+  }],
 }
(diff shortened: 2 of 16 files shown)


kfly8:
CI's Go template adapter suite caught a real failure on the new `static-nested-loop-ref` fixture — not a bug in this PR's client-JS fix, but a pre-existing, unrelated gap in the Go template adapter's own SSR static-loop-baking (`analyzeBakeableStaticElementLoop` explicitly bails when a static array's row contains a nested `.map()`, per that module's own documented scope). This fixture's shape happened to be the first one to exercise "static outer array + nested inner loop" against that Go-specific code path.

Filed https://github.com/piconic-ai/barefootjs/issues/2893 to track it and pinned it in `conformance-pins.ts` (same mechanism as the existing `static-array-from-props` / `filter-nested-callback-predicate` pins) in 2eb2c8e. Verified locally that the pin resolves both the failing "JSX Conformance Tests" and "claim-plan conformance" checks; this PR's own fix (client-JS ref/reactive-binding wiring) is untouched.

---

Issue #2894, opened 2026-09-08
go-template and erb: collectStringValueNames doesn't resolve renamed body-destructured prop aliases (#2883 sibling gap)


kfly8:
## Summary

#2883 fixed Mojolicious's `collectStringValueNames` (the string-typed-name witness that decides `eq`/`ne` vs `==`/`!=`, and `.`-concat vs numeric `+`) to also resolve a bare-props-form BODY-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) via the shared `resolveBodyDestructuredPropAliases` helper. Two sibling adapters have the identical missing-witness gap, with live (not just "parity") consumers:

- **`packages/adapter-go-template/src/adapter/props/prop-classes.ts:53`** (`collectStringValueNames`) feeds `_isStringValueName` (`go-template-adapter.ts:5319`), which `binary()` (`go-template-adapter.ts:5309`) uses via `isStringConcatBinary` to choose `bf_add` vs `bf_concat_str` for `+`. A renamed body-destructured string prop used in a `+` concat (`someString + skipLabel`) would misclassify as numeric addition instead of string concatenation.
- **`packages/adapter-erb/src/adapter/props/prop-classes.ts:79`** feeds `_isStringValueName` (`erb-adapter.ts:1951`), which `isStringTypedOperand` (`expr/operand.ts`) uses to choose Hash-vs-Array index-access lowering. A renamed body-destructured string prop used as an index (`obj[skipLabel]`) would misclassify the same way.

Confirmed by reading both consumers directly: neither has any `resolveBodyDestructuredPropAliases` handling analogous to the fix landed for Mojolicious, and both downstream checks look at the raw identifier name with no alias translation upstream.

## Found via

Surfaced by pullfrog review on #2890 (the #2883 fix) — flagged as informational, not a blocker for that PR since it's pre-existing and untouched by that diff.

## Suggested fix

Apply the same `resolveBodyDestructuredPropAliases`-based fix to `collectStringValueNames` in both `packages/adapter-go-template/src/adapter/props/prop-classes.ts` and `packages/adapter-erb/src/adapter/props/prop-classes.ts`, mirroring the Mojolicious change. Per this repo's three-piece convention, land with a conformance fixture pinning the corrected behavior for each adapter (a renamed prop used in a `+` concat for go-template; used as an index-access key for erb).

Tier: `+ bug` (silent divergence — output silently differs, no diagnostic).

Issue #2491, opened 2026-08-02
ERB/Go/Twig: dynamic-key element access on a loop row (`tone[k]`) renders empty at SSR


kfly8:
Found while landing the `loop-param-shadows-record-template-span` fixture (#2482 audit follow-up). The compiler-side fold bug that fixture primarily pins is fixed; ERB and Go then fail the fixture's fallback path for their own reasons.

## Repro

```tsx
const tone = { a: 'outer-a', b: 'outer-b' }

export function LoopParamShadowsRecordTemplateSpan({ items, k }: { items: { id: number; a: string; b: string }[]; k: string }) {
  return (
    <ul>
      {items.map((tone) => (
        <li key={tone.id} data-t={`t ${tone[k]}`}>{tone.a}</li>
      ))}
    </ul>
  )
}
```

With `props: { items: [{ id: 1, a: 'row1-a', … }], k: 'a' }` the reference renders `data-t="t row1-a"`.

## Observed

Both ERB and Go render `data-t="t "` — the dynamic lookup silently produces nothing (not the outer const, not the row value).

**ERB mechanism** — key-type mismatch. The emission is right:

```erb
<li … data-t="<%= bf.h('t ' + bf.string(tone[v[:k]])) %>" …>
```

but row hashes are deserialized with SYMBOL keys (`tone[:id]`, `tone[:a]`) while `v[:k]` yields the STRING `"a"` — `tone["a"]` on a symbol-keyed hash is `nil`. Dynamic lookups need `.to_sym` (or indifferent access) to match the fixed-key accessors emitted elsewhere in the same template.

**Go mechanism** — needs a dig; the emitted template contains no baked const (that part is fine post-fix), but the dynamic-key lookup over the row struct/map resolves to empty at execute time.

## Pins

Conformance fixture `loop-param-shadows-record-template-span` (all other adapters pass) + `renderDivergences` entries carrying this URL in: erb, go-template.

kfly8:
[...]

PR #2896, opened 2026-09-08
Fix #2894: collectStringValueNames misses renamed body-destructured prop aliases (go-template, xslate, twig, blade)

kfly8:
## Summary

Sibling gap to #2883 (Mojolicious), surfaced by pullfrog review on #2890. Two adapters were named in #2894; only one turned out to have a live bug.

**go-template (real fix):** `collectStringValueNames` (`packages/adapter-go-template/src/adapter/props/prop-classes.ts`) builds the string-typed-name witness `isStringConcatBinary` reads to decide `+` string-concat (`bf_concat_str`) vs numeric addition (`bf_add`). A bare-props-form BODY-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) has no `propsParams` entry of its own, so the witness never saw it. A `+` concat against the local alias (`t.label + skipLabel`) fell through to `bf_add`, which coerces both operands through `toFloat64` — the string operand silently evaluated to `0` instead of concatenating. `collectStringValueNames` now resolves this alias family via the shared `resolveBodyDestructuredPropAliases` helper, mirroring the fix already shipped for Mojolicious.

**ERB (investigated, not a live bug):** #2894 also named ERB's `collectStringValueNames` as feeding `isStringTypedOperand`/`emitIndexAccessRuby` for `obj[index]` Hash-vs-Array lowering. Tracing the actual call site: since #2491, `emitIndexAccessRuby` ignores its `isStringName` parameter entirely and always lowers to the runtime's polymorphic `bf.get(obj, index)` helper (tries Hash key, then Symbol, then String, then falls back to Array index) rather than choosing a lowering at compile time. So the missing witness never reaches a decision point — adding the alias-resolution block to ERB would change no output. Rather than claim a fix that isn't one, I added ERB-adapter-scoped tests (`packages/adapter-erb/src/__tests__/erb-adapter.test.ts`) pinning this finding: a real end-to-end Ruby render proving the renamed-alias key resolves correctly, plus an assertion that the emitted template routes through `bf.get`.

Design consulted with `fable` before implementing (one PR since #2894 is one issue/one unit; no ERB code change since the alias block would feed a witness nothing reads; fixture pairing kept to one adapter's discriminating shape each rather than symmetric coverage).

## Test plan

- [x] New unit test `packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts` (#2894 describe block) — asserts `bf_concat_str .Label $.FallbackLabel`, not `bf_add .Label $.FallbackLabel`. Confirmed it fails pre-fix (stashed the `prop-classes.ts` change and reran) and passes post-fix.
- [x] New conformance fixture `string-concat-plus-renamed-prop` (registered in `packages/adapter-tests/fixtures/index.ts`), `expectedHtml` generated from the Hono reference via `generate-expected-html.ts` — unaffected by the fix since Hono's native `+` needs no witness.
- [x] `bun test packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts` — full file, 1987 pass / 0 fail / 26 skip (skips are pre-existing real-Go-render tests, no `go` toolchain in this environment).
- [x] `bun test packages/adapter-erb/src/__tests__/erb-adapter.test.ts` — full file, 1685 pass / 8 fail / 22 skip. All 8 failures are `date-tolocale-named-tz` (missing `tzinfo` Ruby gem in this environment) — pre-existing, unrelated to this change (confirmed by name; nothing in this diff touches Date/timezone lowering).
- [x] Fixture cross-checked against every environment-available real toolchain (Hono, Jinja, minijinja) — 18 pass / 0 fail; Perl/PHP-backed adapters (Mojolicious, Xslate, Twig, Blade) skip real rendering in this sandbox (toolchain not installed), same as every other fixture here.
- [x] `bunx tsgo --noEmit` clean for both `packages/adapter-go-template` and `packages/adapter-erb`.
- [x] `ui/compat.lock.json` regenerated — 558/558 cells ok, 0 diagnostics. `coverage-map.json` / `generated-data-points.json` regenerated, diff limited to the new fixture's contribution.

Stacked on the now-merged #2891/#2892 (this PR's base is `main`, which already includes both).

---

### DIFF
--- a/.changeset/go-template-renamed-prop-concat-2894.md
+++ b/.changeset/go-template-renamed-prop-concat-2894.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Fix #2894 (go-template sibling of #2883): a `+` concat against a bare-props-form BODY-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) lowered to Go's numeric `bf_add` instead of `bf_concat_str`, because `collectStringValueNames`'s string-typed-name witness had no notion of this alias family. `bf_add` coerces both operands through `toFloat64`, so the string operand silently evaluated to `0` instead of concatenating. `collectStringValueNames` now also resolves this alias family via the shared `resolveBodyDestructuredPropAliases` helper, mirroring the fix already shipped for the Mojolicious adapter.
--- a/.changeset/xslate-twig-blade-renamed-prop-concat-2894.md
+++ b/.changeset/xslate-twig-blade-renamed-prop-concat-2894.md
@@ -1,0 +1,7 @@
+---
+"@barefootjs/xslate": patch
+"@barefootjs/twig": patch
+"@barefootjs/blade": patch
+---
+
+Fix #2894 sibling gap (surfaced by the `string-concat-plus-renamed-prop` fixture added in this same PR): a `+` concat against a bare-props-form BODY-destructured, RENAMED prop (`const { fallbackLabel: skipLabel } = props`, the #2788 alias family) lowered to each adapter's numeric `+` instead of its string-concat operator (Kolon's `~` for Xslate, `~` for Twig, `.` for Blade), because `collectStringValueNames`'s string-typed-name witness had no notion of this alias family. Xslate's `+` coerces a non-numeric string operand to `0`; Twig and Blade's PHP arithmetic underneath fatals with "Unsupported operand types". `collectStringValueNames` now resolves this alias family via the shared `resolveBodyDestructuredPropAliases` helper in all three adapters, mirroring the fix already shipped for Mojolicious (#2883) and go-template (this PR).
--- a/packages/adapter-blade/src/__tests__/blade-adapter-unit.test.ts
+++ b/packages/adapter-blade/src/__tests__/blade-adapter-unit.test.ts
@@ -758,3 +758,33 @@ export function Counter() {
     expect(template).not.toContain('<link')
   })
 })
+
+describe('BladeAdapter - collectStringValueNames resolves renamed body-destructured prop aliases (#2894)', () => {
+  // Sibling gap to #2883 (Mojolicious) and this same PR's go-template fix:
+  // a `+` concat against a bare-props-form BODY-destructured, RENAMED prop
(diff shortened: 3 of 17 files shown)


kfly8:
The `lockfile` check failed: `ui/support-matrix.lock.json` was stale relative to the fixture added in this PR — my local regeneration ran under heavy CPU contention from several concurrent background test jobs and produced byte-identical output instead of picking up the new fixture. Not caught locally because the freshness test happened to pass against that same stale state.

Fixed in 8b97bf0: rebuilt all 8 dist-consuming adapter packages the way the CI job does, regenerated `ui/support-matrix.lock.json` fresh (confirmed a real 190-line diff this time), and re-ran `bun test packages/compat` clean (527 pass / 0 fail, including the freshness test). `ui/compat.lock.json`, `coverage-map.json`, and `generated-data-points.json` were re-checked against the same clean rebuild and already matched — no further drift.

---

---

kfly8:
Three `test` checks failed on the previous commit (blade, twig, xslate). Root cause: the `string-concat-plus-renamed-prop` fixture added in this PR runs against every adapter, and CI's real Perl/PHP toolchains caught that **Xslate, Twig, and Blade each carry their own independent copy of `collectStringValueNames` with the identical missing-alias-witness gap** this PR fixes for go-template — undetected locally since those adapters' real-render tests silently skip here (no Perl/PHP toolchain in this sandbox).

Fixed in aaedb31: applied the same `resolveBodyDestructuredPropAliases`-based fix to all three adapters' `props/prop-classes.ts`, added a regression test per adapter pinning the correct concat operator (`~` for Xslate/Twig, `.` for Blade), and verified the emitted template output directly via a standalone compile+generate script since real rendering isn't available locally. `bun test` across the three adapter packages: 5111 pass / 0 fail. `tsgo --noEmit` clean. `bun test packages/compat`: 527 pass / 0 fail (no lockfile drift).

This widens the PR beyond the two adapters #2894 named, but it's the same fix already proven twice in this stack (Mojolicious #2883, go-template earlier in this PR) — not a new investigation, and leaving it unfixed would mean either reverting the fixture (losing go-template's regression coverage) or shipping a fixture that's known-red on three adapters.

---

---

Issue #2925, opened 2026-09-10
go-template: a signal getter handed to a plain child-component prop renders empty (static constructor baker drops the field)


kfly8:
## Problem

Found while building the #2760 fixture (`component-prop-bare-getter`). A component prop whose value is (or contains) a local signal/memo getter — `<Display value={count} />` — compiles clean on go-template, but renders the field as Go's zero value instead of the signal's value: `NewCounterProps`'s constructor-time build of the nested `DisplaySlot0` field omits `Value` from the `DisplayInput{...}` literal entirely.

## Reproduction

```tsx
'use client'
import { createSignal } from '@barefootjs/client'

function Display(props: { value: () => number }) {
  return <span class="value">{props.value()}</span>
}

export function Counter() {
  const [count, setCount] = createSignal(5)
  return (
    <div class="root">
      <Display value={count} />
    </div>
  )
}
```

Hono (reference) renders `<span ...>5</span>`. Go renders `<span ...></span>` — empty. The generated `Counter.types` constructor:

```go
func NewCounterProps(in CounterInput) CounterProps {
	...
	return CounterProps{
		...
		Count: 5,
		DisplaySlot0: NewDisplayProps(DisplayInput{
			ScopeID: scopeID + "_s0",
			BfParent: scopeID,
			BfMount: "s0",
			// Value: ... is missing — DisplayInput.Value silently defaults to nil
		}),
	}
}
```

`Count: 5` proves `count`'s own signal-to-Go-constant baking works fine; the bug is specifically in threading a signal-getter-valued prop into a **nested child's** constructor-time `Input` struct literal — that baker path drops the field instead of emitting `Value: 5` (or an `in.Count`-style forward, if the general case needs to stay live rather than baked).

**Reproduces identically with the already-legal object-literal-wrapped form** (`<Display value={{ v: count }} />` → `DisplayInput{ ... }` also missing `Value:`), so this predates #2760 and is orthogonal to its diagnostic-gate decision — same underlying gap, just never exercised by any existing fixture before (the 66 measured real-world Context-Provider uses of this idiom all route through `provideContext`/`useContext`, not through a plain child-component's constructor-time `Input` struct).

## Relationship to #2700

[...]

Issue #2924, opened 2026-09-10
A signal getter passed as a component prop crashes on CSR-fresh-mount (module-scope `template` lambda can't reach an init-scoped local)


kfly8:
## Problem

Found while building the #2760 fixture (`component-prop-bare-getter`). A component whose child prop value is (or contains) a local signal/memo getter reference compiles cleanly, but throws a `ReferenceError` at runtime whenever the PARENT is CSR-mounted fresh — i.e. `createComponent('Parent', props)` invoking `def.template(props)` (`packages/client/src/runtime/component.ts:1240`), which happens for any component created client-side without going through SSR+hydration (a new loop row, a portal, a conditionally-mounted subtree, etc.). Hydration of SSR-rendered markup is unaffected — this is CSR-fresh-mount only.

## Reproduction

```tsx
'use client'
import { createSignal } from '@barefootjs/client'

function Display(props: { value: () => number }) {
  return <span class="value">{props.value()}</span>
}

export function Counter() {
  const [count, setCount] = createSignal(5)
  return (
    <div class="root">
      <Display value={count} />
    </div>
  )
}
```

Compiles with zero errors. The emitted `Counter.client.js`:

```js
export function initCounter(__scope, _p = {}) {
  ...
  const [count, setCount] = createSignal(5)
  ...
  initChild('Display__c3b9abc9', _s0, { get value() { return count } })  // correct: closes over count
}

hydrate('Counter', {
  init: initCounter,
  template: (_p) => `<div class="root">${renderChild('Display__c3b9abc9', {value: count}, undefined, 's0')}</div>`
  //                                                                       ^^^^^ ReferenceError: count is not defined
})
```

`template` is registered as a **module-scope** lambda (`hydrate('Counter', { template: ... })` is a top-level call), so it has no closure access to `count`, a local declared inside `initCounter`'s own function scope. `initChild` (used during hydration/init) gets this right — it runs inside `initCounter` and closes over `count` correctly. Only the `template` lambda's own `renderChild(...)` props-literal is wrong.

**Reproduces identically with the already-legal object-literal-wrapped form** (`<Display value={{ v: count }} />` → `renderChild('Display__c3b9abc9', {value: { v: count }}, ...)`), so this is independent of #2760's diagnostic-gate decision — it predates that change and affects both forms of the Context-Provider idiom equally when passed as an ordinary (non-Context) component prop.

## Why the 66 measured Context-Provider uses (#2760) don't hit this

[...]

PR #2926, opened 2026-09-10
Fix #2760: allow a bare signal/memo getter passed directly to a component prop

### LINKED ISSUE #2760: BF044 refuses a getter passed directly to a component prop but allows the same getter wrapped in an object literal

kfly8:
## Summary

Resolves #2760's tracked BF044 inconsistency: `<Foo x={val} />` (a bare signal/memo getter passed directly to a component prop) refused with BF044, while `<Foo x={{ val }} />` (the identical accessor wrapped in an object literal) passed silently. Per the issue's own analysis and user confirmation, the two forms are one deliberate idiom (this codebase's Context-Provider pattern — measured: 66 real-world uses, none a forgotten `()`), not an inconsistency to resolve by tightening the direct form.

- `checkBareSignalOrMemoIdentifier` (`packages/jsx/src/jsx-to-ir.ts`) now gates its entire check — top-level included, not just its nested-descent walk — on whether the position is genuinely RENDERED (a DOM attribute, a JSX text child). A component prop never is, so both forms now compile identically. Renamed the option `descendNested` → `renderedPosition` since it now controls both levels.
- Added a positive test case (`signal-getter-not-called.test.ts`) and documented the exemption in `spec/compiler.md`'s BF044 section, plus `docs/core/advanced/error-codes.md`'s BF044 example (caught by Pullfrog review — its example used exactly the now-exempted shape).
- New conformance fixture `component-prop-bare-getter` (`packages/adapter-tests/fixtures/`) proves the newly-legal form compiles and renders correctly, per `spec/subset-conformance.md`'s change-time coupling rule.

## Two pre-existing defects surfaced (not caused by this change)

Building the fixture exercised this idiom outside a Context Provider for the first time in the corpus (the 66 real-world uses all route through `provideContext`/`useContext`, never a plain child-component prop), which surfaced two independent, pre-existing bugs that reproduce **identically with the already-legal object-literal-wrapped form** — verified directly against both forms before filing:

- **#2924** — the CSR-fresh-mount `template` lambda (`hydrate('X', { template: ... })`, module scope) references a local signal getter from `renderChild(...)`'s props literal, a guaranteed `ReferenceError` whenever the component is created via CSR without SSR hydration. Pinned in `csr-skip-set.ts` (`CSR_SKIP_FIXTURES`) and `client-js-scope.test.ts`'s `KNOWN_UNDECLARED` scope-gate ledger.
- **#2925** — go-template's constructor-time nested-child-props baker (`NewCounterProps`) silently omits the field when its value is a signal getter, so Go renders it as the zero value instead of the signal's value. Pinned in `render-divergences.ts` (adapter-conformance `skipJsx`).

Both are filed as `known-limitation + bug` with full repro, root-cause analysis, and fix directions in their issue bodies, following the repo's fixture-first defect workflow.

## Test plan

- [x] `bun test packages/jsx/src/__tests__/signal-getter-not-called.test.ts packages/jsx/src/__tests__/invalid-signal-usage.audit.test.ts packages/jsx/src/__tests__/doc-examples.test.ts` — clean, including the new positive case.
- [x] `bun test packages/jsx/src packages/adapter-tests/src packages/compat/src packages/compat/__tests__ ui/components` — 7788 pass, 0 fail.
- [x] `GOTOOLCHAIN=go1.25.6 bun test packages/adapter-go-template/src` — 2111 pass, 0 fail (with real `go run` execution), including the new fixture correctly rendering the CSR-skip/render-divergence pins.
- [x] `bun test packages/adapter-erb/src packages/adapter-jinja/src packages/adapter-rust/src -t "component-prop-bare-getter"` — clean on all three (ERB/Jinja/minijinja route the accessor through live template evaluation, not a static constructor baker, so they don't hit #2925's class of bug).
- [x] `bun run packages/adapter-tests/scripts/generate-expected-html.ts` — regenerated `expectedHtml` for the new fixture from the Hono reference; clean beyond the intended addition.
- [x] `bun packages/adapter-tests/scripts/coverage-map.ts` / `bun run support-matrix:lock` / `bun run compat:lock` — regenerated all derived lock artifacts, 0 diagnostics.
- [x] CI — all checks green (34/34), including `e2e-blade`/`e2e-php`/`e2e-laravel` confirming the PHP-backed adapters are unaffected. Pullfrog reviewed twice (initial finding: stale `docs/core/advanced/error-codes.md` example — fixed; follow-up: no new issues).

---

### DIFF
--- a/.changeset/bf044-component-prop-getter-subset.md
+++ b/.changeset/bf044-component-prop-getter-subset.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/jsx": patch
+---
+
+BF044 (signal/memo getter passed without calling it) no longer refuses a bare getter passed directly to a component prop (`<Foo x={val} />`) — it now compiles, matching the already-legal object-literal-wrapped form (`<Foo x={{ val }} />`). Both forms are this codebase's deliberate Context-Provider idiom: the child owns *when* to call the accessor, so it can subscribe reactively at its own read site instead of the value being frozen at the parent's render time (measured: 66 real-world uses of this idiom, none a forgotten `()`). The diagnostic's entire check — top-level included, not just its nested-descent walk — is now gated on whether the position is genuinely RENDERED (a DOM attribute, a JSX text child), which a component prop never is (#2760).
--- a/.changeset/go-template-component-prop-getter-pin.md
+++ b/.changeset/go-template-component-prop-getter-pin.md
@@ -1,0 +1,5 @@
+---
+"@barefootjs/go-template": patch
+---
+
+Pins a newly-discovered divergence on the `component-prop-bare-getter` conformance fixture (#2760): a plain (non-Context) child-component prop whose value is a local signal/memo getter compiles clean but renders empty on Go's real backend — `NewCounterProps`'s constructor-time build of the nested child's `Input` struct silently omits the field. Predates #2760 and reproduces identically with the object-literal-wrapped form; tracked at #2925.
--- a/docs/core/advanced/error-codes.md
+++ b/docs/core/advanced/error-codes.md
@@ -361,19 +361,28 @@ function Child({ initialCount }: Props) {
 
 ### BF044 — Signal/Memo Getter Not Called
 
-**Trigger:** Signal/memo getter passed without calling it.
+**Trigger:** Signal/memo getter passed without calling it in a RENDERED
+position — a DOM element attribute or a JSX text child, where the value
+becomes literal output.
 
 ```tsx
 // ❌ BF044
-<Child count={count} />  // Passing getter function, not the value
+<div count={count} />  // Passing getter function, not the value
 ```
 
 **Fix:**
 
 ```tsx
 // ✅ Fixed
-<Child count={count()} />
+<div count={count()} />
 ```
+
+**Not triggered on a component prop:** `<Child count={count} />` compiles —
+a component prop is an opaque value handed to the child, not rendered
+output, and passing a live getter there is this codebase's deliberate
+Context-Provider idiom (the child calls it at its own read site). See
+[`spec/compiler.md`'s BF044 section](https://github.com/piconic-ai/barefootjs/blob/main/spec/compiler.md#signalmemo-getter-not-called-bf044)
+for the full rule.
 
 <a id="bf049"></a>
 
--- a/packages/adapter-go-template/src/render-divergences.ts
+++ b/packages/adapter-go-template/src/render-divergences.ts
@@ -37,4 +37,19 @@
 
 import type { RenderDivergences } from '@barefootjs/jsx'
 
-export const renderDivergences: RenderDivergences = {}
+export const renderDivergences: RenderDivergences = {
+  // #2925: a plain (non-Context) child-component prop whose value is a
+  // local signal/memo getter — `<Display value={count} />`, and identically
+  // `<Display value={{ v: count }} />` — compiles clean, but
+  // `NewCounterProps`'s constructor-time build of the nested `DisplaySlot0`
+  // field silently omits `Value` from the `DisplayInput{...}` literal, so Go
+  // renders the field as its zero value instead of the signal's value.
+  // `count`'s own SSR-default baking works fine (`Count: 5` lands); the gap
+  // is specifically in threading a signal-getter-valued prop into a NESTED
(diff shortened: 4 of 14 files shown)


kfly8:
The `lockfile` check failed on the first push: `support matrix freshness > committed ui/support-matrix.lock.json matches a fresh computeSupportMatrix()`.

Root cause: local `bun run support-matrix:lock`/`compat:lock` resolved `@barefootjs/go-template` through its stale, pre-existing `dist/index.js` (built before this PR's `render-divergences.ts` edit — 8 of 9 adapter packages resolve via `dist/`, not `src/`, per `ci-compat.yml`'s own note), so the committed lock files never picked up the new `component-prop-bare-getter` render-divergence entry. CI's `lockfile` job rebuilds `dist/` fresh before regenerating, so it correctly computed a different (and correct) matrix than what I'd committed.

Fixed by rebuilding `@barefootjs/go-template` locally before regenerating both lock files, and pushed the corrected `ui/compat.lock.json` / `ui/support-matrix.lock.json` (35-line diff, exactly the expected `component-prop-bare-getter` → `go-template` gap addition — verified via `bun test packages/compat`, all 526 tests green). Not a flake; a real staleness bug on my end, now fixed.

---

kfly8:
Fixed in 16a3967c4 — swapped `docs/core/advanced/error-codes.md`'s BF044 example for a genuinely rendered DOM-attribute case and added the same component-prop exemption note as `spec/compiler.md`, cross-linked to it.

---

PR #2932, opened 2026-09-11
Move B: keep destructured "use client" props reactive

kfly8:
## Summary

Second of four PRs implementing the "Prop Boundary Contract" design (co-designed with Fable). Stacked on #2931 (Move A), which is stacked on #2926.

Today, a `"use client"` component that destructures its props in the function parameter (`function Child({ value, onClick })`) gets exactly one `const value = _p.value` extraction at the top of `init*`, and every downstream reference — effects, memos, handlers, most text positions — reads that captured-once local instead of re-reading `_p.value` live. The parent side is always live (`initChild('Display', _s0, { get value() { return count() } })`), so the child silently never sees updates past its first render. BF043 exists to steer users away from destructuring entirely for exactly this reason.

This PR makes destructuring behave exactly like `function Component(props)` mode already does: every value-position read of a destructured prop becomes a live `_p.<callerKey>` read via a final whole-init-body AST rewrite pass (`rewriteDestructuredPropReads`, `ir-to-client-js/rewrite-destructured-props.ts`), sharing one formula (`livePropReadExpr`, `props-binding.ts`) with the existing props-object codepaths. Since destructured props are now always reactive, BF043 is retired.

A second, previously-unaccounted-for site was found and fixed along the way: the Phase-1 CSR-only `template:` literal (`rewriteBarePropRefs` in `jsx-to-ir.ts`) never applied a destructured prop's default value, so e.g. `data-label={label}` with `label = 'none'` silently omitted the attribute on a client-only first render even though SSR always got it right. Fixed via the same `livePropReadExpr` formula.

## What changed

- `rewriteDestructuredPropReads` (new) — the single door: rewrites every non-shadowed destructured-prop identifier in the finished init body to a live `_p.<callerKey>` read (or `(_p.<callerKey> ?? <fallback>)`).
- `livePropReadExpr` (new, `props-binding.ts`) — the one shared fallback formula, replacing three previously-duplicated copies.
- `props-extraction.ts` — now only emits the one-time `children` extraction (its getter instantiates child components, so it must stay captured-once).
- `props-event-handlers.ts` — drops the destructured-mode handler alias, which would otherwise shadow the rewrite.
- BF043 removed (`analyzer.ts`, `analyzer-context.ts`, `errors.ts`) — the warning is no longer true.
- `props-reactivity-comparison` fixture flipped from pinning non-reactivity to pinning props-object/destructured parity.
- New `destructured-props-live` fixture exercising every live-read position (text, memo, effect, handler, attribute default, aliasing, shadowing).
- Docs updated (`spec/compiler.md`, `docs/core/reactivity/props-reactivity.md`, `docs/core/advanced/error-codes.md`, etc.) to describe destructuring as reactive.

## Test plan

- [x] `packages/jsx` full suite: 3402 pass / 27 skip / 3 todo / 0 fail (277 files)
- [x] `packages/adapter-tests` full suite: 2604 pass / 5 skip / 0 fail (23 files), including `coverage-map.test.ts` and `csr-conformance.test.ts`
- [x] `tsgo --noEmit -p .` clean
- [x] `generate-expected-html.ts` regenerated and diffed by hand — this branch is stacked, so CI's `update-expected-html` fixture-drift job will not run on it automatically
- [ ] E2E/Playwright — could not run in this sandbox (no browser binaries); not attempted
- [ ] CI

---

---


kfly8:
The `check` job (Changeset check) failed because this PR's own `base...head` diff carries no added `.changeset/*.md` file. That's intentional, not an oversight: this PR is stacked under #2933 (Move C), and at the maintainer's request the two PRs' changes are described by one consolidated changeset — `.changeset/prop-boundary-contract.md` — added at the top of the Move B→C stack (in #2933) rather than duplicated per PR. Since this PR's diff alone will never carry it, I've added the `no-changeset` label to skip the gate here; the actual changelog entry lands with #2933.

---

PR #2933, opened 2026-09-11
Move C: enforce the function-prop hydration boundary at the Hono root

kfly8:
## Summary

Third of four PRs implementing the "Prop Boundary Contract" design (co-designed with Fable). Stacked on #2932 (Move B), which is stacked on #2931 (Move A), which is stacked on #2926.

A function-typed prop that a hydration ROOT's own client-side code calls live used to be silently dropped from the `bf-p` JSON hydration payload — a generated `typeof x !== 'function'` guard filtered it out before it ever reached `serializeHydrationProps` — so the client hydrated against `undefined` and crashed or misbehaved with no diagnostic pointing at the cause. (A CHILD component never has this problem: it always receives props live via `initChild`/`upsertChild`, never via `bf-p`.)

This PR fixes the boundary at the only place it actually needs to be enforced — the Hono adapter's SSR serialization path for a hydration root — rather than adding a new compile-time diagnostic (a component can legitimately be used as both a root and a child in different contexts, and only the runtime "is this a root mount" fact — `__bfChild` — decides whether serialization is even reachable).

## What changed

- `analyzeClientNeeds` (`packages/jsx`) now also computes `clientAnalysis.liveOnlyProps`: the subset of a component's client-read props whose declared type resolves to a function type, keyed by caller-facing prop name and valued by the declared type's printable text. Deliberately not folded into BF049 (which fires at declaration time, blind to root-vs-child usage).
- `hono-adapter.ts` now gates the whole `__hydrateProps`/`serializeHydrationProps` computation behind a generated `if (!__bfChild)` check — a child's `__bfPropsJson` is never read (`bf-p` is only emitted when `!__bfChild`), so serializing for a child was always wasted work. This also fixes a pre-existing bug where a child carrying an unrelated rich-type prop (`Map`, `BigInt`, ...) could 500 at SSR even though the serialization result was always discarded.
- `serializeHydrationProps` (`utils.ts`) takes the `liveOnlyProps` map as a new parameter and throws a clear, actionable `TypeError` only when a live-only prop's runtime value is actually a function on an actual root mount — a function-typed prop that isn't live-only is left to `JSON.stringify`'s existing silent-drop behavior (unchanged from today).

## Test plan

- [x] `packages/jsx` full suite: 3411 pass / 27 skip / 3 todo / 0 fail (278 files)
- [x] `packages/adapter-hono` full suite: 1864 pass / 20 skip / 0 fail (22 files), including new `live-only-props.test.ts` and extended `serialize-hydration-props.test.ts` (root throw with pinned message, client-parent child, server-parent child, gating-regression Map case)
- [x] `packages/adapter-tests` full suite (pre-fixture-regen): 2604 pass / 5 skip / 0 fail; targeted subset re-run post-regen (`ssr-hydration-contract`, `hydration-props-conformance`, `snapshot-determinism`, `render-contract`, `csr-conformance`): 1564 pass / 0 fail
- [x] `generate-expected-html.ts` regenerated and diffed by hand — stacked branch, so CI's fixture-drift job won't run on it automatically. Only `.ssr.tsx` (generated SSR source) diffs; `.html`/`.client.js` outputs byte-identical.
- [ ] E2E/Playwright — could not run in this sandbox (no browser binaries)
- [ ] CI

---

---


kfly8:
The `pullfrog` check failed, but not because of this PR's code. From the job logs (run [34592244569](https://github.com/piconic-ai/barefootjs/actions/runs/34592244569)), Pullfrog completed its review and drafted an approval (`No new issues found`, one informational nitpick about `computeLiveOnlyProps`'s docstring not also covering a 2+-arm function-type union) — but its own `create_pull_request_review` call hit a "diff coverage pre-flight" nudge asking it to re-read some generated `__snapshots__/` fixture files, and before it could retry and submit, it hit its 20-minute agent timeout. The review was never posted; nothing in this PR's diff is implicated.

This check reports run status only and doesn't gate merging. I've re-run the failed job once (`rerun_failed_jobs` on the run above) to get an actual review posted.

---

Issue #2940, opened 2026-09-12
A body-destructured prop's arrow-function default emits invalid JS (`_p.x ?? (v) => ...`)


kfly8:
## Summary

`const { fmt = (v) => 'v' + v } = props` (a body destructure whose default value is an arrow function) compiles to a genuine JavaScript `SyntaxError` in the emitted client JS. This is pre-existing — it reproduces identically before and after #2934's fix (which makes a body-destructured pure alias's *non-function* default live) — and is orthogonal to that fix's scope.

## Repro

```tsx
'use client'
function Child(props: { fmt?: (v: number) => string }) {
  const { fmt = (v) => 'v' + v } = props
  return <span>{fmt(1)}</span>
}
```

Compiled `init*` body:

```js
const fmt = _p.fmt ?? (v) => 'v' + v
```

`a ?? (v) => expr` is not valid JS — an arrow function cannot appear directly as the right-hand operand of `??` without parentheses (`a ?? ((v) => expr)`). This is a genuine parse error in the emitted bundle, not merely a stylistic nit.

The equivalent PARAMETER-destructured form (`function Child({ fmt = (v) => 'v' + v }: Props)`) does NOT have this problem — its default is wrapped correctly (`(_p.fmt ?? ((v) => 'v' + v))`), matching `ParamInfo.defaultContainsArrow`'s existing paren-wrapping in `propReadFallback` (`packages/jsx/src/props-binding.ts`).

## Root cause

`analyzer.ts` (~line 3160-3163) builds the body-destructured constant's `value` text for a defaulted binding as `${baseValue} ?? ${defaultValueExpr}` unconditionally — it does not check whether `defaultValueExpr` contains an arrow function/function expression and wrap it in parens the way the parameter-destructured path already does via `ParamInfo.defaultContainsArrow`.

[...]

PR #2945, opened 2026-09-12
fix(jsx): make body-destructured props read live in props-object mode

### LINKED ISSUE #2934: Body-destructured props (`const { x } = props`) still capture once, with no diagnostic since BF043's removal
## Summary

Move B (#2932, "Prop Boundary Contract" series) made *parameter*-destructured props (`function Child({ value })`) reactively live, and removed BF043 ("props destructuring breaks reactivity") since it's no longer true for that shape.

However, *body*-destructured props in props-object mode — `function Child(props) { const { value } = props; ... }` — are a different code path (`localConstants` pure-alias classification, not `rewriteDestructuredPropReads`) and still capture once, exactly like before Move B. Verified by compiling both forms:

```js
// BODY: const { value } = props        // PARAM: function Child({ value })
const value = _p.value                  // still captured once
createEffect(() => { console.log(value) })   // dead — never re-runs
```

```js
// PARAM form (Move B): reads live
createEffect(() => { console.log(_p.value) })
```

BF043 never actually covered the body-destructure shape either (it only fired for parameter destructuring), so this isn't a regression — but with BF043 gone entirely, there is now **no diagnostic anywhere** steering a user away from this pattern, and the docs (updated by Move B) now describe destructuring generally as reactive without calling out this exception.

## Impact

A user who writes `function Component(props) { const { value } = props; ... }` (a very natural pattern) gets silently stale reads in effects/memos/handlers, with no warning.

## Suggested fix

Extend Move B's `rewriteDestructuredPropReads`-style live-read rewrite to body-destructured pure aliases (`resolveBodyDestructuredPropAliases` in `packages/jsx/src/props-binding.ts` already identifies these bindings), following the same "single door" AST-rewrite approach. Per this repo's fixture-first rule, land this with a conformance fixture demonstrating the live read, not just prose.

## Context

Found during Opus's review pass on Move B (PR #2932). Not blocking Move B's own scope (which was explicitly parameter-destructuring only) — filed as a tracked follow-up.

kfly8:
Fixes #2934

## The bug

The Prop Boundary Contract series (#2931/#2932/#2933/#2936) made a PARAMETER-destructured `"use client"` prop read live:

```tsx
function Child({ value }: Props) {
  createEffect(() => console.log(value)) // live — reads _p.value at each run
}
```

But the syntactically equivalent BODY-destructured form in props-object mode was left behind:

```tsx
function Child(props: Props) {
  const { value } = props
  createEffect(() => console.log(value)) // captured once at mount — stale forever
}
```

Any effect/memo/handler referencing a body-destructured prop went stale and never saw a later update from the parent, with no diagnostic warning about it — BF043 used to flag this, but was retired when the parameter form was fixed since it no longer applied there.

## The fix

Routes body-destructured bindings through the **same** live-read rewrite door the parameter form already uses (`rewriteDestructuredPropReads` / `livePropReadExpr`), rather than building a second implementation — per this repo's "one decision, two implementations" rule.

The subtle part: the captured-once alias declaration (`const value = _p.value`) has to be **suppressed from emission entirely**, not left in place and pruned afterward. The rewrite walks the whole `export function init<Name>(__scope, _p = {}) { … }` function as a `FunctionDeclaration` with a `ts.Block` body, and `scopeFrameOf` (`prop-rewrite.ts`) *does* collect every one of that block's top-level `const`s as a shadow. So if the alias line stayed in the text, the walk would see it as a real local shadowing `value` and silently skip rewriting every reference to it — the "fix" would be a silent no-op. Suppression happens at the final emission step in `emitSortedDeclarations` (`init-declarations.ts`), leaving `classifyLocalDeclarations`/`compute-scope.ts` untouched (smallest blast radius; `neededProps` is already derived correctly from the `props.<key>` regex path, verified).

New structured IR field (`ConstantInfo.propDestructure { key, defaultValue?, defaultContainsArrow? }`, stamped by the analyzer's `collectConstant`) carries what the destructure pattern said, so a new helper — `bodyDestructuredPropParams` (`props-binding.ts`) — can synthesize a `ParamInfo` per eligible binding without a second AST walk. This is used at three sites:

1. `init-declarations.ts` — suppress the alias declaration.
2. `rewrite-destructured-props.ts` — mode-switch to route props-object-mode bindings through the live-read rewrite (parameter mode is unchanged).
3. `jsx-to-ir.ts`'s `getDestructuredPropNames` — merge into the Phase-1 CSR `template:` lambda rewrite, closing a separate divergence: the CSR template previously emitted a bare `_p.label` for a defaulted body destructure, while Hono's SSR applied the default — a silent CSR/SSR gap now closed the same way Move B closed it for the parameter form.

Also corrected an inaccurate docstring in `rewrite-destructured-props.ts` that claimed the init body "parses as a statement list" and that `scopeFrameOf` "would NOT see" a shadow there — false for this text shape, and exactly why the suppression above is mandatory.

## Edge cases handled

- **`children`** — excluded by **caller key** (not local name), so both `const { children } = props` and a rename (`const { children: kids } = props`) keep their real captured-once extraction unchanged. Its value is a getter that instantiates child components; a live re-read would re-mount them.
- **Renames** (`{ n: count }`) — rewrite to `_p.n` (the caller-facing key), not `_p.count`.
- **Defaults** (`{ label = 'n' }`) — survive as a live `(_p.label ?? 'n')` fallback, with correct arrow-default parenthesization, and no synthesized `?? []`/`?? {}` fallback (matches the removed captured-once line's behavior).
- **`let` / mutation** — only `declarationKind === 'const'` **and** `!mutatedAfterDeclaration` entries are eligible; a `let { value } = props` that's reassigned later stays a real, captured, mutable local.
- Controlled-signal sync's pre-existing gap for a renamed body alias (`createSignal(count)` with `{ n: count }`) is untouched — out of scope for this fix, not made worse.

## Test plan (all run locally, numbers below)

**`packages/jsx` full suite:** 3422 pass, 27 skip, 3 todo, **0 fail** (Ran 3452 tests across 278 files). Added a new `describe` block to `destructured-props-live.test.ts` covering: no captured-once locals + live reads (incl. defaults), a direct assertion the alias line is *absent* (not just implied), a renamed destructure, `children` plain/renamed, `.map()`/handler-local shadowing, a `let`-destructured mutated binding, and the CSR `template:` default fix.

**`packages/adapter-tests` full suite:** 2604 pass, 5 skip, **0 fail** (Ran 2609 tests across 23 files), including `coverage-map.test.ts` and `csr-conformance.test.ts` (382 pass alone). Extended the shared `DestructuredPropsLive` fixture with a `BodyLiveChild` sibling (props-object mode, body-destructured, same memo/effect/handler/`data-label` shape as `LiveChild`), wired to the same `value`/`label`/`onPick` source so a discriminating click sequence (`.live-child` pick → `.btn-increment` → `.body-live-child` pick) proves the body form tracks live state independently.

**`tsgo --noEmit -p .`** (jsx package): clean.

**`generate-expected-html.ts`**: regenerated; `git status --short` clean afterward (base is `main`, so `update-expected-html` CI will also run on this PR's own head).

**Real-browser hydration e2e** (`fixture-hydrate.playwright.ts` + `oracle.playwright.ts`, Chromium): 207 passed, 1 pre-existing skip (unrelated `carousel` idempotence) — includes the extended `destructured-props-live` fixture's snap/three-point/idempotence checks and its dedicated hydration interaction test.

## Deviations from the design

None. Followed the verified design exactly, including the mandatory-suppression invariant and all five edge-case resolutions.

---

---


kfly8:
Closing as superseded — #2934 was independently fixed and merged via #2941 (a different session) while this PR was in progress. #2941 is also more thorough than this branch: it additionally found and fixed two real pre-existing reactivity regressions (`AccordionTrigger`'s `className`, `DropdownMenuSubTrigger`'s `disabled`-derived attributes) and filed a separate follow-up (#2940) for an arrow-function-default edge case this branch didn't cover. Nothing here needs to be carried forward.

---

PR #2957, opened 2026-09-12
Re-set the logo wordmark as "BarefootJS" in Instrument Serif

kfly8:
## Summary

The logo's outlined wordmark still read **Barefoot.js**, the project's old spelling. This PR re-sets it as **BarefootJS**, in Instrument Serif Regular (SIL OFL) — the display face the overview deck established — next to the footprint mark. Logo assets only; no site design changes.

## What changed

- **Wordmark**: "BarefootJS" shaped with harfbuzz (kerning applied) and outlined with fontTools from the same Instrument Serif the deck embeds, at a 27-unit cap height with −0.03em tracking. The viewBox keeps its 46-unit height (width 200 → 175.64), so every consumer that sizes the logo by height is unaffected: the site header (`h-[1.65rem]`), the README (`width="400"`), the playground (`/static/logo-for-dark.svg`).
- **Regenerated from the same outlines**: `images/logo/logo.svg`, `logo-for-light.svg`, `logo-for-dark.svg`, `text.svg` (the ones `site/core/build.ts` copies into `dist/`), `integrations/shared/styles/logo.svg`, and the inline `<Logo>` in `site/shared/components/logo.tsx`.
- **Footprint mark**: same paths as before, taken from `logo.svg`. `logo-for-light.svg` and `logo-for-dark.svg` previously carried the mark ~3.5 units higher than the other files; they now share the one position, so all six files agree.
- **Integrations header** (`integrations/shared/styles/layout.css`): `.bf-header-logo-img` paints `logo.svg` through a fixed box that was tuned for the old 200×46 viewBox; its width is now derived from the new viewBox at the 26px height, so `mask-size: contain` fills the box edge to edge on all 18 integration demo headers.
- **Untouched**: the footprint icon files (`icon.svg`, `<LogoIcon>`), `favicon.svg`, the PNG icons, and the OG image route (it renders page titles in Inter, not the logo).

## Test plan

- [x] Rendered every regenerated asset in light and dark in headless Chromium: the wordmark reads "BarefootJS", no stray outlines, the mark sits vertically centred on the cap height.
- [x] Site header from the local dev server: light, dark (theme class), and mobile (icon only) all render as before with the new wordmark.
- [x] Integrations header box: measures 99.27×26 (ratio 3.818, same as the SVG) and the masked wordmark fills it edge to edge.
- [x] `grep` for `Barefoot.js` across the repo: the only remaining hit is the spelling rule in `.github/workflows/ci-docs.yml`, which lists it as a misspelling.
[...]

PR #2958, opened 2026-09-12
Retry cpanm install in the xslate & mojolicious integrations' Docker images

kfly8:
## Summary

`main`'s `Deploy Sites` workflow failed twice in a row on `deploy-integrations-xslate`
(run [34721646388](https://github.com/piconic-ai/barefootjs/actions/runs/34721646388),
attempts 1 and 2, both after PR #2957 merged). The Docker build's
`cpanm --notest --local-lib=/opt/perl-deps Text::Xslate Plack Starman` step failed
because the CPAN mirror served a truncated/corrupt tarball for `Cpanel-JSON-XS-4.52`:

```
gzip: stdin: not in gzip format
! Failed to unpack Cpanel-JSON-XS-4.52.tar.gz: no directory
! Installing the dependencies failed: Module 'Cpanel::JSON::XS' is not installed
```

This cascades: `Cpanel::JSON::XS` → `JSON::MaybeXS` → `HTTP::Entity::Parser` → `Plack`
→ `Starman`, so the whole install aborts. Neither the merged PR nor any recent change
touches this Dockerfile — it's CPAN mirror flakiness, not a code regression, but the
Dockerfile had no retry so every hit of that flakiness takes `main` red.

pullfrog's review on this PR also flagged that `integrations/mojolicious/` has the
identical unguarded `cpanm` pattern, subject to the same failure mode in the same
workflow — folded into this PR rather than filed as a separate follow-up.

- Wrapped the `cpanm` install in `integrations/xslate/Dockerfile` (build stage),
  `integrations/xslate/Dockerfile.dev`, `integrations/mojolicious/Dockerfile` (build
  stage), and `integrations/mojolicious/Dockerfile.dev` in a 3-attempt retry loop.
- Clear cpanm's download cache (`~/.cpanm`) between attempts so a retry re-fetches
  the tarball instead of re-unpacking the same corrupt file from cache.

## Test plan

- [x] Verified the retry-loop shell syntax with `bash -n`.
- [ ] Could not run a full `docker build` locally (no Docker daemon available in this
      session) — please confirm `deploy-integrations-xslate` and
      `deploy-integrations-mojolicious` go green on this PR's CI run (or the next
      `main` deploy) before merging.

https://claude.ai/code/session_01ERHEWeUqM6YY4V2uzJ4gwa

### DIFF
--- a/integrations/mojolicious/Dockerfile
+++ b/integrations/mojolicious/Dockerfile
@@ -11,7 +11,16 @@ RUN apt-get update \
  && rm -rf /var/lib/apt/lists/*
 # Install CPAN deps into a self-contained prefix so the runtime stage can copy
 # a single directory instead of guessing which /usr/local/... paths exist.
-RUN cpanm --notest --local-lib=/opt/perl-deps Mojolicious
+# Retried: the CPAN mirror occasionally serves a truncated/corrupt tarball,
+# which cpanm treats as a hard failure. Clear its download cache between
+# attempts so a retry re-fetches instead of re-unpacking the same bad file.
+RUN for i in 1 2 3; do \
+      cpanm --notest --local-lib=/opt/perl-deps Mojolicious && exit 0; \
+      echo "cpanm attempt $i failed, retrying..." >&2; \
+      rm -rf /opt/perl-deps ~/.cpanm; \
+      sleep 5; \
+    done; \
+    exit 1
 
 FROM perl:5.38-slim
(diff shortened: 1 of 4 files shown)


kfly8:
Good catch on `integrations/mojolicious/` having the same unguarded `cpanm` call — agreed it's worth the same retry loop, but keeping this PR scoped to the failure that actually broke `main` (xslate). Will follow up separately for mojolicious rather than widen this one.

---

PR #2970, opened 2026-09-13
Fix #2924: CSR template substitutes the SSR noop for a bare signal setter

kfly8:
## Summary

Stacked on #2969 (base of this 2-PR stack for #2924).

Setter-symmetric follow-up to #2969's getter fix: a component prop whose value was a bare — uncalled — reference to a local signal **setter** (`<Display update={setCount} />`, any non-`on*`-prefixed prop name) hits the identical CSR-fresh-mount `ReferenceError` as the getter case, because `csrSubstitute` had no substitution entry for a signal's setter name at all — it leaked into the module-scope `template` lambda the same way a bare getter did before #2969.

**Fix**: a bare setter reference now substitutes to the reference (Hono) adapter's own SSR noop shim (`() => {}`), registered as an `identifier`-kind entry — not `call`-kind like the getter fix: a setter is never itself invoked as a zero-arg accessor the way a getter is, so no thunk-wrapping is needed (confirmed by inspecting Hono's actual generated SSR shim for this shape: `const setCount: (...) => void = () => {}`).

- Adds case E to `csr-template-scope-soundness.test.ts` and matching `csrSubstitute` unit tests.
- Adds a new fixture, `component-prop-bare-setter`, with a deliberately non-event-shaped prop name (`update`, not `onX`) so the CSR template's event-handler-prop filter doesn't dodge the bug by construction.
- Verified clean on Go (real `go run`), ERB, and Mojolicious (no divergence pin needed — unlike the getter case's #2925, the setter shape doesn't hit Go's constructor-baker gap).

## Test plan

- [x] `packages/jsx/src/__tests__/csr-substitute-bare-getter.test.ts` (setter describe block, new)
- [x] `packages/jsx/src/__tests__/csr-template-scope-soundness.test.ts` case E (new)
- [x] `packages/adapter-tests/src/__tests__/csr-conformance.test.ts`, `csr-skip-rot.test.ts`, `client-js-scope.test.ts` — 833 pass, 0 fail
- [x] `coverage-map.test.ts` — fresh after regen
- [x] Go (`go run`, toolchain 1.25.6), ERB, Mojolicious — clean for the new fixture

Note: since this PR is stacked (base is not `main`), `update-fixtures.yml`'s drift job will not run on it per this repo's own CLAUDE.md guidance — `generate-expected-html.ts` and `coverage-map.ts` were run locally in the primary checkout and `git status --short` confirmed empty before this push.

https://claude.ai/code/session_01WS7NbHsxByQGgmwxRvYELn

---
_Generated by [Claude Code](https://claude.ai/code/session_01WS7NbHsxByQGgmwxRvYELn)_


kfly8:
@pullfrog CI is fully green on this PR — could you review it?

---

kfly8:
Thanks for catching this. Fixed in 5988f6b rather than deferring — it's small and directly analogous to the getter-alias fix (#2813): `extractSsrDefaults`'s alias-resolution step now also unions in setter names (kept local to `ssr-defaults.ts` rather than widening the shared `collectAliasableGetterNames`, since that helper also feeds the Go adapter's `rootFieldRef` eligibility set, which has no notion of a setter alias). Added a unit test mirroring the existing getter-alias one.

---

kfly8:
`e2e-laravel` failed with the same `bun install` error as #2969's `e2e-php`: `No version matching "^3.4.0" found for specifier "@csstools/css-calc" (but package exists)`. Identical failure on two different commits/PRs neither touching `package.json` confirms an upstream registry blip, not this PR's. Will re-run the failed job once the in-progress workflow run finishes (GitHub currently refuses a rerun while the run is still active).

---

kfly8:
@pullfrog New commits landed since your last review: fixed the setter-alias SSR stash-seeding gap you flagged, plus rebased onto #2969's updated base after resolving its merge conflict with `main`. CI is green and this is mergeable — could you take another pass on the current head?

---

PR #2972, opened 2026-09-13
Split adapter-tests out of ci.yml's `rest` shard to fix the CI critical path

kfly8:
## Summary

Analysis of `ci.yml`'s wall-clock time (GitHub Actions performance metrics + sampled run/job timings via the Actions API) confirmed `test (rest)`'s "Run unit tests" step as the workflow's sole critical path:

| job | unit-test step time (measured across several runs) |
|---|---|
| `test (rest)` | **~9–10 min** |
| `test (jsx-1)` | ~4 min |
| `test (jsx-2)` | ~4 min |
| everything else (test-ui, e2e-site-core, fixture-hydrate, e2e-site-ui×4) | ≤2.5 min each |

`rest` ran `packages/client packages/test packages/adapter-tests packages/router` as one serial `bun test` invocation. Isolating each package showed `packages/adapter-tests` alone accounts for ~9 of `rest`'s ~10 minutes; `client` + `test` + `router` finish in well under a minute combined. Every other job in the workflow completes in ~4 minutes or less, so `rest` — really, `adapter-tests` — was the one thing keeping the whole workflow around 10–11 minutes instead of closer to the requested 5-minute target.

- Split `packages/adapter-tests` into its own 3-way sharded matrix (`adapter-tests-1/2/3`), reusing the existing size-based bin-packing shard script (the same tool `jsx-1`/`jsx-2` already use), and shrank `rest` to just `client`/`test`/`router`.
- Plain file size under-predicts a handful of `adapter-tests` files whose cost comes from looping the ~360-fixture corpus rather than from source size:
  - `ssr-hydration-contract.test.ts` — 3 full-corpus loops, ~300s alone despite being a mid-sized 12.6KB file
  - `csr-conformance.test.ts` — ~120s alone despite being one of the smallest files (2.2KB)
  - `client-js-scope.test.ts` — ~115s alone (2 full-corpus loops + a nested per-child-component pass)

  Added `WEIGHT_OVERRIDES` to `scripts/ci/shard-test-files.ts` so the bin-packer uses measured weight instead of byte size for these three. This balanced the three new shards to within ~1.25x of each other (vs. a naive size-based split's ~3.2x spread between the fastest and slowest shard).

## Expected impact

With `adapter-tests` parallelized across 3 shards each roughly matching `jsx-1`/`jsx-2`'s own runtime, the workflow's critical path should now be bounded by the jsx shards (~4 min) plus job start/build overhead, rather than by `rest`'s previous ~10 min serial run.

## Test plan

- [x] Verified the 3 adapter-tests shards partition all 23 test files exactly once (no gaps, no duplicates)
- [x] Verified `rest` (`client`+`test`+`router`) still passes standalone (822 tests, 95 files)
- [x] Verified `ci.yml` YAML parses correctly
- [x] Confirmed on this PR's own CI run ([run 34743435489](https://github.com/piconic-ai/barefootjs/actions/runs/34743435489)): all jobs green, total wall time from push to last job completing dropped from ~10–11 min to **~4m41s** (created 06:43:04 → last job done 06:47:45). `test (rest)` fell from ~10 min to 1m06s; the new shards (`adapter-tests-1/2/3`: 4m36s / 3m43s / 3m02s) are now roughly on par with `jsx-1`/`jsx-2` (4m32s / 4m17s), which are the new critical path.

https://claude.ai/code/session_015pGL18oZ1HEymCVqUGwWuD

---
_Generated by [Claude Code](https://claude.ai/code/session_015pGL18oZ1HEymCVqUGwWuD)_


kfly8:
Thanks for the review. On the nitpick (`WEIGHT_OVERRIDES` paths not existence-checked): leaving as-is — same pattern the `jsx-1`/`jsx-2` split already relies on (no existence check on its inputs either), and the failure mode is a silent fallback to byte-size weighting (wall-clock regression, not correctness) rather than a wrong result. Happy to add a startup assertion in a follow-up if this file set churns enough to make it worth the extra code.

---

PR #2987, opened 2026-09-14
fix: module-level helper functions incorrectly inlined in CSR client JS (#2986)

### LINKED ISSUE #2986: CSR: module-level helper functions referenced from a component are inlined incorrectly (lose param scope / break async boundary)
## Summary

When a `'use client'` component calls a plain helper function defined at module
top level (outside the component body), the compiler appears to try inlining
that helper's body into the generated client code, and gets it wrong in two
ways:

1. A **synchronous** top-level helper that takes a parameter gets inlined with
   the parameter reference left dangling, producing a runtime
   `ReferenceError` instead of a working closure.
2. An **async** top-level helper gets inlined such that its `await` ends up
   outside of an `async` function, breaking the build at the esbuild step.

Both were found using the CSR Adapter (`@barefootjs/client/csr-adapter`); I
haven't checked whether SSR adapters hit the same code path.

## Environment

- `@barefootjs/{client,jsx,vite,router}` `0.35.7`
- Adapter: `CSRAdapter` (`@barefootjs/client/csr-adapter`) via `@barefootjs/vite`'s `barefoot()`
- Repro project: a Vite CSR app, `vite build`

## Repro 1 — sync helper, `ReferenceError` at runtime

```tsx
'use client'
import { createMemo, createSignal } from '@barefootjs/client'

type Doc = { title: string }

// Module-level helper, NOT inside the component.
const score = (doc: Doc, term: string): number => (doc.title.includes(term) ? 1 : 0)

function Search() {
  const [term, setTerm] = createSignal('')
  const [docs, setDocs] = createSignal<Doc[]>([{ title: 'hello' }])

  const results = createMemo(() => docs().map((doc) => ({ doc, s: score(doc, term()) })))

  return <div>{results().length}</div>
}

export default Search
```

`vite build` succeeds, but the browser throws at mount:

```
ReferenceError: doc is not defined
```

Looking at the compiled output, `score`'s body was spliced directly into the
component's init function, with `doc` referenced but never bound in that
scope — e.g. (from the real case that surfaced this, a search box scoring
function):

```js
function Kt(e,t={}){if(!e)return;e.getAttribute("bf-s"),doc.title.toLowerCase(),doc.tags.join(" ").toLowerCase(),doc.text.toLowerCase();const n=(u,p)=>{ /* the correctly-scoped version also present */ ...
```

Moving `score` to be a local function declared inside `Search()` fixes it —
no other change needed.

## Repro 2 — async helper, build fails

```tsx
'use client'
import { createSignal, onMount } from '@barefootjs/client'

// Module-level async helper.
const loadData = async (): Promise<string[]> => {
  const res = await fetch('/data.json')
  return res.json()
}

function Widget() {
  const [items, setItems] = createSignal<string[]>([])
  onMount(() => {
    void loadData().then(setItems)
  })
  return <div>{items().length}</div>
}

export default Widget
```

`vite build` fails at the esbuild transform step:

```
[vite:esbuild] Transform failed with 1 error:
.../Widget.tsx:8:19: ERROR: "await" can only be used inside an "async" function
```

The error points at a location inside compiler-generated code where
`loadData`'s body (including its own `await`) was inlined into a
non-`async` context.

Moving `loadData` to be a local function/closure declared inside `Widget()`
(or just inlining the `fetch().then()` chain manually) avoids the error.

## Workaround

Define every helper a component calls as a local function/const **inside**
the component body, even when it doesn't close over any component state.
Don't reference module-top-level functions from component code.

## Notes

- Found while migrating a real site's search island
  (Solid → BarefootJS, CSR Adapter) — the original code had `loadIndex`
  (async, module-level) and `scoreDoc` (sync, module-level), both moved
  inside the component to fix.
- I haven't dug into `packages/jsx`'s compiler internals to find the exact
  inlining pass responsible; happy to help narrow it down further if useful.

kfly8:
## Summary

Fixes #2986.

A `'use client'` component calling a plain helper declared at module top
level as `const name = (params) => { ... }` (arrow form) crashed two ways:

1. A sync helper's inner declarations leaked into the component's init
   function with their parameter references unbound → `ReferenceError` at
   mount.
2. An async helper's inner `await` leaked into a non-async scope →
   esbuild parse failure at build time.

The equivalent `function name(params) { ... }` declaration form was
unaffected.

### Root cause

`packages/jsx/src/analyzer.ts`'s module-level `visit()` walk descended
into the body of a module-level arrow-valued `const`, collecting that
body's own inner `const`/`function` declarations as if they were
declarations of the file itself — hoisted into both the component's init
function and the marked template. `visitComponentBody` already had the
guard that prevents exactly this; `visit` never did. The `function`
declaration form was immune only because the module-level branch
`return`s immediately after collecting it, so `visit` never walks its
internals.

### Fix

One shared `isFunctionScope` predicate (arrow / function expression /
function declaration / method / accessors / constructor), called by both
`visit`'s module-level recursion and `visitComponentBody`'s recursion —
replacing two independently-drifted implementations with one. `visit`
keeps two measured, load-bearing carve-outs: `FunctionDeclaration`
children must still be visited (the only walk that reaches a
component-body `function` declaration), and the target component's own
arrow body must still be descended into to reach nested `function`
declarations.

Design by Claude Opus (background investigation, verified by actually
executing the compiler against both repros rather than static reading —
confirmed the issue's own minimal sync repro was over-minimised and
doesn't reproduce as written; the block-bodied variant in the new test
file does). Full design write-up available on request.

A related, separate defect found during the investigation — a
three-way disagreement between `compute-scope.ts`, `relocate.ts`, and
`compute-inlinability.ts` about whether a module-scope arrow constant is
safe to reference by name from a template lambda — produces a different,
quieter symptom (a silently empty SSR seed, not a crash) and is filed
separately as #2988 rather than bundled here, to keep this fix's output
churn at zero.

## Testing

- `packages/jsx/src/__tests__/module-helper-scope-2986.test.ts` — new
  compiler unit test file. 11 cases: sync/async/object-method/nested-
  function leak shapes, a false-BF011 regression, the two carve-outs a
  naive fix would break, and parity assertions between the `const` arrow
  and `function` declaration forms. Verified to fail on the pre-fix tree
  (10/11 directly; the 11th — component-body object-method widening —
  fails against the true pre-fix baseline specifically).
- `packages/adapter-tests/fixtures/module-helper-async.ts` — new CSR
  conformance fixture. Pre-fix, the duplicated `await` makes the compiled
  client JS fail to parse entirely (verified: `"await" can only be used
  inside an "async" function"`, the exact esbuild failure from the issue).
  `expectedHtml` generated from the Hono reference adapter; confirmed
  zero drift across the rest of the fixture corpus.
- Full `packages/jsx/src/__tests__` suite: **3440 pass, 0 fail** (27
  skip, 3 todo — pre-existing).
- `bun run build` at repo root: clean.

## Test plan

- [x] Compiler unit test pinning the corrected scope classification
- [x] CSR conformance fixture covering the async module-helper shape
      from the issue's repro 2
- [x] `bun run packages/adapter-tests/scripts/generate-expected-html.ts`
      run — `git status --short` clean except this fixture's own
      `expectedHtml`
- [x] Full `packages/jsx` unit suite green, zero regressions
- [x] Changeset added

---

https://claude.ai/code/session_019iWh5jUeHj5h7YtkQnWDLv


kfly8:
`check` (changeset gate) is failing because this PR currently only contains WIP investigation/experimentation commits (scratch repro harnesses + an in-progress scope-boundary prototype in `analyzer.ts`) with no changeset yet. This is expected at this stage — a `.changeset/*.md` entry will be added once the fix design is finalized and the real change lands, which will resolve this check on the next push. No action needed in the meantime.

---

kfly8:
Status: all CI checks are green (46/46) and Pullfrog has reviewed twice with no issues found. The PR is mergeable.

Summary:
- Fix implemented and verified (compiler unit suite: 3440 pass / 0 fail, zero regressions; new 11-case regression test file; new CSR conformance fixture that reproduces the exact pre-fix esbuild failure).
- Changeset added.
- Follow-up known-limitation filed separately as #2988 (the scope-classification asymmetry that causes a silent empty SSR seed — a different, lower-severity defect, deliberately not bundled here to keep this fix's output churn at zero).

Leaving this for @kfly8 to merge at your discretion.

---

PR #2990, opened 2026-09-14
Fix(jsx): provision `batch` runtime import for cross-file reactive factories

### LINKED ISSUE #2985: compiler: cross-file reactive factory leaves batch() unimported in generated client JS
### What happened?

When a cross-file reactive factory imports and calls `batch` from `@barefootjs/client`, the compiler inlines the factory and provisions other runtime primitives such as `createSignal`, but does not provision `batch`.

Compilation succeeds with no diagnostics. The generated client module contains a bare `batch(...)` call but its `@barefootjs/client/runtime` import does not include `batch`, causing `ReferenceError: batch is not defined` when the callback runs.

### Component source that reproduces it (if handy)

```tsx
// useCounter.tsx
'use client'

import { batch, createSignal } from '@barefootjs/client'

export function useCounter() {
  const increment = () => batch(() => setCount(count() + 1))
  return { count, increment }
}

// Counter.tsx
'use client'

import { useCounter } from './useCounter'

export function Counter() {
  const { count, increment } = useCounter()
  return <button onClick={increment}>{count()}</button>
}
```

### Expected vs. actual (if known)

Expected: the generated client module imports both `batch` and `createSignal` from `@barefootjs/client/runtime`, and clicking the button increments the count.

Actual: compilation reports no error and emits code equivalent to:

```js
import { createSignal, /* other helpers */ } from '@barefootjs/client/runtime'

const [count, setCount_bf0] = createSignal(0)
const increment = () => batch(() => setCount_bf0(count() + 1))
```

Clicking the button throws:

```text
ReferenceError: batch is not defined
```

On current main, `batch` is exported by the client runtime but is absent from `RUNTIME_IMPORT_CANDIDATES` in `packages/jsx/src/ir-to-client-js/imports.ts`.

### Environment (if relevant)

- BarefootJS version: `main` at `8cc668200a6db1d56623546d1a71b7edc86b447e` (also reproduced with `0.35.7`)
- Adapter: compiler-level reproduction with `TestAdapter`; originally observed with Go Template
- Runtime: Bun `1.4.2`; browser runtime failure
- OS: macOS / Darwin `25.6.0`

kfly8:
## Summary

- `batch`, exported by `@barefootjs/client`, was missing from `RUNTIME_IMPORT_CANDIDATES` (`packages/jsx/src/ir-to-client-js/imports.ts`) — the list `detectUsedImports` scans to decide which `@barefootjs/client/runtime` helpers a generated client bundle needs to import.
- A cross-file reactive factory that imports and calls `batch` (e.g. `useCounter.tsx` calling `batch(() => setCount(...))`, imported by a consumer component) compiled with no diagnostics, but the emitted client module's runtime import never included `batch` — a bare `batch(...)` call in the bundle threw `ReferenceError: batch is not defined` at hydration.
- Added `batch` to the candidate list, right next to the other core reactive primitives.

## Test plan

- Added `M17b` to `packages/jsx/src/__tests__/reactive-factory-cross-file.test.ts`, mirroring the existing `M17` (`onMount`) regression pin: a cross-file `useCounter` factory that imports and calls `batch`, asserting the generated client module's `@barefootjs/client/runtime` import includes `batch`.
- Verified the new test fails without the fix (`ReferenceError`-shaped repro: `batch` absent from the emitted import) and passes with it.
- `bun test packages/jsx/src/__tests__/reactive-factory-cross-file.test.ts` — 40 pass.
- `bun test packages/jsx/src/__tests__` — full suite green aside from two pre-existing failures in `primitive-resolver-alias.test.ts` unrelated to this change (verified they fail identically on `main`).

Closes #2985

https://claude.ai/code/session_01TWia6nuNDPEB5QUToHSAyu

---
_Generated by [Claude Code](https://claude.ai/code/session_01TWia6nuNDPEB5QUToHSAyu)_


kfly8:
@kfly8 this is ready for review after the Polish pass.

**What**: Fixes #2985 — `batch` (from `@barefootjs/client`) was missing from `RUNTIME_IMPORT_CANDIDATES` in `packages/jsx/src/ir-to-client-js/imports.ts`, the list `detectUsedImports` scans to decide which runtime helpers the generated client bundle needs. A cross-file reactive factory that called `batch(...)` compiled clean but the emitted `@barefootjs/client/runtime` import omitted it, throwing `ReferenceError: batch is not defined` at hydration. The fix adds `batch` to that list (every other reactive primitive was already there).

**How to verify**:
- Code review + simplify pass on this diff turned up nothing to change — it's a one-line addition to an existing candidate list plus a regression test, already following the established pattern for every other runtime helper in that array. No fixup commits were needed.
- Run the new regression test: `bun test packages/jsx/src/__tests__/reactive-factory-cross-file.test.ts` — this includes the new `M17b` case (`batch is provisioned from usage for a cross-file factory (#2985)`).
- Optionally reproduce the original bug by checking out the parent commit (`git show HEAD~2:packages/jsx/src/ir-to-client-js/imports.ts` to diff) and re-running the same test — `M17b` fails without the fix.

**Expected result**: `bun test packages/jsx/src/__tests__/reactive-factory-cross-file.test.ts` reports `40 pass, 0 fail`. The `M17b` assertion specifically confirms the generated client JS's `@barefootjs/client/runtime` import line now includes `batch`, matching `/import\s*\{[^}]*\bbatch\b[^}]*\}\s*from\s*'@barefootjs\/client\/runtime'/`.

---

PR #2983, opened 2026-09-13
feat(spring): add integrations/spring — Spring Boot Pebble demo (#2101)

kfly8:
## Summary

Stacked on #2982 (Phase 6 docs). Adds `integrations/spring/` — a full-parity Spring Boot demo app — completing the last deliverable listed on the Pebble adapter tracking issue (#2101):

> Integration example — recommend Spring Boot for reach (accept the image size), `integrations/spring/`; standard demo set + blog; shared e2e green; SSE must not block the servlet thread pool (use async/SseEmitter — same lesson as fastapi's asyncio SSE).
> Wiring: dev proxy, docker-compose, README, `ci-java.yml` (setup-java + gradle/maven cache), deploy job; container listens on 8080 (#2080 postmortem).

### Pebble Java runtime additions (`packages/adapter-pebble/java`)

All additive — existing constructors/behavior unchanged, `gradle test` stays 512/512 green:

- **`Bf.java`**: `register_script`/`register_preload` now record into a shared, deduped list, plus a `scripts()` readback method — mirrors the Rust runtime's `Arc<RenderSession>`-shared `scripts`/`script_seen`; a host application has no other way to recover the accumulated `<script>`/`<link rel="modulepreload">` tags for a render tree. `Bf.newRoot(scopeId, sibling[, rootProps])` lets several independently-rendered top-level islands on one page (a region-shell layout) share one accumulator. A new `(rootScopeId, engine, manifest, rootProps)` constructor seeds a root render's `bf-p` hydration payload, matching `render.rs`'s `render_root` — without it, a `@client` root re-hydrates its reactive state from nothing rather than the props it was actually rendered with, silently discarding an SSR'd list/object prop on hydration (a real bug caught via an actual browser, not just a code read).
- **`ChildMeta.java` / `DeriveStashFromDefaults.java`**: made public so a host app can use them outside the package.
- **`ManifestLoader.java`** (new): `load()` extracted out of `Main` (same behavior, now shared) plus `loadFromBuildManifest()`, a faithful Java port of the Rust runtime's `register_components_from_manifest` — registers every compiled component from a real `bf build` `manifest.json` in one call, the production path `integrations/spring`'s `Render` class uses.

### `integrations/spring/`

[...]

Issue #2994, opened 2026-09-14
Non-JS SSR adapters silently render empty for a module-scope helper (function or arrow) called in a template position


kfly8:
## Summary

Split out of #2988 during fixture verification for that issue's fix.

A module-top-level helper — either declaration form, `function fmt(s) { ... }` or the arrow form `const fmt = (s) => ...` — that is safe to reference by bare name from a component's template (it closes over nothing but its own parameters) renders correctly on the Hono reference adapter (real JS execution) but does NOT render correctly on any non-JS "Marked Template" adapter. **All 8 non-Hono adapters were directly verified against their real toolchains in this investigation** (ERB/Ruby, Jinja/Python, minijinja/Rust, Twig/PHP, Blade/PHP, Go, Text::Xslate/Perl, Mojolicious/Perl).

The failure mode differs by adapter family:
- **ERB, Jinja, minijinja (Rust), Twig, Blade, Text::Xslate**: silently render the slot **empty**, with no compile-time diagnostic.
- **Go**: `html/template` treats the bare reference as a struct-field lookup and the generated program **crashes at render time** (`go run` exits non-zero: `template error: ... executing "Widget" at <.Fmt>: can't evaluate field Fmt in type main.WidgetProps`).
- **Mojolicious**: `Mojo::Template` runs under Perl `strict` and the bare `$fmt` **crashes at render time** too, with a different error: `Global symbol "$fmt" requires explicit package name (did you forget to declare "my $fmt"?)`. (ERB was ported FROM this adapter but does NOT share this crash — Ruby's ERB binding just stringifies a missing local to empty instead.)

Either way this is a silent divergence at the TS/JSX **compile** layer (`known-limitation` + `bug` tier per this repo's compatibility policy: the compiler itself emits no diagnostic for a shape it cannot faithfully lower on these targets) — the failure only surfaces once the generated Go/Ruby/Python/Perl/PHP program actually runs.

## Repro

```tsx
'use client'
import { createSignal } from '@barefootjs/client'

const fmt = (s: string) => s.toUpperCase()
// or: function fmt(s: string) { return s.toUpperCase() }

export function Widget({ label }: { label: string }) {
  const [count, setCount] = createSignal(0)
  return (
    <div>
      <span>{fmt(label)}</span>
      <button onClick={() => setCount(count() + 1)}>{count()}</button>
    </div>
  )
}
```

- Hono (real JS execution): `<span>HI</span>` — correct.
[...]

PR #2995, opened 2026-09-14
feat(adapter-pebble): production manifest loading + script/preload accumulation (#2101)

kfly8:
## Summary

Stacked on #2982 (Phase 6 docs). Split out of the original #2983 (per review discussion — the combined diff was too large for pullfrog's automated review to complete; splitting on a natural seam keeps each PR independently reviewable and small).

Additive Java runtime changes to `packages/adapter-pebble/java` needed by a real production host (the `integrations/spring` example, stacked on this PR). Existing constructors/behavior unchanged, `gradle test` stays 512/512 green.

- **`Bf.java`**: `register_script`/`register_preload` now record into a shared, deduped list, plus a `scripts()` readback method — mirrors the Rust runtime's `Arc<RenderSession>`-shared `scripts`/`script_seen`; a host application has no other way to recover the accumulated `<script>`/`<link rel="modulepreload">` tags for a render tree. `Bf.newRoot(scopeId, sibling[, rootProps])` lets several independently-rendered top-level islands on one page (a region-shell layout) share one accumulator. A new `(rootScopeId, engine, manifest, rootProps)` constructor seeds a root render's `bf-p` hydration payload, matching `render.rs`'s `render_root` — without it, a `@client` root re-hydrates its reactive state from nothing rather than the props it was actually rendered with, silently discarding an SSR'd list/object prop on hydration (a real bug caught via an actual browser during the Spring integration's own build).
- **`ChildMeta.java` / `DeriveStashFromDefaults.java`**: made public so a host app can use them outside the package.
- **`ManifestLoader.java`** (new): `load()` extracted out of `Main` (same behavior, now shared) plus `loadFromBuildManifest()`, a faithful Java port of the Rust runtime's `register_components_from_manifest` — registers every compiled component from a real `bf build` `manifest.json` in one call, the production registration path a real host uses.

## Test plan

- [x] `gradle test` (`packages/adapter-pebble/java`) — 512/512, before and after
- [x] `bun run lint` — clean

https://claude.ai/code/session_017mtcDc4LcsxrtiRc9TLJPR

---
_Generated by [Claude Code](https://claude.ai/code/session_017mtcDc4LcsxrtiRc9TLJPR)_


kfly8:
The `lockfile` check failed on `ci-compat.yml`'s `support matrix freshness > committed ui/support-matrix.lock.json matches a fresh computeSupportMatrix()` test. Investigated — this doesn't look like this PR's fault:

- This PR only touches `packages/adapter-pebble/java/**` (new Java source + test files); no TS/compat-layer code changed.
- Reproduced locally on this exact commit (000113f): `bun test packages/compat/__tests__/support-matrix.test.ts` — all 8 tests pass (with a longer timeout; the default 5s times out in this sandbox purely on `computeSupportMatrix()`'s own runtime, unrelated to correctness).
- Regenerated `ui/support-matrix.lock.json` via `bun run support-matrix:lock` locally and diffed against the committed file — zero drift.
- CI's own failure completed in 106ms (a real comparison mismatch, not a timeout), which I could not reproduce locally at all — points to non-determinism in `computeSupportMatrix()` itself rather than genuine content drift from this PR.

Re-ran the failed job once (`rerun_failed_jobs` on run 34844258453) to confirm the flake theory, per this repo's CI posture.

---

kfly8:
Follow-up on the `lockfile` failure: the re-run failed identically (same off-by-one `pass`/`total` diff for `pebble` across many constructs). I dug further and now have strong evidence this is **not** reproducible from this PR's diff, though I can't fully explain the CI-side result:

**Why this PR's diff structurally cannot cause it**: `computeSupportMatrix()` (`packages/compat/src/support-matrix.ts`) is a pure join over the *committed* `packages/adapter-tests/coverage-map.json` (denominators — `total`) and each adapter's *static* `conformancePins`/`renderDivergences` exports (numerators — `pass`/`gaps`). I grepped `packages/compat/src/` for any Java/Gradle coupling — none exists; nothing here shells out to `java`/`gradle` or reads anything under `packages/adapter-pebble/java/`. This PR only touches files under that directory.

**Three independent clean-room reproductions, all green**:
1. A fresh `git worktree` + fresh `bun install` + fresh adapter builds on the **base branch** (`claude/charming-faraday-dp4a4g-6-docs`, pre-dating this PR) — `bun test packages/compat` passes fully.
2. The same clean-room setup on **this PR's branch** (`claude/charming-faraday-dp4a4g-7a-pebble-runtime-api` @ 000113f) — `bun test packages/compat/__tests__/support-matrix.test.ts` passes fully (8/8), including the exact test CI fails.
3. The same worktree running CI's **exact command**, `bun test packages/compat` (all 12 files, not just the one), to rule out cross-test-file state pollution — 623/623 pass.

Also confirmed `packages/adapter-tests/coverage-map.json` and `ui/support-matrix.lock.json` are byte-identical between my local checkout and `origin/claude/charming-faraday-dp4a4g-7a-pebble-runtime-api` (no push drift).

I don't have a way to reproduce or further root-cause this from here, and I'd rather not blindly edit `ui/support-matrix.lock.json` to match a CI-computed diff I can't explain — that risks masking a real, separate problem or just reverting on the next run if it's genuinely CI-side non-determinism. `test`, `naming`, `check`, and `pullfrog` are all green on this PR; only `lockfile` is red.

@kfly8 — flagging for a decision here rather than guessing further: worth a look at the actual runner, or should I try something else (e.g. merging the base branch in to force a clean re-checkout, in case something about this branch's git state itself is the issue)?

---

kfly8:
Follow-up: root-caused this, and it wasn't a flake or CI non-determinism after all.

`main` had drifted ~60 commits ahead of this stack's merge-base (`5a3cad8`), several of which regenerated `packages/adapter-tests/coverage-map.json` / `ui/compat.lock.json` / `ui/support-matrix.lock.json` for unrelated new fixtures added on `main` after this stack branched off. This stack's own branches were never synced with those commits, so every branch's committed `support-matrix.lock.json` was computed against a now-stale `coverage-map.json` — CI's fresh `computeSupportMatrix()` run (using the *current* generator + *current* `coverage-map.json`) legitimately disagreed with the committed file, while every local reproduction attempt (including the three clean-room ones above) used the same stale, self-consistent branch state as CI's checkout — so of course they all "passed": there was nothing to disagree with until `main`'s newer data actually entered the branch.

Fix: merged `origin/main` into the whole stack (branches 4/5/6/7a/7b, bottom-up), then regenerated `ui/compat.lock.json` and `ui/support-matrix.lock.json` via their real generators (`bun run compat:lock` / `bun run support-matrix:lock`) rather than trusting git's textual merge of the JSON files — the merge itself was clean but left `support-matrix.lock.json` internally inconsistent (other adapters' cells reflected `main`'s newer data, pebble's own cells stayed stale). Pushed the regenerated lockfile on the Phase 5 branch (`b072cc3`); this PR now carries that fix through its base chain.

---

kfly8:
Update: the `lockfile` job failed again after the stack was rebased onto `main` (this time also reproducing on #2982), with a *different* root cause than the drift explained above.

`ci-compat.yml`'s `lockfile` job does a fresh `bun install` + explicit `bun run --filter '@barefootjs/<pkg>' build` for all 9 dist-based adapters *before* running `bun test packages/compat` / `bun run compat:lock`. My local regeneration after the rebase reused this session's existing `packages/adapter-pebble/dist/**`, which had been built *before* the `render-divergences.ts` fix for `module-const-arrow-helper` (#2994) landed on this branch — so `bun run compat:lock`/`support-matrix:lock` picked up a stale Pebble `dist/index.js` missing that divergence declaration, and committed an inconsistent lockfile.

Fixed by rebuilding every dist-based adapter fresh (mirroring the CI step exactly), then regenerating both lockfiles against that clean build. `bun test packages/compat` — **631/631 pass** — and re-pushed. Lesson for next time: after touching adapter source, rebuild that adapter's `dist/` before regenerating any lockfile that reads it, not just before pushing.

---

Issue #2073, opened 2026-07-02
Template adapters: value-producing `.map(cb)` (e.g. `.map(cb).join(sep)`) can't lower (BF101)


kfly8:
Split out from the known-limitations catalog #2069 ("Value-producing `.map(cb).join(sep)` can't lower (BF101)").

## Limitation

A `.map()` whose callback returns a **value** (string / number / object), consumed as an array — most commonly chained into `.join(sep)` — is refused on the template-string adapters (Go + Mojo + Xslate): `map` sits in `UNSUPPORTED_METHODS` in `packages/jsx/src/expression-parser.ts` and is deliberately excluded from `CALLBACK_METHODS` ("a value-returning `.map` has no template lowering"). Only the JS-runtime adapters (Hono / h3 / Elysia) evaluate it. The JSX-returning `.map()` lowers separately as an `IRLoop` and is unaffected.

Surfacing shape (the `@barefootjs/router` blog showcase, #1938/#1939):

```tsx
meta={`${p.date} · ${p.tags.map((t) => `#${t}`).join(' ')}`}
```

Workaround was precomputing the string upstream and passing it as a plain prop.

## Fix

The #2018 runtime evaluator already generalizes exactly this shape — `.flatMap(proj)` lowers by serializing the projection body and projecting per element (`flat_map_eval`). A value-producing `.map(cb)` is the same operation without the one-level flatten:

- add `map` to `CALLBACK_METHODS` (the JSX form is an IRLoop before recognition, mirroring `flatMap`),
- add a `map_eval` arm to each template adapter's `callbackMethod` dispatch (`bf->map_eval` / `$bf.map_eval` / `bf_map_eval`),
- add the corresponding runtime helpers (Perl `BarefootJS::Evaluator::map_json`, Go `MapEval`),
- keep the `UNSUPPORTED_METHODS` refusal for the fall-throughs (a bare `arr.map` reference or a function-reference callback), like `reduce` does.

Chaining composes through the existing `.join` lowering (`join(' ', @{…})` / `bf_join (…)`).

Refs #2069.

PR #2998, opened 2026-09-14
Fix module-scope arrow-valued const helper's empty CSR template seed

### LINKED ISSUE #2988: CSR: module-scope arrow-valued const helper silently emits an empty SSR seed in the template lambda (compute-scope/relocate/compute-inlinability disagree)
## Summary

Split out of #2986 during root-cause investigation for that issue's crash fix (the analyzer recursion-guard fix in #2987 does **not** touch this).

A module-top-level helper declared as `const name = (params) => expr` (arrow form) that is safe to reference from a component's CSR template lambda — i.e. it closes over nothing but its own parameters — is not actually referenced there. Instead the template's slot silently renders `${''}` (empty), even though SSR renders the correct value. The equivalent `function name(params) { ... }` declaration form works correctly.

This is a silent divergence (`known-limitation` + `bug` tier per this repo's compatibility policy), not a crash — hence tracked separately from #2986's `ReferenceError`/esbuild-parse-failure crashes, which #2987 fixes.

## Repro

```tsx
'use client'

const fmt = (s: string) => s.toUpperCase()

export function Widget({ label }: { label: string }) {
  return (
    <div>
      <span>{fmt(label)}</span>
      <button onClick={() => setCount(count() + 1)}>{count()}</button>
    </div>
  )
}
```

SSR (Hono reference adapter) renders the correct uppercased value. The CSR-emitted template lambda renders an empty string for the `fmt(label)` slot instead:

```js
// function fmt(s) { return s.toUpperCase() }          <-- correct SSR seed
hydrate('Widget', { …, template: (_p) => `<div>…${escapeTextOrMarkup(fmt(_p.label))}…` })

// const fmt = (s: string) => s.toUpperCase()          <-- silently empty
hydrate('Widget', { …, template: (_p) => `<div>…${''}…` })
```

## Root cause

Three independent, disagreeing implementations of the same question — "is this module-level declaration safe to emit at module scope / reference by name from a template lambda?" — this repo's CLAUDE.md explicitly calls out "one decision, two implementations, no test comparing them" as a recurring defect class to avoid reintroducing, and this is a three-way instance of it:

1. `packages/jsx/src/ir-to-client-js/compute-scope.ts`'s `classifyConstant` answers it correctly for `function` declarations (a forward-reachability fixpoint over `ctx.localFunctions`) but drops a module-level arrow `const` to `'init'` scope unless it's a reassignment target (`c.isModule && initStmtAssigned.has(c.name)`, a narrow #933 carve-out) — there is no fixpoint-reachability equivalent for constants.
2. `packages/jsx/src/ir-to-client-js/compute-inlinability.ts`'s `classifyConstantInitial` answers it again for functions (`fn.isModule && !functionReferencesDeclaredName(...)`), but for constants unconditionally returns `{ kind: 'arrow-literal' }` on `c.containsArrow`, ignoring `c.isModule` entirely. This feeds `unsafeLocalNames`, which is the direct cause of the `${''}` — `html-template.ts`'s `transformExpr` trips `UNSAFE_TEMPLATE_EXPR` → `EMPTY_MARKUP` for the substituted expression.
3. `packages/jsx/src/relocate.ts`'s `buildRelocateEnvFromFields` answers it a third time from the raw static `c.isModule` flag alone (`'module-local'` binding kind), independent of what (1) or (2) actually decided.

## Suggested fix shape (from the #2986 design investigation)

One producer, three consumers, instead of three independent answers:

1. In `compute-scope.ts`, fold constants into the *existing* function reachability fixpoint instead of pre-classifying them: seed with signals/memos/props, add module-level value-bearing non-reassigned non-JSX constants (alongside functions) to `pending`, strip each constant's own arrow parameter names from its reference set (mirroring what already happens for `fn.params`), and let survivors land at `'module'`.
2. Export/store that answer (e.g. a computed `isModuleScopeSafe?: boolean` field on `ConstantInfo`, mirroring how `FunctionInfo` already carries `isModule`/`isAsync`/`declarationKind`) and have `compute-inlinability.ts` read it rather than re-derive `arrow-literal` unconditionally — add a `module-scope-safe` `ConstantInlinability` variant so a module-scope arrow const is referenced **by name** from the template lambda.
3. Have `relocate.ts`'s `buildRelocateEnvFromFields` read the same field rather than raw `c.isModule`.
4. Add a cross-site test asserting that, for the same source, `compute-scope`'s constant scope, `compute-inlinability`'s constant classification, and `relocate`'s binding kind all agree for a given name — and that the `function` form and `const` arrow form of the same helper yield the same triple. That test is what would have caught this drifting apart.

Two implementation traps to watch for (measured during the #2986 investigation):

- `packages/jsx/src/ir-to-client-js/emit-module-level.ts` emits `var ${c.name} = ${c.name} ?? ${c.value}`. For an arrow value this needs parenthesization (`var f = f ?? (${c.value})`) — `var f = f ?? (s) => …` is a **SyntaxError** as written (the `??` operand swallows into the arrow's parameter list). Only wrap when `c.containsArrow`, so existing fixtures with parenthesised object literals don't churn.
- `ConstantInfo.freeIdentifiers` is not scope-aware inside an arrow's block body (e.g. `const f = (a) => { const b = a * 2; return b }` yields `freeIdentifiers: ["b"]` — the param is correctly excluded, the body-local isn't). Feeding this into the fixpoint is fail-safe (a body-local colliding with a signal name just demotes to init scope) but imprecise; that's an acceptable, separately-trackable imprecision, not a blocker for this fix.

## Test / fixture plan

Per CLAUDE.md, this needs a fixture asserting the CORRECT output (a module arrow helper applied to a prop, non-empty SSR seed) plus a pin on the broken side carrying this issue's URL — and since the Hono reference adapter itself currently emits the broken `${''}` for this shape, the fixture's `expectedHtml` needs to be **hand-authored** to the correct value and registered in `generate-expected-html.ts`'s `SKIP_AUTO_UPDATE` until this graduates (per CLAUDE.md's "the one exception is where Hono is itself the broken side").

## Related, same family, also out of scope for #2987

Found during the same investigation, not yet filed as separate issues:

- **Module-level `if`/block statement leak**: `if (cond) { const flag = 1 }` at module level leaks `flag` into module constants, and the `if` statement itself disappears from the template. Not a function-scope boundary, so #2987's analyzer fix doesn't touch it.
- **Declaration-order gate**: a module `const` declared *after* the component that references it is never collected (`analyzer.ts`'s `!ctx.componentNode` gate), so both the client bundle and the SSR template reference an undeclared name → `ReferenceError`. The `function` form has no such gate and works. Arguably the most user-visible remaining `function`-vs-`const arrow` asymmetry after #2987.

kfly8:
## Summary

Fixes #2988: a module-top-level helper declared as `const name = (params) => expr`
(arrow form) that closes over nothing but its own parameters silently rendered an
empty CSR template seed (`${''}`) instead of being referenced by bare name — the
equivalent `function name(params) { ... }` declaration form already worked correctly.

Per the issue's root-cause analysis, this was a "one decision, three independent
implementations" bug:

1. `compute-scope.ts`'s forward-reachability fixpoint decided module-vs-init
   emission scope for **functions** but had no equivalent for **constants** — an
   arrow-valued module const fell straight to `'init'` scope regardless of whether
   it was actually safe to hoist.
2. `compute-inlinability.ts`'s `classifyConstantInitial` returned `arrow-literal`
   (template-unsafe) **unconditionally** for every arrow-valued constant, never
   consulting `isModule` at all — this is the direct cause of the empty CSR seed.
3. `relocate.ts`'s `buildRelocateEnvFromFields` answered the same question a third
   way, from the raw `isModule` flag alone.

## Fix

- **`compute-scope.ts`**: folds arrow-valued module-level constants into the *same*
  forward-reachability fixpoint already used for module-level functions (seeded with
  signals/memos/props/init-scope constants; a constant/function that transitively
  references anything in that set demotes to `'init'`; survivors are `'module'`).
  Scoped specifically to `c.isModule && c.containsArrow` — not every value-bearing
  module constant — to keep this fix targeted at the reported divergence rather than
  broadening emission-scope placement for shapes the issue didn't report.
- **`compute-inlinability.ts`**: `classifyConstantInitial` now calls
  `computeDeclarationScopes` (the same producer `compute-scope.ts` already computes)
  and reads its verdict for arrow-valued constants instead of re-deriving one. A new
  `ConstantInlinability` variant, `'module-scope-safe'`, mirrors the existing
  `FunctionInlinability` variant of the same name — `toLegacyInlinability` treats it
  as "own routing" (not inlined-by-value, not unsafe), same as `jsx-inline` /
  `system-construct`. `populateCsrInlinable`'s CSR-substitution pass gives such a
  constant an **identity** substitution (`ctx.csrInlinable` maps the name to itself)
  instead of `null`, so `mergeCsrNullUnsafe` doesn't re-fold it back into the unsafe
  set.
- **`emit-module-level.ts`**: fixes the parenthesization trap the issue flagged —
  `var f = f ?? (s) => …` is a `SyntaxError` (`??`'s right operand can't be an
  unparenthesized arrow function). Now wraps the initializer in parens when
  `containsArrow` is set, a no-op for the existing parenthesized-object-literal
  shapes.
- I deliberately did **not** touch `relocate.ts`'s `buildRelocateEnvFromFields` (item
  3 above) — its `'module-local'` binding-kind answer is consumed only via
  `compute-inlinability.ts`'s `buildEnvFromCtx`, and for the concrete reported bug
  `classifyConstantInitial`'s `containsArrow` branch short-circuits before ever
  reaching that code path, so it isn't the proximate cause here. Unifying it onto
  the same `computeDeclarationScopes` answer would need threading a `ReferencesGraph`
  into a currently graph-agnostic, widely-used env builder — a larger, separately
  scoped change I'm not comfortable making inside this fix's blast radius.

## Also found while verifying: a pre-existing, unrelated cross-adapter gap (#2994)

Writing the regression fixture (`module-const-arrow-helper`, exercised across every
adapter's conformance suite by this repo's shared-corpus architecture) surfaced that
non-JS "Marked Template" adapters (Go, ERB, Jinja, Twig, Blade, Xslate, Mojolicious,
minijinja/Rust) share the *same* `compute-inlinability.ts`-driven template emission
as the CSR path — so a module-scope-safe helper (arrow **or** the pre-existing
`function` form — verified both) is now *also* referenced by bare name in their SSR
templates, which they have no way to execute (no JS runtime). This renders silently
empty there too, with no compile diagnostic.

This is **not** caused by this fix (the `function` form already had the identical
gap, confirmed directly) and is out of #2988's scope, so I filed it separately as
#2994 (`known-limitation` + `bug`) and added a `render-divergences.ts` pin on all
8 non-Hono adapters — **every one directly verified against its real toolchain**
in this session (installing what was missing: `composer install` for Twig/Blade,
CPAN's `Text::Xslate`/`Mojolicious` modules, and fetching a compliant Go 1.25.6
via `GOTOOLCHAIN` where the preinstalled Go was 1.24):

- **ERB, Jinja, minijinja (Rust), Twig, Blade, Text::Xslate**: silently render the
  slot **empty**, no compile diagnostic.
- **Go**: `html/template` fails template EXECUTION outright (`can't evaluate field
  Fmt in type main.WidgetProps`).
- **Mojolicious**: also fails template execution, but differently — Perl `strict`
  rejects the bare `$fmt` (`Global symbol "$fmt" requires explicit package name`).
  (ERB was ported FROM this adapter but does NOT share this crash.)

Each `render-divergences.ts` entry's reason string reflects the adapter's actual
observed failure mode, not a generic one.

While verifying, this fix also graduated a stale, PRE-EXISTING CSR-conformance
skip: `array-map-function-reference` (#2073) hit the exact same
unconditional-`arrow-literal` gap independent of the `.map()` callback's own
structural splice mechanism — its containing template expression's whole-string
safety check saw the bare word `format` and fell back to empty. Removed from
`csr-skip-set.ts` with a graduation note; `csr-skip-rot.test.ts` (which fails loudly
on a stale skip entry) confirms.

`packages/compat`'s `render-divergences.test.ts` gate (fixture exists in the corpus,
compiles clean, not also pinned in `conformancePins`) passes for every entry added.
`ui/compat.lock.json` / `ui/support-matrix.lock.json` regenerated via `bun run
compat:lock` / `bun run support-matrix:lock` and committed — `ci-compat.yml`'s drift
gate would otherwise fail on this PR.

## Test / fixture plan

- `packages/adapter-tests/fixtures/module-const-arrow-helper.ts` — the reproducing
  fixture from the issue body (module arrow helper applied to a prop). `expectedHtml`
  is auto-generated from the Hono reference adapter (real JS execution) and is
  correct as-is — Hono was **not** the broken side here (verified directly:
  pre-fix, Hono's real SSR already rendered `HI` correctly; only the CSR template
  lambda was broken), so no `SKIP_AUTO_UPDATE` entry was needed.
  - Passes SSR conformance on Hono; skipped (via a new `render-divergences.ts` pin,
    per adapter) on all 8 non-Hono adapters for the unrelated #2994 gap — each pin
    verified against the adapter's real toolchain.
  - Passes `csr-conformance.test.ts` — this is the fix's actual regression coverage.
- `packages/jsx/src/__tests__/module-scope-arrow-const-2988.test.ts` — compiler unit
  tests exercising `compileJSX` directly (the same entry point
  `@barefootjs/vite`'s plugin and the CSR-conformance harness call): the reported
  repro renders the real value (not `${''}`) and is byte-parenthesized correctly;
  an arrow/function-declaration parity check; a negative control (an arrow const
  that reads a signal must still demote to `'init'`, guarding against
  over-widening the fixpoint); and a negative control for a component-scope
  (non-module) arrow const.
- `packages/jsx/src/__tests__/module-helper-scope-2986.test.ts` — two pre-existing
  assertions updated: `score`/`loadData` (arrow-valued module helpers with no
  component-scope references) now correctly promote to true module scope
  (`var score = score ?? ((title, term) => {...})`) instead of staying a
  per-instance `const` inside `initSearch`/`initWidget` — this is the intended
  consequence of #2988's fix, not a regression (the "no duplicate inner
  declaration" invariant #2987 protects is unaffected, confirmed by the
  surrounding, unchanged assertions still passing).

## Real-pipeline verification

**Real entry point**: `packages/jsx/src/ir-to-client-js/generate-init.ts`'s
`generateInitFunction` (called by `compileJSX` → `generateClientJsWithSourceMap`,
which is what `@barefootjs/vite`'s plugin (`packages/vite/src/plugin.ts:284,401`)
calls for every component file it compiles).

**pipelineAssumptions**, each verified against that real entry point (not just a
hand-built unit test):

- *`computeInlinability`'s fresh call to `computeDeclarationScopes(ctx, graph)`
  sees the same `(ctx, graph)` pair `classifyLocalDeclarations` already computed
  a scope answer from, unmutated in between.* Verified by reading
  `generate-init.ts:66-73`: `graph` is built once via `buildReferencesGraph(ctx,
  ir.root)`, then passed unchanged to `classifyLocalDeclarations(ctx, graph)` and,
  three lines later, to `buildInlinableConstants(ctx, graph, ir.root)` →
  `computeInlinability(ctx, graph, irRoot)` — and both `computeDeclarationScopes`
  (`compute-scope.ts`) and `classifyLocalDeclarations` (`init-declarations.ts`) are
  pure reads of `ctx`/`graph` with no mutation (read in full).
- *`ConstantInfo.freeIdentifiers` (the constant fixpoint's reference set) already
  excludes an arrow's own bound parameter names, so no extra stripping is needed
  (unlike the function branch, which strips `fn.params` from the graph's
  regex-derived reference set).* Verified by reading
  `analyzer.ts:2909-2948`'s `extractFreeIdentifiersFromNode` — its
  `ts.isArrowFunction` branch (2935-2942) adds param names to `boundNames` before
  descending and removes them after, for every arrow node in the tree, not just
  the top-level declaration form.
- *Non-JS adapters sharing the same emission path is a real, not hypothetical,
  interaction.* Verified empirically (not just reasoned about) by running the real
  per-adapter `render*Component` test-render functions directly, both before and
  after adding this PR's `render-divergences.ts` pins — see the #2994 section above.

**pipelineVerifiedAgainst**: `packages/jsx/src/ir-to-client-js/generate-init.ts:66-103`
(the real `compileJSX`/`generateClientJsWithSourceMap` pipeline), cross-checked via
`packages/adapter-tests/src/csr-render.ts`'s `renderCsrComponent` (same
`generateClientJs` entry point the CSR-conformance suite and this PR's new fixture
both exercise) and `packages/adapter-hono/src/test-render.ts`'s
`renderHonoComponent` (real `compileJSX` + genuine Hono/JSX execution) for the
#2988 repro itself, both re-run after the fix.

Closes #2988


kfly8:
The `check` / "Changeset check" job is failing because this PR modifies published package source (`packages/jsx/src/ir-to-client-js/compute-scope.ts`, `compute-inlinability.ts`, `emit-module-level.ts`) without adding a changeset.

This PR is still being produced by an in-progress `implement-issues` workflow run (Implement/Polish phases). To avoid racing a concurrent push to this same branch, I'll add the changeset (`bunx changeset`) once that run finishes, then re-check CI.

---

kfly8:
Thanks for the review. The function-side analog is real and confirmed pre-existing (untouched by this PR's diff) — filed as #3000 (`bug` + `known-limitation`, same tier as #2994) rather than expanding this PR's scope, following the precedent this PR already set for #2994.

---

kfly8:
@kfly8 This PR is ready for another pass.

**What**: Fixes #2988 — a module-top-level `const name = (params) => expr` helper that closes over nothing but its own params was silently rendering an empty CSR template seed (`${''}`) instead of being referenced by name, because `compute-inlinability.ts` treated every arrow-valued constant as template-unsafe unconditionally, never consulting whether `compute-scope.ts`'s existing module/init fixpoint (already used for functions) would actually hoist it to real module scope. The fix folds arrow-valued module constants into that same fixpoint and has `compute-inlinability.ts` read its verdict instead of re-deriving one, plus fixes a parenthesization trap in `emit-module-level.ts` (`x ?? (s) => ...` is a SyntaxError).

Since Polish's own pass, two more things landed:
- A stale `ui/compat.lock.json` entry (`go-template`/`mojolicious` divergence reasons for the new fixture) that was the actual cause of the "lockfile" CI job being red — fixed and pushed.
- The "Changeset check" CI job was also red (no changeset for a PR touching published `packages/jsx/src`) — added `.changeset/module-scope-arrow-const-2988.md` (patch, `@barefootjs/jsx`).
- Pullfrog's review flagged one pre-existing, non-blocking gap it found while reviewing (the function-side analog of this bug — `functionReferencesDeclaredName` is a third, more conservative producer, untouched by this PR). Filed separately as #3000 (`bug` + `known-limitation`), same treatment as #2994.

**How to verify**:
1. `cd packages/jsx && bun test src/__tests__/module-scope-arrow-const-2988.test.ts src/__tests__/module-helper-scope-2986.test.ts` — compiler-unit regression coverage for the fix itself (including negative controls: a signal-reading arrow const and a component-scope arrow const both correctly stay template-unsafe).
2. `cd packages/adapter-tests && bun test src/__tests__/csr-conformance.test.ts -t "module-const-arrow-helper|array-map-function-reference"` — the new fixture and the graduated pre-existing skip both pass CSR conformance byte-for-byte against the Hono reference.
3. `bun test packages/adapter-tests/src/__tests__/coverage-map.test.ts` — the new `array-method`/`arrow` catalogue entries are covered (this repo's same-PR fixture-coverage rule).
4. `bun run compat:lock && bun run support-matrix:lock && git status --short` — clean.
5. Full typecheck: `cd packages/jsx && bunx tsc --noEmit -p .` — clean.

**Expected result**: All the above pass with no failures/diffs. CI is now green across all 46 checks on the current head, mergeable state is clean, and Pullfrog's review has no remaining blocking findings.

---

---

PR #3018, opened 2026-09-15
minijinja (Rust): scope module-scope-helper BF101 refusal by callee identity

### LINKED ISSUE #3012: Non-JS adapters: `_boolContext` exempts BF101 by structural position, not callee identity — any helper called in a boolean-test position silently keeps the pre-#2994 broken fallback
## Summary

Found by Pullfrog's review of #3011 (the fix for #2994).

#3011 taught each non-JS "Marked Template" adapter's `call()` fallback to refuse a bare-name call to an unresolvable identifier with **BF101** — except when the call's return value is only ever consumed for truthiness (`_boolContext`). That exemption exists specifically so `ui/components/ui/slot/index.tsx`'s `isValidElement(children)` guard keeps compiling on the 6 adapters that lack a dedicated shape-check primitive (ERB, Jinja, minijinja/Rust, Twig, Blade, Text::Xslate — Go and Mojolicious instead special-case `isValidElement` by name via a runtime primitive, so their generic refusal has no carve-out).

The problem: `_boolContext` is scoped by **structural position** (any call inside a condition/ternary-test/unary-`!` operand), not by **callee identity**. So on those 6 adapters, `if (someOtherHelper(x))` or a ternary condition calling *any* other user-authored helper silently continues the pre-#3011 broken behavior (undefined-variable semantics, typically empty/falsy) instead of getting the BF101 refusal the rest of this PR's title promises. This path has **zero fixture/test coverage** — the graduated fixtures (`module-const-arrow-helper`, `module-function-helper-chain`) only call their helper in text position, never inside a condition, so nothing in the corpus catches a regression here or confirms the intended scope.

## Compounding detail: Xslate already has the runtime capability Go/Mojolicious used

Xslate already `use`s the shared `BarefootJS` Perl runtime (`packages/adapter-xslate/lib/BarefootJS/Backend/Xslate.pm:101`), which has the same `is_element` method Mojolicious's dedicated `isValidElement` primitive calls (`packages/adapter-perl/lib/BarefootJS.pm:1074` — explicitly documented in the Mojolicious changelog as "also used by Xslate"). So for Xslate specifically, the identity-scoped fix Go/Mojolicious got was already free (no new runtime code needed) and wasn't taken during #3011; it fell back to the coarser `_boolContext` exemption like the adapters that genuinely lack a shape-check helper.

## Verified

Read `packages/adapter-erb/src/adapter/expr/emitters.ts:554`:
```ts
if (callee.kind === 'identifier' && !this._boolContext) {
```
— confirms the refusal gate has no callee-identity check at all; any identifier call under `_boolContext: true` is exempted, not just `isValidElement`.

## Suggested fix shape

- **Xslate**: register `isValidElement` as an identity-scoped primitive (mirroring Go/Mojolicious) via the already-`use`d `is_element` method, and drop the `_boolContext` exemption entirely for this adapter — closes the gap with zero new runtime code.
- **ERB / Jinja / minijinja (Rust) / Twig / Blade**: would need net-new runtime helper code to add an identity-scoped `isValidElement` primitive before the coarse `_boolContext` exemption could be narrowed or dropped — reasonably separate scope from the above.
- Either way: add a fixture (or widen an existing one, e.g. `module-const-arrow-helper`-style) that calls a *non-`isValidElement`* helper from inside a boolean-test position (an `if`/ternary condition), asserting BF101 fires — this is what would have caught the gap and is what's needed to pin the fix once made.

## Test / fixture plan

Per CLAUDE.md: a conformance fixture exercising a user-authored helper called from a condition/ternary-test position, asserting BF101 (the CORRECT output per this PR's own stated scope), plus — until the identity-scoped primitive lands per-adapter — a pin on the broken side (a `conformancePins`-style exemption) carrying this issue's URL for each of the 6 still-coarse adapters.

Related: #2994 (closed by #3011), #3000, #2988.

---
_Filed from Pullfrog's review of #3011; see that PR for full context._

kfly8:
## Summary

Stacks on #3017 (the Jinja fix) — same fix, applied to the **minijinja (Rust)** adapter. See #3016/#3017 and #3012 for the full `_boolContext`/#2994/#3011 background.

## What this PR does

- Adds `runtime::is_element` (`packages/adapter-rust/runtime/src/runtime.rs`), a Rust port of the shared Perl runtime's `is_element` method: true only for a `JsValue::Object` carrying both a `tag` and a `props` key (case-insensitively). Wired into the `call_method`-style dispatch match as the `"is_element"` arm, alongside `"length"` etc.
- Registers `isValidElement` as an identity-scoped `templatePrimitive` (`bf.is_element`, `lib/constants.ts`) resolved ahead of `call()`'s generic fallback.
- Removes the `_boolContext` structural exemption entirely (the field, `_underBoolContext`, and the `boolContext` parameter threaded through `convertExpressionToJinja` / `convertConditionToJinja` / `renderParsedExprToJinja` in `minijinja-adapter.ts`) — every other bare-name call on minijinja now refuses with BF101 regardless of position.
- Graduates `module-helper-boolcontext-call` from `render-divergences.ts` to a `conformance-pins.ts` BF101 pin, mirroring the two existing module-scope-helper pins.

## Verification

- New `is_element_helper` test in `runtime/tests/template_primitives.rs` (tag+props, case-insensitive match, missing tag, missing props, a plain string child, `Null`, an array) plus the full `cargo test` suite for the runtime crate: 0 failures across every test binary.
- `packages/compat`'s `compat-pins.test.ts` + `render-divergences.test.ts`: 219 pass, 0 fail.
- `packages/adapter-rust`'s real-minijinja (`bf-render` binary) conformance suite, filtered to the boolcontext/isValidElement/slot-related tests: 411 pass, 25 skip, 0 fail.
- `ui/compat.lock.json` and `ui/support-matrix.lock.json` regenerated from the rebuilt package and committed alongside; `coverage-map.json` recomputed with no diff.

## Scope note

This does **not** close #3012: the Twig / Blade gap it also describes stays open, tracked by #3012 itself, addressed in PRs stacked on top of this one.

---


kfly8:
@kfly8 This PR is ready for review.

**What**: Closes #3012's gap for the minijinja (Rust) adapter, same pattern as #3016/#3017. `isValidElement` is now resolved as an identity-scoped `templatePrimitive` (`bf.is_element`, backed by a new `is_element` function in the Rust runtime, wired into `BfInstance::call_method`'s dispatch) ahead of `call()`'s generic BF101 fallback, and the `_boolContext` structural exemption is removed entirely. `module-helper-boolcontext-call` graduates from a `render-divergences.ts` pin to a `conformance-pins.ts` BF101 pin.

**How to verify**:
1. `bun install && bun run --filter '@barefootjs/rust' build`
2. `bun test --cwd . packages/compat/src/__tests__/compat-pins.test.ts packages/compat/src/__tests__/render-divergences.test.ts`
3. `cargo test --manifest-path packages/adapter-rust/runtime/Cargo.toml` — includes the new `is_element_helper` test.
4. `bun test --cwd . packages/adapter-rust/src/__tests__` — full minijinja conformance suite against the real `bf-render` binary.
5. Manual repro: compile a component with `<span title={someHelper(x) ? 'a' : 'b'}>` (a non-`isValidElement` module-scope helper in a ternary test) through the minijinja adapter and confirm BF101 fires instead of silently rendering `'b'`.

**Expected result**: Steps 2–4 all green (cargo test: 0 failures across every test binary). Step 5 raises BF101 naming the helper.

---

PR #3019, opened 2026-09-15
Twig: scope module-scope-helper BF101 refusal by callee identity

### LINKED ISSUE #3012: Non-JS adapters: `_boolContext` exempts BF101 by structural position, not callee identity — any helper called in a boolean-test position silently keeps the pre-#2994 broken fallback

kfly8:
## Summary

Stacks on #3018 (the minijinja fix) — same fix, applied to the **Twig** adapter. See #3016/#3017/#3018 and #3012 for the full `_boolContext`/#2994/#3011 background.

## What this PR does

- Adds `BarefootJS::is_element` (`packages/adapter-php/src/BarefootJS.php` — the shared PHP runtime **both** the Twig and Blade adapters back onto), a port of the shared Perl runtime's `is_element` method: true only for a JS *object* value (`isJsObject`'s existing definition: a `stdClass` or a non-list PHP array) carrying both a `tag` and a `props` key (case-insensitively). Because the runtime is shared, **this one addition also satisfies Blade's half of #3012** — only Blade's own compiler-side wiring remains, in the next stacked PR.
- Registers `isValidElement` as an identity-scoped `templatePrimitive` (`bf.is_element`, `lib/constants.ts`) resolved ahead of `call()`'s generic fallback.
- Removes the `_boolContext` structural exemption entirely (the field, `_underBoolContext`, and the `boolContext` parameter threaded through `convertExpressionToTwig` / `convertConditionToTwig` / `renderParsedExprToTwig` in `twig-adapter.ts`) — every other bare-name call on Twig now refuses with BF101 regardless of position.
- Graduates `module-helper-boolcontext-call` from `render-divergences.ts` to a `conformance-pins.ts` BF101 pin, mirroring the two existing module-scope-helper pins.

## Verification

- New `is_element` test in `packages/adapter-php/tests/test_template_primitives.php` (tag+props by array and by `stdClass`, case-insensitive match, missing tag, missing props, a plain string child, `null`, a list) plus the full PHP runtime suite (`tests/run.php`): 575/575 pass.
- `packages/adapter-twig/php`'s real-Twig render test (`tests/test_render.php`, resolving the shared runtime through its composer path-repository symlink): 10/10 pass.
- `packages/compat`'s `compat-pins.test.ts` + `render-divergences.test.ts`: 219 pass, 0 fail.
- `packages/adapter-twig`'s real-Twig conformance suite, filtered to the boolcontext/isValidElement/slot-related tests: 411 pass, 25 skip, 0 fail.
- `ui/compat.lock.json` and `ui/support-matrix.lock.json` regenerated from the rebuilt package and committed alongside; `coverage-map.json` recomputed with no diff.

## Scope note

This does **not** close #3012: the Blade compiler-side wiring stays open, tracked by #3012 itself, addressed in the next stacked PR (the shared runtime half is already done here).

---


kfly8:
@kfly8 This PR is ready for review.

**What**: Closes #3012's gap for the Twig adapter, same pattern as #3016/#3017/#3018. Adds `BarefootJS::is_element` to the **shared PHP runtime** (`packages/adapter-php`, used by both Twig and Blade), so this same addition also covers Blade's runtime half — only Blade's compiler-side wiring remains (next PR in the stack). `isValidElement` resolves as an identity-scoped `templatePrimitive` ahead of `call()`'s generic BF101 fallback, and the `_boolContext` structural exemption is removed entirely on Twig. `module-helper-boolcontext-call` graduates from a `render-divergences.ts` pin to a `conformance-pins.ts` BF101 pin.

**How to verify**:
1. `bun install && bun run --filter '@barefootjs/twig' --filter '@barefootjs/php' build`
2. `bun test --cwd . packages/compat/src/__tests__/compat-pins.test.ts packages/compat/src/__tests__/render-divergences.test.ts`
3. `php packages/adapter-php/tests/run.php` — full shared PHP runtime suite, includes the new `is_element` cases.
4. `php packages/adapter-twig/php/tests/test_render.php` — real Twig render through the shared runtime via composer path-repository symlink.
5. `bun test --cwd . packages/adapter-twig/src/__tests__` — full Twig conformance suite against real Twig.
6. Manual repro: compile a component with `<span title={someHelper(x) ? 'a' : 'b'}>` (a non-`isValidElement` module-scope helper in a ternary test) through the Twig adapter and confirm BF101 fires instead of silently rendering `'b'`.

**Expected result**: Steps 2–5 all green (PHP runtime suite 575/575 pass, Twig render test 10/10 pass). Step 6 raises BF101 naming the helper.

---

Issue #3032, opened 2026-09-16
Author /* @client */ escape corpus fixtures for the module-scope-helper-call BF101/BF102 refusal family


kfly8:
## Summary

#3030's compat-issue-freshness check flagged `unescapable: { issue }` (and the paired top-level `issue`) citations on 27 `conformancePins` entries across all 9 non-Hono adapters, all pointing at issues that are now closed (#2994, #3012, and Go's own #2266) — because those issues tracked *landing the refusal itself*, which is done, not the separate, still-open piece of work this issue now tracks: authoring a corpus-pinned `/* @client */` escape twin.

Affected fixtures, each refused with `unescapable` on every non-Hono adapter (Go's `module-helper-boolcontext-call` refuses BF102, all others BF101):

- `module-const-arrow-helper`
- `module-function-helper-chain`
- `module-helper-boolcontext-call`

None of the three fixtures declares an `escapes` twin, and no `-client` sibling fixture exists in `packages/adapter-tests/fixtures/`. Every adapter's own pin comment says the same thing: `/* @client */` "works informally but isn't pinned as a corpus fixture yet."

## Why this is its own issue, not a re-point to #2994/#3012/#2266

Per this repo's escape-coverage floor (#2613, `packages/compat/src/__tests__/escape-coverage.test.ts`), `unescapable: { issue }` means "an escape is owed but not authored yet" and needs its own tracking pointer, distinct from the pin's top-level `issue` (which tracks the refusal/limitation itself). #2994 and #3012 fully resolved what they set out to (turn the silent divergence into a loud, correct refusal) and are rightly closed; neither ever promised a corpus escape fixture as an acceptance criterion, so re-opening either would misrepresent their scope. #2266 established Go's identity-scoped `isValidElement` resolution (also fully landed) and is unrelated to escape-authoring.

## Required outcome

Author a `client-directive` escape fixture (empty-region `expectedHtml`, `ssrCost: 'client-render'`, mirroring the existing `filter-nested-callback-predicate` / `filter-nested-callback-predicate-client` twin pattern) for each of the three fixtures above, declare it in the fixture's `escapes` array, and confirm `packages/compat/src/__tests__/escape-coverage.test.ts` verifies it (compiles clean, unpinned, non-divergent, not CSR-skipped) on every adapter that currently marks the shape `unescapable`. Delete the now-satisfied `unescapable` fields as each twin lands (shrink-only, same discipline as `KNOWN_HOLES`).

## References

[...]

Issue #3009, opened 2026-09-15
Keyed .map() row's ref doesn't re-run when a branch is re-entered after a key round-trip through another branch


kfly8:
## Summary

A keyed `.map()` row whose body branches on a condition (`cond ? <A ref={...}/> : <B/>`) does not re-run branch A's `ref` when the row's key returns to branch A after a period on branch B — even though [#2927](https://github.com/piconic-ai/barefootjs/issues/2927)'s own investigation established that a branch's `bindEvents()` (and so its refs) *does* re-run on every re-entry. Confirmed by direct real-browser reproduction (compiled output inspection + Playwright), not just `bf debug graph`.

## Minimal repro shape

```tsx
type Row = { key: string } & (
  | { kind: 'a'; label: string }
  | { kind: 'b' }
)

function List(props: { rows: Row[] }) {
  return (
    <div>
      {props.rows.map(row => (
        <div key={row.key}>
          {row.kind === 'a' ? (
            <span ref={el => { el.dataset.mounted = 'yes' }}>{row.label}</span>
          ) : (
            <span>placeholder</span>
          )}
        </div>
      ))}
    </div>
  )
}
```

Drive it through three renders with the **same key** each time:
1. `rows = [{ key: 'x', kind: 'a', label: 'hi' }]` — mounts fine, `data-mounted="yes"` present.
2. `rows = [{ key: 'x', kind: 'b' }]` — switches to the `b` branch, as expected.
3. `rows = [{ key: 'x', kind: 'a', label: 'hi' }]` again — **the `a` branch's `<span>` element exists in the DOM, but `ref` never ran**: no `data-mounted` attribute, and any imperative setup the ref does (attaching a child, calling an API, etc.) never happens.

## What we actually observed (real app, not the toy example above)

We hit this in a production component (`peitho-studio`, a Tauri app) with a keyed `.map()` row like:

```tsx
props.entries.map(entry => (
  <div key={entry.kind === 'rendered' ? entry.slide.key : entry.key}>
    ...
    {entry.kind === 'rendered' ? (
      <div ref={el => {
        el.dataset.slideCanvasKey = entry.slide.key
        mountSlideCanvas(el, /* ... */)
      }} />
    ) : (
      <div>placeholder</div>
    )}
  </div>
))
```

[...]

Issue #2992, opened 2026-09-14
Go-template cross-file type resolution is alphabetical-order-dependent, not import-order-dependent


kfly8:
## Summary

#2991 (fixing #2984) added a cross-file type registry to `GoTemplateAdapter`
(`crossFileTypeAliases` / `crossFileStructFields` / `crossFileTypeDefinitions`,
populated inside `buildLocalTypeTables` — see
`packages/adapter-go-template/src/adapter/go-template-adapter.ts`) so a client
component that only *imports* an object-shaped prop type (rather than
declaring it locally) can resolve that type against an earlier-compiled
component's own type table instead of falling back to `interface{}`/`nil`.

That registry is populated as a side effect of compiling the type's
*defining* file. It only has an answer for a consumer if the definer has
already gone through `buildLocalTypeTables` by the time the consumer is
compiled — i.e. correctness depends on **compile order**.

The real production pipeline does not compile in dependency order. `@barefootjs/vite`'s
`discoverComponentFiles` (`packages/vite/src/discover.ts`) walks each
directory and sorts entries with plain `localeCompare` — alphabetical by
filename, not by import graph. `emitTemplatesFor` / `runDevEagerPass`
(`packages/vite/src/plugin.ts`) then compile `discovered` in that exact flat
order, for both the dev eager pass and the production `writeBundle` pass.
There is no import/dependency-graph ordering anywhere in that pipeline, and
the single `GoTemplateAdapter` instance used across a build accumulates the
registry strictly in file-discovery order.

## Concrete failure case

A project with `App.tsx` importing an object-shaped prop type from
`Zebra.tsx` (`'App' < 'Zebra'` alphabetically) hits the *original*
`interface{}` / `nil` fallback in a real `vite build` or dev pass, because
`App.tsx` compiles — and tries to resolve `Zebra`'s type — before
`Zebra.tsx` has registered anything into the cross-file registry. This is
not a new regression introduced by #2991 (it's the same broken output the
compiler produced before that PR), but it means #2991 only closes #2984 for
the favorable ordering that issue's own repro happened to use
(`Child.tsx` defining the type, `Parent.tsx` consuming it — `'Child' <
'Parent'`), not the general case of "a component imports an object-shaped
prop type declared in another file."

## Why this is `+bug`, not a permanent design position

The compiler emits no diagnostic when this happens — it silently falls back
to `interface{}` / `nil` for an unfavorably-named file pair, which then
[...]

PR #3033, opened 2026-09-16
Fix #3030, #3009, #2992: stale compat pins, keyed-row ref re-entry, Go-template cross-file type order

### LINKED ISSUE #2946: A module-level static array `.map()`'d directly in a client component fails SSR on every non-Hono adapter
## Summary

A `'use client'` component that declares a static array at **module scope** (outside the component function) and `.map()`s it directly — no signal wrapper — fails to render its rows on every non-Hono adapter. Hono (the reference adapter) renders correctly.

Declaring the exact same array **inside** the component function (component-local `const`) instead of at module scope works everywhere. So the trigger is specifically "static array declared at module scope, then `.map()`'d directly in a client component" — nothing to do with events, containers, or any other feature.

## Minimal reproduction

```tsx
'use client'
const items = ['a', 'b', 'c']
export function Repro() {
  return (
    <div className="container">
      {items.map(item => <div key={item} data-item={item}>row {item}</div>)}
    </div>
  )
}
```

Verified locally against `renderErbComponent` (`@barefootjs/erb`):

```
Error: ruby render failed (exit 1):
(erb):4:in `render': undefined method `each_with_index' for nil (NoMethodError)
```

Moving `const items = ['a', 'b', 'c']` from module scope into the component body (otherwise identical source) renders correctly:

```html
<div class="container" bf-s="test" bf-r="" bf="s2"><!--bf-loop:l0-->
<div data-item="a" data-key="a" bf="s1">row <!--bf:s0-->a<!--/--></div>
<div data-item="b" data-key="b" bf="s1">row <!--bf:s0-->b<!--/--></div>
<div data-item="c" data-key="c" bf="s1">row <!--bf:s0-->c<!--/--></div>
<!--bf-/loop:l0--></div>
```

## Observed on other adapters (via CI on #2944, same module-scope-array shape, plus an unrelated pair of event handlers that turned out not to matter)

- **ERB, Twig, Blade, Jinja, Xslate, Rust (minijinja)**: SSR renders the container element but completely empty — the loop body never appears (`<div class="container"></div>`).
- **Mojolicious**: template compile error — `Global symbol "$items" requires explicit package name (did you forget to declare "my $items"?)`. The generated template never declares `my $items` before the `for` loop.
- **Go template**: template execution error — `can't evaluate field Items in type main.ReproProps` — the adapter is treating the module-level array as if it needed to be a props-struct field, and never populates one.
- **Hono (reference)**: renders correctly.

This is consistent with the `INITIAL_VALUE` signal-seed family of bugs (#2862, #2815, #2794) — the analyzer/adapters resolving a bare identifier back to a module-level const declaration — but none of those covered a **directly-`.map()`'d static array with no signal wrapper in between**. `module-const-array-seed.ts`'s existing fixture only covers the signal-seeded case (`createSignal(INITIAL)`), which already works; the direct-`.map()` case does not.

## Suggested next step

Per this repo's fixture-first policy: land a minimal fixture pinning the correct (Hono) output, with `skipJsx`/equivalent pins on each broken adapter referencing this issue, then root-cause per-adapter (likely each non-Hono adapter's static-array codegen path needs to resolve a `.map()` target identifier back to its module-level declaration the same way the signal-seed fix path already does, rather than assuming it's always a props/local reference).

Found while adding a conformance fixture for #2930 (unrelated) — see PR #2944, where the fixture was changed to use a component-local array instead to avoid conflating the two issues.

kfly8:
## Summary

Work in progress covering three issues:

- [x] #3030 — compat-issue-freshness flagged 27 stale closed-issue citations (`unescapable`/`issue` fields in `conformance-pins.ts`) across all 9 non-Hono adapters, plus one mis-pointed citation on Pebble. Re-pointed the module-scope-helper-call family (#2994/#3012/Go's #2266) to a new tracking issue (#3032, since none of those closed issues ever promised the corpus-pinned `/* @client */` escape twin `unescapable` tracks), and Pebble's `module-const-loop-source-computed` pin from closed #2946 to the already-correct-on-every-sibling-adapter open #2321. Lock regeneration (`bun run compat:lock && bun run support-matrix:lock`) still pending a full workspace build.
- [ ] #3009 — keyed `.map()` row's `ref` doesn't re-run when a branch is re-entered after the row's key round-trips through another branch. Investigating root cause in the client runtime's branch-swap/`bindEvents` machinery.
- [ ] #2992 — Go-template cross-file type resolution is alphabetical-order-dependent (depends on file *discovery* order, not import order). Investigating a lazy/on-demand resolution fix.

## Test plan

- [ ] `bun run compat:lock && bun run support-matrix:lock` regenerated and committed for #3030
- [ ] `bun test packages/compat` green
- [ ] Regression test/fixture added for #3009 (compiler unit and/or CSR conformance, per CLAUDE.md's testing table)
- [ ] Pinned regression test in `packages/adapter-go-template/src/__tests__/go-template-adapter.test.ts` flipped to the correct output for #2992
- [ ] `bun test` full suite green

https://claude.ai/code/session_016xEZffxdj8h42EA56mQXVb

---
_Generated by [Claude Code](https://claude.ai/code/session_016xEZffxdj8h42EA56mQXVb)_


kfly8:
Closing this in favor of three separate PRs, one per issue, so each can be reviewed/merged/rolled back independently:

- #3034 — #3030 (stale compat-issue-freshness pins)
- #3035 — #3009 (loop row branch ref re-entry)
- #3036 — #2992 (Go-template cross-file type resolution order)

---

Issue #2927, opened 2026-09-10
createEffect() inside a conditional branch's ref leaks one effect per re-entry


kfly8:
## Summary

`createEffect()` called from a `ref` callback inside a conditional (`cond ? <A ref={...}/> : <B/>`) branch leaks one effect per re-entry into that branch — the effect from every prior visit to the branch keeps running forever, even though its host element has been unmounted.

## Root cause (traced through the compiled output + runtime source)

`insert()` in `@barefootjs/client`'s `runtime/index.js` has a `branchCleanup` mechanism specifically designed for this: it calls `branch.bindEvents(...)`, stores whatever that call returns as `branchCleanup`, and — the next time the condition flips — calls `branchCleanup()` before evaluating the new branch:

```js
// runtime/index.js (insert())
const cleanup = branch.bindEvents(region.bindScope, { isFirstRun: false });
branchCleanup = typeof cleanup === "function" ? cleanup : null;
```

The problem is on the compiler side: the `bindEvents` the compiler emits for a branch containing a `ref` callback that calls `createEffect()` never returns anything, so `cleanup` is always `undefined` and `branchCleanup` stays `null` forever. `createEffect()`'s own disposal is never wired into the branch's cleanup, so the runtime's teardown mechanism has nothing to call.

Compiled output for the repro below (`Repro.tsx`, readable with variable names de-minified for clarity):

```js
bindEvents: (scope, { isFirstRun }) => {
  const [hostEl] = find(scope, "s3")
  hostEl && (el => {
    createEffect(() => { /* ... */ }) // <- return value discarded, nothing returned from bindEvents
  })(hostEl)
}
```

## Minimal reproduction

Scaffolded with `npm create barefootjs@latest` (Hono / Cloudflare Workers + UnoCSS template), `@barefootjs/client@0.35.5`.

`components/Repro.tsx`:

```tsx
'use client'

import { createSignal, createEffect } from '@barefootjs/client'

export function Repro() {
  const [show, setShow] = createSignal(true)
  const [tick, setTick] = createSignal(0)

  return (
    <div>
      <button id="toggle" onClick={() => setShow(!show())}>toggle</button>
      <button id="bump" onClick={() => setTick(tick() + 1)}>bump</button>
      <div id="log">0</div>
      {show() ? (
        <div
          id="host"
          ref={el => {
            createEffect(() => {
              tick()
              el.dataset.ticks = String(Number(el.dataset.ticks ?? '0') + 1)
[...]

Issue #2347, opened 2026-07-21
Nested conditional around an element with a dynamic attribute/event handler double-binds effects/listeners and orphans the attribute effect


kfly8:
## Description

When a `"use client"` component renders a list where each item's element is wrapped in a conditional (e.g. `cond ? null : <el .../>`, or a deeper `cond1 ? null : cond2 ? <el/> : <other/>`), and that conditionally-rendered element itself has a **dynamic attribute** (e.g. `className={...}`) and/or an **event handler** (`onClick={...}`), the compiler emits binding code for that element **twice**:

1. Once directly inside the *immediately enclosing* branch's `bindEvents`, via `qsa(...)` + `addEventListener` / `createEffect(...)` against the element as it exists in that branch's own initial template.
2. Again inside the *nested* nested conditional's own `insert(...)` call, which creates its own copy of the same markup and (re)binds the click handler / text effects against that separately-created node.

At runtime this causes two distinct problems, both reproduced below at the compiler-output level:

- **Duplicate event handling**: the click handler fires twice per click (confirmed by counting `addEventListener('click'` occurrences in the generated client JS — see Repro 2).
- **Stale/orphaned reactive attribute effect**: on first hydration, `insert()`'s own branch machinery creates a *fresh* DOM node for the branch content instead of adopting the one already present in the parent's initial per-item template. The `class` attribute effect registered at the outer level (Repro 1) stays bound to the original (now-detached) node, so it silently stops updating, while the text content and click handler (bound via the nested `insert()` to the new node) keep working. I hit this "live" while building a followers/following list in an app: the follow button's label (`フォロー` / `フォロー中`) updated correctly on click, but its CSS class (`.on`) never changed — because the class effect was updating a detached clone of the button, not the one visible in the DOM.

Both issues went away once I flattened the JSX so the conditionally-rendered element was the *only* level of branching (no nested ternary, and no per-item dynamic attribute layered on top of a second, unrelated conditional) — see workaround below.

## Steps to Reproduce

Compile the following with `compileJSX` (both snippets are self-contained; run each as a `bun:test` file with `packages/jsx` installed, e.g. inside the `barefootjs` monorepo where `@barefootjs/client` resolves):

### Repro 1 — stale attribute effect (orphaned node)

```ts
import { compileJSX } from './packages/jsx/src/compiler'
import { TestAdapter } from './packages/jsx/src/adapters/test-adapter'

const source = `
  'use client'
[...]

PR #3035, opened 2026-09-16
Fix #3009: re-run a loop row's branch ref on every activation

kfly8:
## Summary

A `ref` on an element inside a keyed `.map()` row's conditional branch (`cond ? <A ref={...}/> : <B/>`) only fired if that branch happened to be active when the row was CREATED — never again on a later activation (starting on the other branch, or a keyed row's identity round-tripping away from and back to this branch).

Root cause: `collectLoopChildRefs` (`ir-to-client-js/reactivity.ts`) descended straight through a nested reactive conditional and hoisted the branch's ref to ROW level, where `emitLoopChildRefs` fires it unconditionally once per `mapArray` renderItem call — row creation, not every mount. Its siblings `collectConditionalBranchEvents` / `collectConditionalBranchRefs` already stop at a reactive conditional for exactly this reason (#2347); `collectLoopChildRefs` was the one collector that never got that treatment.

- Branch-interior refs are now collected onto a new `LoopChildBranchSummary.refs` field (mirroring `.events`) via `collectConditionalBranchRefs`.
- Threaded through a new `BranchRefBindingsPlan` (`plan/loop-child-arm.ts`) built by `buildBranchRefBindingsPlan`.
- Emitted by a new `stringifyBranchRefs` inside the branch's own `insert()` bindEvents — the same place events already (re-)fire on every branch swap (#2927).
- Wired into both arm builders that produce a `LoopChildArmPlan`: `buildLoopChildArmPlan` (nested loop-child conditionals) and `buildOuterArm` (`build-reactive-effects.ts`, the row-direct conditional the #3009 repro actually exercises).
- `lazy-conditional.ts`'s `wiringOn` now also names `refs`, so a branch with a ref can never be silently routed through the lazy-row fast path that bypasses this fix.

Verified the new regression test fails without the fix (temporarily reverted via `git stash`) and passes with it.

## Test plan

- [x] `packages/jsx/src/__tests__/issue-3009-loop-branch-ref.test.ts` — emission pin (ref lands in the branch's own bindEvents using `__branchScope`, not hoisted into the row-level renderItem body)
- [x] `packages/client/__tests__/runtime/issue-3009-loop-branch-ref-reentry.test.ts` — real DOM behavior via happy-dom: re-entry after a round-trip, and first activation when the row is created on the other branch

https://claude.ai/code/session_016xEZffxdj8h42EA56mQXVb

---
_Generated by [Claude Code](https://claude.ai/code/session_016xEZffxdj8h42EA56mQXVb)_


kfly8:
@kfly8 this is ready for review.

**What**: A `ref` on an element inside a keyed `.map()` row's reactive conditional branch only fired once, at row creation — never again on a later branch activation (starting on the other branch, or a keyed row's identity round-tripping away from and back to this branch). Fixed by collecting branch-interior refs onto the branch's own `insert()` bindEvents (mirroring how `events` already work), instead of hoisting them to the row-level `mapArray` renderItem body.

**How to verify**:
```
bun test packages/jsx/src/__tests__/issue-3009-loop-branch-ref.test.ts packages/client/__tests__/runtime/issue-3009-loop-branch-ref-reentry.test.ts
```
You can also confirm the regression test actually catches the bug: `git stash` the `packages/jsx/src/ir-to-client-js/` changes only, re-run the client test — 2 of 3 tests fail with `data-mounted` staying `null` after re-entering the ref-bearing branch; `git stash pop` restores the fix.

**Expected result**: All tests pass with the fix in place. The compiler-emission test confirms the ref call lands inside the branch's `bindEvents` using `qsa(__branchScope, ...)`, not hoisted into the row-level `mapArray` body. The runtime test confirms `data-mounted` is set on every branch activation (initial mount, round-trip back to the branch, and first activation when the row was created on the other branch).

Both `/code-review` and `/simplify` passes came back clean, and Pullfrog's automated review found no issues.

---

Issue #3046, opened 2026-09-18
Proposal: reusable bounded state-space exploration for compiler and hydration semantics


AnaTofuZ:
## Summary

Would it be useful to generalize the existing pairwise browser-oracle work into a reusable bounded state-space exploration harness for BarefootJS?

This would not try to formally verify arbitrary TSX. The proposed scope is a small, deterministic model-based testing layer that explores bounded component states and transition sequences, then compares the observable results across compiler/runtime paths.

The goal is to make bugs involving combinations of reactivity, component boundaries, loops, hydration, and incremental DOM updates easier to discover and reduce mechanically.

## Relationship to #2481

#2481 and its browser-oracle implementation in #2752 already provide much of the necessary foundation:

- generated feature combinations
- deterministic cases
- SSR / hydration / CSR comparison
- hydration-snap detection
- interaction idempotence
- quarantine and fixture-promotion workflow

This proposal is not intended to replace that work.

The possible next step is to separate the useful pieces into a more general exploration loop where a test case consists of:

1. an initial component state
2. a bounded set of actions
3. a state reducer or expected state transition
4. one or more execution modes
5. invariant-based oracles

The existing pairwise generator could then remain one producer of scenarios rather than being the boundary of the system.

## Motivation

Some failures depend not only on a static feature combination, but on the state from which a transition occurs.

One example is this shape:

```tsx
function Child({ data }: { data: { items: Item[] } }) {
  return (
    <ul>
      {data.items.map((item) => (
        <li key={item.id}>{item.label}</li>
      ))}
    </ul>
  );
}

function Parent() {
  const [data, setData] = createSignal({ items: [] });

  onMount(() => {
    setData({ items: [{ id: "1", label: "loaded" }] });
  });

  return <Child data={data()} />;
}
```

A compiler may correctly:

- pass a reactive getter across the component boundary
- render the initially empty loop
- hydrate without throwing

while still failing to install a reconciliation path capable of creating the first row after the array changes from empty to non-empty.

A static feature tuple such as:

```text
signal x child component x keyed loop
```

does not fully describe the relevant case. The transition shape is also significant:

```text
empty array -> non-empty array
```

Other potentially distinct transitions include:

```text
non-empty -> empty
append
insert
delete
reorder
replace with equal-looking values
conditional false -> true
child mount -> update -> unmount
```

## Proposed model

A small scenario could be represented conceptually as:

```ts
interface Scenario<State, Action> {
  initialState: State;
[...]

Issue #2797, opened 2026-08-31
component-row-root-loop: a function-reference event handler prop has its declaration dropped from the emitted module


kfly8:
Found by #2481's pairwise t=3 sweep (variable-strength promotion over the fragile axes, #2796). Case ids:

```
state-prop__structure-component-row-root-loop__binding-class__event-handler-reading-loop-param__callback-function-reference
state-prop-shadowing-signal__structure-component-row-root-loop__binding-text__event-handler-reading-loop-param__callback-function-reference
```

Related to #2750 (same symptom class — a preamble-declared function-reference const is referenced inside a `mapArray` row callback but never declared in the emitted module, so hydrate/csr-mount throws `ReferenceError`) but a **distinct code path**, not a duplicate:

- #2750 is a `ref` callback's call site (`qsa(...)` + a direct function call) dropped specifically inside a **nested** (depth-2) loop. Its own body states single-depth loop structures emit the ref const fine.
- This is a single-depth loop (`component-row-root-loop`, not nested) where the row itself is a **child-component call** (`createComponent('PairwiseRow__...', { onClick: handleRowClick, ... })`), and the dropped identifier is passed as a **prop value** (`onClick`), not called directly.

## Emitted client JS (case 1)

```js
export function initPairwiseCase(__scope, _p = {}) {
  if (!__scope) return
  const __scopeId = __scope.getAttribute('bf-s')

  const [items, setItems] = createSignal([ /* … */ ])

  const [_s2] = $(__scope, 's2')

  mapArray(() => items(), _s2, (row, idx) => String(row.id), (row, idx, __existing) => {
    if (__existing) { initChild('PairwiseRow__f359d3e7', __existing, { get label() { return row().label }, onClick: handleRowClick, get "data-pw-event"() { return "1" }, /* … */ }); return __existing }
    return createComponent('PairwiseRow__f359d3e7', { get label() { return row().label }, onClick: handleRowClick, /* … */ }, row().id)
  }, 'l0')
}
```

`handleRowClick` is never declared anywhere in the module — `initPairwiseCase`'s preamble only emits `items`/`setItems`. Loading the case in `csr-mount` mode throws immediately:

```
PAGEERROR: handleRowClick is not defined
ReferenceError: handleRowClick is not defined
    at .../__client.js:46:117
    at createRoot (.../__runtime.js:460:12)
    at createItemScope (.../__runtime.js:2446:3)
    at mapArray (.../__runtime.js:2504:3)
[...]

PR #3050, opened 2026-09-18
Explore: conditional and child-lifecycle scenarios; unblock the nightly pairwise sweep (#3046)

kfly8:
Second slice of #3046, plus the nightly pairwise sweep brought back to green.

## Explore: two new scenarios

- `conditional-toggle` — one boolean signal gating an `&&` block and a ternary, starting false (`show` / `hide` / `toggle`, depth 3, 2 states / 40 paths). **Green on every path and state.**
- `child-mount-unmount` — a child's lifetime driven by one signal and its prop by another, starting unmounted (`mount` / `unmount` / `relabel`, depth 3, 7 states / 40 paths). **42 rows red, all one mechanism**, quarantined: every one is the sole difference `label="a"` on `Badge`'s root. The hydration effect mirrors the string prop onto an SSR-adopted child (mounted states fail snap / three-point), and a child the conditional constructs client-side after `mount` never receives it (every path ending mounted fails both transition legs against a fresh render that has it). Label *text* is correct on every path. Both faces are the registered `child-prop-mirror-attr-ssr` mechanism and cite it; they graduate together when mirroring becomes consistent.

Full explore sweep on this head: `328 passed, 0 failed` (26 state fixtures, 142 paths).

## Nightly pairwise sweep: red since at least 2026-09-08

`pairwise-sweep.yml` had failed on every scheduled run with the same 22 rows, reproduced identically locally (`918 passed / 22 failed`, 451 cases). Three classes:

| rows | what | action |
|---|---|---|
| 6 | stale quarantine rows — #2797 (all 4, fixed by #2801) and the two #2750 `idempotence` rows (fixed by #2799) | deleted |
| 4 | `nested-loop-depth-2 × ref-callback`: the two #2750 `three-point` rows still failed, but no longer for a `ReferenceError` — the diff is `data-mounted="0"` after hydration, absent from SSR | re-filed under #2714 with the measured diff; their unquarantined `snap` twins added |
| 14 | every `prop-shadowing-signal × controlled-select` case, snap + three-point: the residue of #2758 after #2852 | new registry entry + browser-oracle fixture + rows (below) |

**The select residue.** #2852 made SSR select a hidden placeholder option for an out-of-range controlled select, so the *markup* half is fixed (pinned by `select-value-no-match-ssr`). The *live state* is not: the server has the placeholder selected (`selectedIndex` 0) and hydration's controlled-value effect assigns the out-of-range value, which the browser resolves to `selectedIndex` -1. Both read as blank. Landed as the three-piece set:

- registry entry `select-out-of-range-selected-index` (silent);
- browser-oracle fixture `select-out-of-range-hydration` (fixture-root component + frozen snapshots — the plain corpus fixture never runs in a browser, so it could not be pinned executably);
- an `oracle-quarantine.ts` row for it, and the 14 pairwise rows citing the entry.

Pairwise after triage: `940 passed, 0 failed`. The ledger's header note about the "deliberately left out" select pair is updated to point here.

## Role split between the sweeps

`spec/testing.md` gains a section on what each of the four browser sweeps owns — frozen corpus (per PR), mutations, static feature combinations (pairwise), transitions (explore) — why a pairwise finding that depends on a sequence belongs in explore and vice versa, the note that pairwise *generates* while explore *models* until a DSL makes the composer a second scenario producer, and the operating rule that a nightly sweep's only steady state is green (a red nightly is a red `main`).

## Test plan

- [x] `bun test packages/adapter-tests/explore` — 35 pass (IR smoke covers all four scenarios).
- [x] `test:explore` locally — 328 passed / 0 failed.
- [x] `test:pairwise` locally — 940 passed / 0 failed (was 918 / 22).
- [x] `oracle.playwright.ts` + `fixture-hydrate.playwright.ts -g select-out-of-range` — 3 passed (snap / three-point quarantined and still failing as recorded).
- [x] `bun test packages/adapter-tests -t select-out-of-range` — 18 pass across 26 files.
- [x] `limitations.test.ts` 199 pass; `limitations-join.test.ts`: all new rows pass (its 11 pre-existing failures reproduce on `main` in this environment).
- [x] `compat:lock`, `support-matrix:lock` and `scripts/coverage-map.ts` regenerated for the new fixture (all three CI freshness checks green on the current head).

Follow-up not in this PR: the 138 pairwise rows still citing `issue: #2714` (and the four corpus rows in `oracle-quarantine.ts`) predate the registry and could migrate to a `ref-effect-attr-state-ssr` entry listing `accordion` / `radio-group` et al.


kfly8:
The red `pullfrog` check on 75c89b2 is not this PR's: the review itself was posted (`✅ No new issues found`, review 5247615414), and the run then failed with `agent run timed out after 20m` while the agent was still writing its own learnings file after the review — the log's last action is an edit to Pullfrog's notes, not anything in this diff. The check's own summary states it reports run status only and gates nothing unless required by branch protection, and every repo CI check on this head is green. Nothing to port or fix here; leaving it as is.

---

Issue #2719, opened 2026-08-26
todo-app: snap/three-point flag the /* @client */ regions materializing after hydration — by design, oracle cannot distinguish placeholder fill from a defect


kfly8:
## Original problem (superseded — see below)

The #2481 oracle harness (PR #2713) `snap` + `three-point` oracles flagged `todo-app` with a large structural diff after hydration (footer/filter/count section reflow) — too big to attribute to one element or attribute at filing time. Filed with `bug` + `known-limitation` on the assumption this might decompose into #2714 (attr reset) or #2715 (rest-spread loss), or be something new and fixable.

## Update: narrowed and reclassified

Re-measured against the stack head with #2737/#2778/#2794/#2800/#2700/#2741/#2702 all applied (none of them touch `TodoApp.tsx`'s SSR markup or client JS, so the result is unaffected by which exact commit is used). The divergence is now **fully narrowed** and is **not a bug**:

`integrations/shared/components/TodoApp.tsx` marks four regions `/* @client */` — the filtered keyed loop (`<ul class="todo-list">`), the two todo-count text expressions (`<strong>`, the `'item'`/`'items'` plural), and the clear-completed conditional button. Per `spec/callback-fidelity.md` §4, `/* @client */` is an explicit, user-controlled escape hatch: **SSR renders an empty placeholder by contract**, and the browser fills it in after hydration. That is exactly what `snap`/`three-point` are seeing and flagging as a "large structural divergence" — it isn't one.

Verified:
- With the four `@client` regions masked out, the hydrated tree is **byte-identical** to the SSR tree.
- The three-point oracle's second comparison (hydrated vs. csr-mount) **agrees** structurally and in DOM state — only the SSR-vs-hydrated leg (comparing against the intentionally-empty placeholder) disagrees.
- `idempotence` passes.
- `todo-app-ssr` — the marker-free twin of this exact component — passes all three oracles cleanly, confirming the mechanism is specifically the `@client` markers, not something structural to `TodoApp.tsx` itself.
- Not #2714 (no SSR-present element has an attribute corrected/reset) and not #2715 (no named-prop mirror involved) — different mechanism from both.
- Already independently acknowledged in `packages/adapter-tests/src/__tests__/ssr-hydration-contract.test.ts` (`todo-app` is listed there with the same explanation).

## Why `known-limitation` (bare tier), not `bug`

[...]

Issue #2718, opened 2026-08-26
Oracle harness: teach the DOM-state normalizer which runtime bookkeeping is by-design invisible (bf-po, data-key, #1971 layout styles)


kfly8:
## Problem

The #2481 oracle harness (PR #2713) compares "user-visible DOM state" across SSR / hydrated / CSR-mount, but its first sweep quarantined three rows whose diffs are (probably) runtime bookkeeping the user never sees — the oracle can't currently tell "expected marker lifecycle" apart from "lost attribute":

- `bf-po` portal-origin markers: present in the SSR placeholder, removed when hydration relocates the portal content (`dialog`/`dropdown-menu`/`popover`/`portal` — the real ordering bug those fixtures also exposed is #2717).
- `tag-cloud`: `data-key="…"` loop-reconciliation markers present in SSR, removed once hydration claims the row.
- `carousel`: SSR bakes `style="transform: translate3d(0px, 0px, 0px)"` on the track; embla re-measures real (CSS-less host page) geometry on init and removes the inline style — the #1971 layout-dependence caveat, not a hydration defect.

Each needs a deliberate decision: if the lifecycle is by-design, the normalizer (`e2e/dom-state.ts` / `oracle.playwright.ts`) should canonicalize it away on BOTH sides and the fixture graduates from the quarantine ledger; if not, it becomes a real bug issue. Decide per row — don't blanket-strip all `data-*`/`bf-*`, since attribute loss is exactly what the oracle exists to catch (#2715).

## Graduation

For each row: confirm the intended lifecycle, either extend the normalizer (with a docstring saying WHY that marker is invisible) or file the bug, then delete the corresponding `oracle-quarantine.ts` entries.

kfly8:
## Re-measured 2026-09-18: the direction of all three rows is backwards

Same misread as #2714's correction. `assertSnapshotsAgree` (`e2e/oracle-core.ts`) runs `expect(normalize(pre)).toBe(normalize(post))`, so in the oracle's diff **Received = SSR** and **Expected = hydrated**. Re-run with the quarantine bypassed on `main` (e5a24e1), reading the diff in that orientation, none of the three rows is "SSR bookkeeping the runtime drops". In every case **hydration adds something SSR never rendered**:

| row | SSR (pre) | hydrated (post) | who writes it |
|---|---|---|---|
[...]

Issue #2717, opened 2026-08-26
Portal content lands at a different document.body child position on hydration vs pure CSR mount


kfly8:
## Problem

For every portal-bearing fixture (`dialog`, `dropdown-menu`, `popover`, `portal`), replaying the same interaction sequence on a hydrated page and on a pure CSR-mounted page leaves the portaled content at DIFFERENT positions in `document.body`'s child order. Found by the #2481 oracle harness (PR #2713) idempotence oracle; the marker-suffix normalization (`canonicalizePortalOriginMarker`) is already applied, so this is a real ordering divergence between the two mount paths, not a normalization gap.

Child order inside `body` is user-visible state: it decides paint order between unpositioned portals, focus traversal, and `querySelector` results. The two paths should agree on where portal content is appended.

Two sub-questions bundled here (same fixtures, same triage):
1. The ordering divergence above — the actionable bug.
2. The SSR placeholder's `bf-po="…"` portal-origin marker is gone after hydration relocates the content. Plausibly the INTENDED cleanup once the portal claims its content — if so, the oracle's normalization should learn to ignore `bf-po` (tracked as a harness refinement in the companion oracle-normalization issue) and the `snap`/`three-point` rows for these fixtures collapse into just the ordering question.

## Repro / pins

Quarantined in `packages/adapter-tests/e2e/oracle-quarantine.ts` (rot-checked), oracles `snap`/`three-point`/`idempotence` per fixture.

## Graduation

Make hydration's portal relocation and CSR mount's portal append converge on one insertion rule (or document + normalize a deliberate difference), confirm the four fixtures pass idempotence, delete the ledger rows (the snap/three-point rows may graduate earlier via the `bf-po` normalization decision).

kfly8:
Re-measured the four rows that cite this issue (`dialog`, `dropdown-menu`, `popover`, `portal`) with the quarantine bypassed, reading the diff in the direction `assertSnapshotsAgree` actually reports (Received = SSR, Expected = hydrated). The `bf-po` marker is not lost after hydration — it is **added**: SSR renders the overlay/content subtree inline inside the component root with no `bf-po`, and the hydrate-time `ref` callback's `createPortal` moves it to the end of `document.body` and stamps `bf-po`. A structural relocation no adapter performs server-side, so not portal cleanup.

[...]

Issue #1971, opened 2026-06-24
Add carousel to the fixture-hydrate corpus: embla import-map + scroll/drag interactions (#1467 Phase 3)


kfly8:
## Summary

The last remaining **#1467 Phase 3** item: bring `carousel` into the real-browser **fixture-hydrate** corpus. Two blockers — embla module resolution in the host page, and a scroll/drag interaction vocabulary.

**Context — already done:** Phase 2 corpus (30 real-browser fixtures) is complete; cross-adapter parity is closed (#1896 Go children, #1897 Mojo/Xslate); `calendar` cross-adapter SSR landed in #1965 (covered by the deterministic compile-conformance test + E2E, intentionally *not* a frozen fixture). `carousel` already compiles clean on all four adapters — its SSR template is static and embla is client-only — so the only gap is the real-browser hydration layer.

## Blocker 1 — embla dynamic import resolution in the fixture-hydrate host

`ui/components/ui/carousel/index.tsx` loads embla via a browser dynamic `import('embla-carousel')`. The fixture-hydrate host page (built in `packages/adapter-tests/e2e/fixture-hydrate.playwright.ts`) serves the frozen template + client JS but has no resolution for the bare specifier `embla-carousel`.

**Precedent — full E2E already solves this:**
- `site/ui/renderer.tsx:66-104` injects `<script type="importmap">` mapping `'embla-carousel' → '/static/lib/embla-carousel.esm.js'`.
- `site/ui/build.ts:470-474` copies `node_modules/embla-carousel/esm/embla-carousel.esm.js` into `dist/lib/`.

**Plan — port that to the fixture-hydrate host:**
1. Serve the embla ESM bundle from the Playwright host (route/copy `node_modules/embla-carousel/esm/embla-carousel.esm.js`).
2. Inject an importmap `<script>` into the host page `<head>` (same map as `renderer.tsx`).
3. **Gate it**: only fixtures that declare an external (e.g. a new `externals: ['embla-carousel']` field on the fixture spec, or auto-detected from the client-JS imports) get the importmap, so other fixtures are unaffected.

## Blocker 2 — scroll / drag interaction vocabulary

[...]

PR #3058, opened 2026-09-18
Graduate the tag-cloud and carousel hydration-parity rows; migrate the portal group to the registry; pin the Go union-typed signal gap

kfly8:
Four commits on the same branch, all fallout of the #2718 / #2717 triage: two fixes that graduate `oracle-quarantine.ts` rows, one ledger migration to the known-limitation registry, one registry pin found while landing #3056, and the `ui/meta` regeneration the carousel change requires.

## 1. Complex flatMap leaves: `data-key` at SSR too (graduates the `tag-cloud` row from #2718)

A keyed leaf inside a complex `.flatMap()` body (statements before the `return`) carried its reconciliation attribute only on the client. The Hono rawBody SSR evaluates the branded TSX through a JSX runtime, which drops `key` from the HTML, while `mapArray` stamps `data-key` on every row it adopts. Hydration therefore changed the DOM on exactly the attribute reconciliation keys on — the `[snap] tag-cloud` / `[three-point] tag-cloud` oracle failures (Received = SSR without `data-key`, Expected = hydrated with it).

- `buildFlatMapCallback` (`packages/jsx/src/jsx-to-ir.ts`) collects each element leaf's `key` attribute from the AST and `flatMapRawBodyWithLeafKeys` splices ` data-key={String(<key>)}` right after it in the branded source text (position-based over attribute spans, no text matching), so the JSX runtime renders and escapes it like any attribute.
- The client string templates go through one door, `leafKeyAsDataKeyAttr` (`packages/jsx/src/ir-to-client-js/html-template.ts`): the leaf's `key` attr is rewritten **in place** to an escaped `data-key` expression attribute for the hydrate template, the CSR descriptor (`renderFlatMapClientBody`), the placeholder template and the module-scope template alike — so SSR, hydrate template and csr-mount agree byte-for-byte, attribute order included. `String(...)` matches `mapArray`'s own `String(__bfD.k ?? __bfI)` stamping.
- `flatmap-segments.test.ts` pins the new contract; the doc-examples snapshot and the `tag-cloud` fixture snapshots are regenerated (`<li data-key="1:a &amp; b" data-tag="a &amp; b">`); the `tag-cloud` quarantine row is deleted. Changeset: `@barefootjs/jsx` patch.

## 2. Carousel: SSR renders the resting state (graduates the `carousel` row from #2718)

Measured with the row bypassed, hydration ADDS two things SSR lacked: embla's resting transform on the track (`style="transform: translate3d(0px, 0px, 0px);"`) and an enabled Next button — the component SSR'd no `style` and a hard-coded `disabled` on both nav buttons while `canScrollNext` started `false`. `ui/components/ui/carousel/index.tsx` now renders the resting state of the default options (first slide, no loop) on the server: the transform on the track, in the form the browser serializes once embla writes `style.transform`; Next enabled (`canScrollNext` initial `true`, no `disabled` literal); Prev still disabled. Embla's first `updateButtons` corrects a single-slide or looped carousel as before. The fixture's own contract already said "at rest on slide 1 prev is disabled and next is enabled". Not the layout-dependence caveat the old row assumed. `ui/meta/carousel.json` regenerated for the new initial value (fourth commit; the `update-meta` job and the cli drift guard caught it).

## 3. Portal group → registry entry `ref-callback-portal-content-inline-at-ssr` (#2717 rows)

Re-measured in the direction `assertSnapshotsAgree` reports (Received = SSR, Expected = hydrated): SSR renders the overlay/content subtree **inline** with no `bf-po`; the hydrate-time `ref` callback's `createPortal` moves it to the end of `document.body` and stamps `bf-po`. So the marker is added, not lost, together with a structural relocation no adapter performs server-side — a `silent` gap, not cleanup. The `dialog` / `dropdown-menu` / `popover` / `portal` rows now cite the entry with corrected reasons and no issue URL. The fix design (lower the `ref` + `createPortal` pattern to a neutral portal node rendered at the adapter's outlet, `BfPortals` on Hono) is #3059.

**Carousel × fragment-wrap idempotence** (`mutation-quarantine.ts`, cited #1971): with Next enabled from the start the click no longer times out, and what remains is the drag step's pixel offset (`-315.48px` vs `-295.64px`) — exactly the pointer-position dependence `IDEMPOTENCE_EXCLUDED` already excludes for the base fixture. The exclusion maps move to `e2e/oracle-exclusions.ts`, shared by `oracle.playwright.ts` and `mutation.playwright.ts` (a mutant replays its base fixture's steps on the same host page), and the row is deleted.

**Left as is:** `todo-app` (#2719). Its SSR-empty `/* @client */` regions are the spec's contract (`spec/callback-fidelity.md` §4); the registry has no kind for that (a `by-design` entry requires a refusal diagnostic), so the row keeps its executable record and issue URL.

## 4. Pin: the Go adapter drops a union-typed signal's literal initial value

Found while landing #3056: `createSignal<string | undefined>('one')` renders `title=""` and empty text on the Go Template adapter, where the reference renders `title="one"` and `one` — the generated Go types the Props field `interface{}` and seeds it `nil`. Three-piece set, not a fix: fixture `signal-optional-init` (attribute + text, click cycling `one` → `undefined` → `one`), registry entry `optional-typed-signal-initial-value-dropped` (silent), go-template `renderDivergences` pin, changeset `@barefootjs/go-template` patch. The other seven template adapters render the literal.

## Verification (head 1cfd244)

- CI: all 48 checks green on this head (`check`, `lockfile`, `update-meta`, `update-expected-html`, `update-doc-snapshots`, every adapter e2e), mergeable.
- Local: `packages/jsx` 3502 pass (doc-examples snapshot regenerated, the only diff is the flatMap example's `data-key`), `tsgo --noEmit` clean; `packages/adapter-hono` 1957 pass; `packages/adapter-tests` 3018 pass; `packages/compat` 1196 pass (`limitations` 235, `limitations-join` 506); `ui` 1235 pass (carousel IR test 17); `carousel-cross-adapter.test.ts` 9 pass; cli `ui/meta/*.json drift guard` 63 pass.
- Real browser: `tag-cloud` and `carousel` fixture-hydrate + oracle `snap` / `three-point` (+ `idempotence` for tag-cloud) pass with their rows gone; oracle + mutation for `dialog` / `dropdown-menu` / `popover` / `portal` / `carousel` / `todo-app`: 34 pass, 14 skipped (rot checks hold on the re-pointed rows). `signal-optional-init`: fixture-hydrate + oracle 4 pass; JSX conformance on blade / erb / jinja / mojolicious / twig / xslate / rust 3 pass each, go-template under `GOTOOLCHAIN=go1.25.6` 1 pass / 1 skip.
- `generate-expected-html.ts`, `coverage-map.ts`, `compat:lock`, `support-matrix:lock`, `meta:extract` re-run with no further drift.


kfly8:
Re Pullfrog's finding on the carousel's SSR-baked `canScrollNext = true` (review 5253685493):

Verified: it does assume more than one slide, and the server cannot know the count — `CarouselNext` consumes context only inside its `ref` callback, and `Carousel`'s children are opaque at SSR (the compiler cannot count `CarouselItem`s across template adapters). So the initial values are a choice between the two pre-hydration states, not something SSR can compute:

- **Before this PR**: `disabled` on both buttons. Correct for a single-slide carousel, wrong for every multi-slide one — hydration flipped Next to enabled (the `[snap] carousel` oracle failure this PR graduates).
- **After this PR**: Prev disabled, Next enabled. Correct for a multi-slide carousel, wrong for a single-slide or empty one — embla's first `updateButtons` on init disables Next, exactly the correction the multi-slide case used to need.

A carousel exists to move between several slides (the shipped demos render 3–5 items; a one-item carousel is a degenerate configuration, not a documented use), so the multi-slide resting state is the one SSR renders. On the no-JS angle: neither button scrolls before embla mounts (the click handlers are attached in `ref` callbacks and embla is a dynamic import), so no-JS delivery is not a goal these attributes serve; the `disabled` attributes only describe pre-hydration state, and the corrected state now matches the common case.

Pushed a comment-only commit that records this trade-off next to the signals in `ui/components/ui/carousel/index.tsx`, so the next reader does not take the initial value for a measured fact. Whether a single-item carousel is a supported use is @kfly8's call; if it is, the alternative is to keep `disabled` on Next at SSR and re-quarantine the oracle row under a registry entry, which trades the common case back for the edge case.

---

PR #3069, opened 2026-09-19
Fall back to selectedIndex 0 for an out-of-range controlled &lt;select&gt; (#3066)

### LINKED ISSUE #3066: Out-of-range controlled `<select>`: placeholder selected on the server, nothing selected after hydration (silent → fix, or at least loud)
Registry entry: `packages/adapter-tests/limitations/select-out-of-range-selected-index.ts` (kind `silent`). Pinned by the `select-out-of-range-hydration` row in `packages/adapter-tests/e2e/oracle-quarantine.ts` (`snap` + `three-point`). Compiler / runtime gap (the controlled-value hydration effect), affecting every adapter equally. History: #2758 → #2852 made SSR select a hidden placeholder option for an out-of-range controlled value instead of the browser's first-option default; this is the surviving half.

## What happens

Given a controlled `<select value={...}>` with statically enumerable `<option>`s whose bound value matches none of them:

- **Expected**: the server-rendered select and the hydrated select agree on the live selection (`selectedIndex` / `value`).
- **Actual**: SSR has the hidden placeholder option selected (`selectedIndex` 0, `value` ""); the hydration controlled-value effect then assigns the out-of-range value, which the browser resolves to no selection (`selectedIndex` -1). Both read as blank to the user, but the live state changes across hydration and the two legs of the oracle disagree.

Reproduce: bypass the row and run `oracle.playwright.ts -g select-out-of-range-hydration` in a real browser.

## Goal

Fix so both sides agree. Two consistent options; pick the one that keeps the reference contract:

1. The controlled-value effect, when the bound value matches no option, selects the placeholder (`selectedIndex` 0) instead of assigning the unmatched value — the hydrated state then equals SSR's, and csr-mount (which builds the same markup) agrees too.
2. Or SSR stops emitting the placeholder as selected and the effect's `-1` becomes the contract on both sides (this changes `expectedHtml` and needs the "Hono is the reference" rule applied explicitly).

Option 1 is the smaller change and matches what #2852 chose for SSR. **Minimum acceptable outcome if neither is reachable: make it loud** — a `BFnnn` diagnostic for a controlled select whose static options provably exclude the bound literal, escapable with `/* @client */`, and convert the entry to kind `refusal`.

## Graduation mechanics (per CLAUDE.md's registry rule)

- Full fix: delete the oracle-quarantine row and the entry; the fixture is the regression test (add an action step that changes the value to an in-range option and back, so the live `selectedIndex` is asserted by `fixture-hydrate`).
- Loud only: the entry becomes `kind: 'refusal'` with `diagnostic`, pinned via `conformancePins`; the oracle row is deleted.
- Re-run `limitations.test.ts` / `limitations-join.test.ts`, the browser oracle, and regenerate the lockfiles.

kfly8:
Closes #3066.

Stacked on #3067 (the branch is `issue-3066-select-out-of-range`, based on `issue-3065-attr-state-ssr`) — this PR's own diff is just the last commit (`067f8267a`); the earlier two are #3067's.

## What was broken

A controlled `<select value={x}>` whose value matches no `<option>` left SSR and the hydrated DOM disagreeing on the *live* selection state, even though both render blank. #2852 already makes SSR select a hidden placeholder (`selectedIndex` 0), but the hydration effect just assigns the out-of-range value to `.value`, which the browser resolves to no selection at all (`selectedIndex` -1). Registered as `select-out-of-range-selected-index`.

## The fix

Option 1 from the issue (the smaller of the two): `emitValueUpdateStatements` — the codegen for a developer-authored `value=` attribute, used identically for every `<select>` regardless of surrounding structure — now checks, after assigning `.value`, whether a single-selection (`!multiple`) `<select>`'s value didn't take, and falls back to `selectedIndex = 0` to match the #2852 placeholder SSR already rendered. `multiple`/list-box selects are excluded: an unmatched value already leaves them with nothing selected on both legs (see `select-multiple-value-no-match-ssr`'s docstring), so there's no placeholder to reconcile against and forcing a selection would be its own behavior change.

## A bonus graduation

Being the single shared codegen path every `<select value={…}>` compiles through, this also fixes the pairwise sweep's 14 `prop-shadowing-signal × controlled-select` out-of-range rows in `pairwise-quarantine.ts` that cited the exact same registry entry for the identical symptom (`SELECT[0]#selectedIndex: 0 !== -1`) across different surrounding structures (early-return, nested loops, conditionals, child components, fragment rows) — none of which touch this codegen function differently. Graduated alongside the dedicated fixture; the registry entry (`select-out-of-range-selected-index`) is deleted since its fixture list would otherwise be empty.

## Verification

- New compiler unit test pinning the emitted fallback shape (`tagName === 'SELECT'`, `!.multiple`, `.selectedIndex = 0`).
- `packages/adapter-tests/src/__tests__/limitations.test.ts` (235 pass) and `packages/compat/src/__tests__/limitations-join.test.ts` (438 pass).
- Full real-browser oracle suite (`oracle.playwright.ts`): 169 passed, 1 pre-existing unrelated skip — no regressions, including every other `<select>`/`<textarea>`/value-bound fixture whose emitted code textually changed (the new branch is inert for non-`<select>` targets) but behaves identically.
- `select-out-of-range-hydration` specifically re-verified green in a real browser with the quarantine row removed (not just "no longer quarantined" — actually ran and passed).
- `doc-examples.test.ts` (0 snapshot drift) and per-fixture Hono + ERB adapter conformance for every `select*`/`textarea*` fixture.
- I did **not** run the full `bun run test:pairwise` pipeline (manifest generation + browser suite) in this sandbox — it didn't complete in the available time (a few hundred generated cases, each a full compile). The 14-row graduation above rests on the mechanical argument that these cases share the exact codegen function with the fixture I did verify live, not on re-running the sweep itself. Worth confirming with `bun run test:pairwise` before merge if that's cheap to do in CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RXKKui1Kr88Rh8ZdM3xKGM)_


kfly8:
Re the pairwise nit: ran the full sweep locally on the current head ad94ab5 (`pairwise-generate.ts` → 451 cases, then `pairwise.playwright.ts` in a real Chromium): **940 passed, 0 failed** (the 18 skips are the documented `LEGITIMATELY_ROW_LESS` combos). With the 14 `prop-shadowing-signal × controlled-select` rows deleted from `pairwise-quarantine.ts`, every one of those cases now runs as an ordinary test and passes, so the graduation is measured, not just argued.

---

kfly8:
`e2e-site-ui (1)` went red on ad94ab5 with `carousel.spec.ts › clicking next navigates to next slide` (`expect(prevBtn).not.toBeDisabled()` failed). That is not this PR's: the diff touches nothing near the carousel, and the same shard was green on the base branch (#3067, 2bbd540) three minutes earlier.

Root cause is a race in the spec itself: the carousel bakes Previous's first-slide `disabled` into SSR, so the test's "wait for embla" guard (`toBeDisabled`) passes before embla's dynamic import resolves, and the single Next click lands on `emblaApi?.scrollNext()` with no api yet. It reproduces deterministically with a 2s delay routed onto the embla chunk. Fix is #3077 (retry the click until embla answers); I merged that branch into this PR as 734952e so CI re-runs against it, and it no-ops once `main` carries it.

---

PR #3076, opened 2026-09-19
Revert "Run every Container integration's server under tini" (#3071, #3072)

kfly8:
Reverts e38359080 (#3071 + #3072) to restore the known-good images.

## Why

After the deploy of e38359080, `/integrations/rails` and `/integrations/sinatra` answer **500** on every request once `sleepAfter` has passed:

```
Error proxying request to container: The container is not listening in the TCP address 10.0.0.1:8080
```

The 500 comes back in ~0.3s, so the Container believes the instance is up (no cold start is attempted) while Puma is no longer listening. Before the deploy the same probe cold-started both apps normally (sinatra 2.43s → 200). The other 14 integrations are healthy after the deploy.

It does not reproduce locally: sending SIGTERM to tini (PID 1) makes Puma log "Gracefully stopping" and the container exits with 143 within a second. So the cause is not understood yet, and per the repo policy the merge is reverted first and the fix goes forward on a branch.

## Effect of the revert

The six integrations whose server ignores SIGTERM as PID 1 (axum, flask, django, php, blade, laravel) go back to running ~15 minutes past `sleepAfter` per wake-up. That is a cost regression only, not an outage.

The tree after this revert is identical to f712ac082 (the commit before the merge).


kfly8:
Closing without merging: instead of reverting all sixteen images, tini is being dropped from the two Puma images (rails, sinatra) only — the only integrations that hit the 500 "container is not listening" state after a rollout. Follow-up PR incoming.

PR #3081, opened 2026-09-19
Demote compiler-ABI exports of @barefootjs/client to @internal

kfly8:
Closes #3080

## What

Tags `@internal` on nine `@barefootjs/client` exports that are only ever called by compiler-emitted code or by tooling, so they leave `docs/core/advanced/api-reference.md` the way `forwardProps`/`unwrap` already do:

- `createDisposableEffect` (was Alpha, 0.1.0) — emitted by `packages/jsx/src/ir-to-client-js` for branch-scoped effects.
- Profiler plumbing (was Alpha, 0.11.0): `beginTurn`, `endTurn`, `createRecordingSink`, `setProfilerSink`, `ProfilerEvent`, `ProfilerEventSink`, `ProfilerEventType`, `RecordingSink`, `SubscriberKind` — compiler-emitted profile-mode code and `bf debug profile`.

No export is removed and no behaviour changes — this is a JSDoc tag + generated-docs change only.

**Not demoted after review: `cleanupPortalPlaceholder` stays Beta.** The issue listed it as a demotion candidate on zero in-repo callers, but nothing in the runtime or the compiler calls it on the author's behalf, `docs/core/components/portals.md` teaches calling it directly, and the analyzer's `BROWSER_ONLY_CLIENT_APIS` classifies it as author-callable. It is a public function with no in-repo users — a remove-or-keep decision for #3078 task C, not a tiering one.

## Where the tags live

- `createDisposableEffect`, `SubscriberKind`, `ProfilerEventSink`, `setProfilerSink`, `beginTurn`, `endTurn` — `packages/client/src/reactive.ts`.
- `createRecordingSink`, `RecordingSink` — `packages/client/src/profiler-events.ts`.
- `ProfilerEvent`, `ProfilerEventType` — declared in `packages/shared/src/profiler-events.ts` and re-exported through the client package; the generator resolves a re-export to its *declaring* symbol (`checker.getAliasedSymbol`), so that's where the tags have to move.

## Beta closure

No Beta API's signature names any of the newly-internal types. `createEffect`'s `__bfKind: SubscriberKind` parameter is already excluded from the beta-closure check by the existing `__bf`-prefixed-parameter rule (`signatureTypeNames` in `scripts/generate-api-reference.ts`), so no `betaClosureExceptions` entry is needed.

## Runtime surface preamble

Left `scripts/generate-api-reference.ts`'s Runtime intro paragraph as-is — it already says "The compiler ABI is not listed at all", which generically covers `@internal` exports.

[...]

PR #3088, opened 2026-09-19
implement-issues workflow: pass the human co-author in; defer commit trailers to CLAUDE.md

kfly8:
## What

`.claude/workflows/implement-issues.js` told its Implement and Polish agents to "commit with correct Co-authored-by trailers per CLAUDE.md". CLAUDE.md's rule for a remote session (git author `Claude`) is to pick the human collaborator via `AskUserQuestion` before the first commit, which a workflow agent cannot do, so on the first real run (#3081 / #3083) the agents produced four commits carrying only the model trailer and the human co-author had to be added by rewriting history afterwards.

## Change

The co-author rule is CLAUDE.md's, shared by every session, so the workflow carries neither a person's identity nor its own copy of the trailer format.

- `args.coauthor` is **required**: the `"Name <email>"` of the human every commit is co-authored with, validated up front with a clear error when missing. The caller resolves it the way CLAUDE.md says (git log for the human identity, `AskUserQuestion` if ambiguous) and passes it down; nothing is hard-coded.
- `COMMIT_TRAILER_RULE`: one string interpolated into both the Implement prompt (step 5) and the Polish prompt (step 4). It defers to CLAUDE.md's "Git Commit" section for the format and order, and only supplies what an agent cannot obtain by itself: the caller-chosen `Co-authored-by:` line, plus the default remote author (`git -c user.name=Claude -c user.email=noreply@anthropic.com`).
- Usage header documents `args.coauthor`.

Prompt text and one validated argument; the workflow's phases, agents and return shape are unchanged.

## Verification

- The script uses a top-level `return` (Workflow runtime), so `node --check` was run on a copy wrapped in an async function with `args = { coauthor: 'A B <a@b.c>' }`: syntax OK.
- The `coauthor` validation regex accepts `Jane Doe <jane@example.org>` and rejects `''`, a bare login, and `x <not-an-email>`.
- `grep 'kfly8 <'` over the file finds nothing: no identity left in the script.

https://claude.ai/code/session_01CcZysQNJ4PgR2wniAMzVv3


kfly8:
@kfly8 ready for review.

**What**: the implement-issues workflow now takes the human co-author from its caller (`args.coauthor`, required and validated as `"Name <email>"`) instead of leaving it to sub-agents that cannot run CLAUDE.md's `AskUserQuestion` step, and the prompt text defers to CLAUDE.md's "Git Commit" section rather than restating the trailer format. No identity is hard-coded.

**How to verify**:
```sh
grep -n "kfly8 <" .claude/workflows/implement-issues.js          # expect no output
grep -n "COMMIT_TRAILER_RULE" .claude/workflows/implement-issues.js  # one definition, two prompt sites
```
Then invoke the workflow once without `coauthor` (any issue number) and once with `coauthor: 'Name <email>'`.

**Expected result**: no hard-coded identity; the constant is defined once and interpolated into the Implement (step 5) and Polish (step 4) prompts; the run without `coauthor` fails immediately with the "args.coauthor is required" error before any agent starts, and the run with it proceeds, its commits ending with the caller's `Co-authored-by:` line after the implementer trailer. CI on `58e03b3` is green and Pullfrog reports no issues.

---

PR #3096, opened 2026-09-19
Try the rails Container on the basic instance type

kfly8:
## Summary

`/integrations/rails` is down in production: after the first idle stop (`sleepAfter`) following each deploy, every request gets

```
500 Error proxying request to container: The container is not listening in the TCP address 10.0.0.1:8080
```

in ~0.4s. The rails Durable Object believes its container is running and healthy (so `containerFetch` skips the start), `wrangler containers instances` reports the instance as `stopped`, and a tail shows no alarm events — so nothing ever resyncs it.

What has been ruled out:

- **tini** — #3087 took it off rails and the failure came back at the first stop after that deploy.
- **Memory** — ~55–70 MiB under a 256 MiB limit, including 30 concurrent requests.
- **The image and the Worker code** — a copy deployed as `barefootjs-rails-debug` (workers.dev only, same image digest, same `lite` config, lifecycle logging added) stopped and cold-started normally through every cycle tried: 30s and 2m `sleepAfter`, and after a rollout that swapped the image while the container was running.

This PR changes a single variable: rails moves from `lite` to `basic` (1 GiB / 1/4 vCPU / 4 GB, the type spring already runs on). The Durable Object and the image stay as they are. The `wrangler.toml` comment says why, so the choice can be reverted to `lite` if it turns out not to matter.

Cost: rails' Container memory/disk is billed at the `basic` size while it is awake (memory 4× `lite`); Durable Object duration is unaffected.

## Test plan

- [ ] After deploy: `/integrations/rails` answers 200
- [ ] Wake it, wait 4 minutes (past the 2m `sleepAfter`), request again: 200 after a cold start, not the 500 — repeated for two cycles

https://claude.ai/code/session_01WqpSzjRDbPxXPDZJkizDAt


kfly8:
Closing without merging: a copy of this worker on `lite` with the same image works normally, so the instance type is unlikely to be the cause. Replacing it with a Durable Object name change for rails (fresh DO + container instance) as the workaround.

PR #3100, opened 2026-09-19
Clamp Popover content position to the viewport

kfly8:
## Summary

`PopoverContent`'s `updatePosition` anchored purely off the trigger's `getBoundingClientRect()` (`rect.bottom + 4` for the default bottom side, `rect.left`/`rect.right`/centered midpoint for align) with no check against `window.innerHeight`/`innerWidth`. When the trigger sits near the bottom of the viewport, the content — a `position: fixed` overlay — renders partly or fully below the visible viewport, with no scroll able to reach it (fixed elements don't move on page scroll).

Found while babysitting an unrelated PR (#3099): `site/ui/e2e/popover.spec.ts`'s "Form Demo › PopoverClose button closes popover" test intermittently fails with "element is outside of the viewport". The Form Demo's trigger sits well down the reference page; Playwright's scroll-into-view for the click lands it near the viewport's bottom edge often enough, and the resulting Content (a multi-row settings form, tall enough to overflow) renders its Cancel/PopoverClose button past `innerHeight` — with nothing to reposition it.

Sibling overlays already handle this: `select`/`combobox` clamp their available height (their content is a scrollable option list), `context-menu` clamps its computed x/y directly against `innerWidth`/`innerHeight`. Popover's content isn't naturally scrollable (arbitrary form content), so this follows `context-menu`'s approach: `top`/`left` are now bounded to `[gap, viewport - contentSize - gap]` on both axes, keeping content anchored as close to the trigger as the viewport allows instead of letting it run off any edge.

## Changes

- `ui/components/ui/popover/index.tsx`: clamp `updatePosition`'s computed `top`/`left` to the viewport.
- `packages/adapter-tests/fixtures/__snapshots__/popover.client.js`: regenerated (`bun run packages/adapter-tests/scripts/snapshot.ts popover`). SSR output (`popover.html`) is byte-identical — only the client bundle's `updatePosition` body changed.
- No changeset: both `@barefootjs/ui` and `@barefootjs/adapter-tests` are in `.changeset/config.json`'s `ignore` list.

## Test plan

- [x] `ui/components/ui/popover/index.test.tsx` — 16 pass, unaffected (no JSX/attribute change)
- [x] `packages/adapter-tests` CSR conformance suite — 413 pass
- [x] `site/ui/e2e/popover.spec.ts` — 9 pass
- [x] Reproduced the original failure locally by running the exact e2e spec at CI's own concurrency (`CI=1 --repeat-each=15 --workers=4`) — not reliably reproducible pre-fix (geometry-dependent), confirmed the fix holds with the same stress run: 135/135 pass on this branch
- [ ] Full `packages/adapter-tests` non-browser suite (running; will report if anything else needs updating)

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01RXKKui1Kr88Rh8ZdM3xKGM)_


kfly8:
Pushed 56a5d908d addressing Pullfrog's finding (no deterministic regression test for the clamp math).

Extracted the clamp math into a pure, exported `clampPopoverPosition` function in a new `ui/components/ui/popover/clamp-position.ts` — a plain `.ts` file rather than the `"use client"` `.tsx`, since these component source files are only meant to be compiled from source text, not executed directly (a direct import from `index.tsx` in the test file hits the wrong JSX runtime, confirmed by trying it first).

Pinned with 7 unit tests in `index.test.tsx`: the three side/align combinations, top-side positioning, and three edge cases — including the exact geometry of the bug this PR fixes (a trigger near the viewport's bottom edge with tall content), a trigger near the right edge, and content larger than the viewport on both axes.

Regenerated `popover.client.js`/`popover.popover.ssr.tsx` (`bun run packages/adapter-tests/scripts/snapshot.ts popover`) — `popover.html` (the actual SSR output) is unchanged, confirmed via `git diff`.

Verified: full `ui/` suite (1244/1244, Popover's own suite now 23/23), `adapter-tests` CSR conformance (413/413), the e2e `popover.spec.ts` (9/9), and a full `site/ui` production build to confirm the new sibling-file import resolves correctly through the real Vite bundling pipeline.

The ℹ️ finding (the same unclamped-position gap exists in `hover-card`/`dropdown-menu`/`menubar`/`navigation-menu`) is a legitimate, separate follow-up — left out of scope for this PR since it touches four other components' own positioning logic, not this one's bug.

---

kfly8:
Resolved a merge conflict with `main` (pushed 0ef551a13).

Turned out this PR's own original clamp fix (404e2071a) had already landed on `main` independently — as commit 67a7cf160, byte-identical content, apparently picked up into #3099's branch at some point before that PR merged. So the conflict was genuinely "the same fix, twice." Resolved by keeping this branch's version (the `clampPopoverPosition`-extracted, test-covered one) over `main`'s inline duplicate — functionally identical, strictly more testable — and regenerated the snapshot fixtures from source rather than hand-resolving the generated-file conflicts.

Re-verified after the merge: full `ui/` suite (1248/1248 pass), `adapter-tests` CSR conformance (410/410 pass — 413→410 is `main`'s own unrelated fixture changes, not this PR), `popover.spec.ts` e2e (9/9 pass), and a fresh `site/ui` production build.

---

kfly8:
Fixed `update-meta` failure: extracting `clampPopoverPosition` into `clamp-position.ts` added a new internal dependency for Popover that `ui/meta/popover.json` didn't yet reflect. Ran `bun run meta:extract` and pushed 7ef98d4dc (this push also carries the snapshot regeneration from the earlier merge-conflict resolution, already validated in the previous comment — only `ui/meta/popover.json` is new content here).

---

kfly8:
Fixed the `fixture-hydrate` failure (4 tests: `Failed to load script at .../popover/__client.js`).

Root cause: `packages/adapter-tests/e2e/fixture-host.ts` serves each fixture's compiled client bundle as a single in-memory string at `/<fixtureId>/__client.js`, with no route for a sibling module. `clamp-position.ts` lived beside `PopoverContent` at `ui/components/ui/popover/`, so the extracted `import { clampPopoverPosition } from './clamp-position'` survived into the compiled snapshot and 404'd when the browser tried to resolve it as a real ES module.

Fix: moved `clamp-position.ts` into `ui/lib/` (matching the existing `track-position.ts` pattern), which `snapshot-generator.ts`'s `inlineUiLibImports` already inlines directly into the compiled bundle instead of leaving an external import. `popover.html` (SSR output) is unchanged.

Verified locally:
- The exact failing Playwright command (`fixture-hydrate.playwright.ts` + `oracle.playwright.ts -g "popover"`): 4/4 pass
- Full CSR conformance suite: 410/410 pass
- Full `ui/` test suite: 1248/1248 pass
- `bun run meta:extract`: no further changes needed

Pushed in 95a9c9c58.

---

